PR #2, opened 2018-11-05
New installation subsection: preparing install medium

CameronNemo:
1. I am not completely satisfied with how the dd warning looked (it does not stick out much)
2. Do you think the installation section should use a table of contents and be split out into different pages? (not just for guides)

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -3,6 +3,7 @@
 - [About](./about/index.md)
   - [History](./about/history/index.md)
 - [Installation](./installation/index.md)
+  - [Preparing Installation Media](./installation/prep.md)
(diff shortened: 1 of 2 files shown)


the-maldridge:
I envision each part of the installation being its own page.  So things like "obtaining media" or "preparing media" could be their own pages.

the-maldridge:
Guides is meant to be, "here's this obscure hardware platform, lets install onto this".

For chroot install guides, we probably need a new subheading, maybe "chroot guides".

CameronNemo:
I have deferred to the website for downloading images, at least for now. It is probably best to prioritize the Wiki info.

I made the section added here a separate page. Let me know if you have any more comments!

PR #3, opened 2018-11-05
Add pages with text from the faq

Duncaen:
I'm not sure about the separation between maintenance and configuration, there is some overlap.


src/maintenance/packages/mirrors.md
```diff
@@ -0,0 +1,48 @@
...
+
+
+## Changing mirrors
+
+Copy all repository configuration files from `/usr/share/xbps.d` to `/etc/xbps.d` and then change the repository urls in `/etc/xbps.d`.
+
+```
+# mkdir -p /etc/xbps.d
+# cp /usr/share/xbps.d/*-repository-*.conf /etc/xbps.d/
+# sed -i 's|https://repo.voidlinux.eu/current|<repository>|g' /etc/xbps.d/*-repository-*.conf
```
CameronNemo: It would be nice to provide a command that works out of the box, but this is liable to become out of date. Perhaps a simple recommendation to edit the copied files directly would be best here.

src/SUMMARY.md
```diff
@@ -3,9 +3,34 @@
 - [About](./about/index.md)
   - [History](./about/history/index.md)
 - [Installation](./installation/index.md)
+  - [Downloading Void Linux](./installation/downloading.md)
   - [Guides](./installation/guides/index.md)
+    - [Foo](./installation/guides/foo.md)
```
CameronNemo: typo or...?

src/maintenance/packages/mirrors.md
the-maldridge: The long term goal is to default to an automatic redirection mirror service.  This service is actually live right now (auto.voidlinux.org) but isn't stable enough to recommend yet.

It would not be too difficult to provide a mirror configuration and chooser program if anyone wanted to write one.

the-maldridge:
Some files have a column wrap and some don't.  We should pick one and go with it.

Vaelatern:
I elect for column wrap at 80 chars, `fmt -w 80`.

PR #6, opened 2018-11-12
installation: import guide, format tables

bobertlo:


the-maldridge:
Please squash your commits and then set name to the format of section: thing you did.

PR #7, opened 2018-11-13
separate commands in codeblocks

platerza:
I've split all separate commands in codeblocks so that they are in their own codeblocks, for the sake of clarity.

For example:
```
$ cat /etc/sv/sshd/run
#!/bin/sh
ssh-keygen -A >/dev/null 2>&1 # Will generate host keys if they don't already exist
[ -r conf ] && . ./conf
exec /usr/bin/sshd -D $OPTS
# echo 'OPTS="-p 2222"' >>/etc/sv/sshd/conf
```
should be
```
$ cat /etc/sv/sshd/run
#!/bin/sh
ssh-keygen -A >/dev/null 2>&1 # Will generate host keys if they don't already exist
[ -r conf ] && . ./conf
exec /usr/bin/sshd -D $OPTS
```
```
# echo 'OPTS="-p 2222"' >>/etc/sv/sshd/conf
```

### DIFF
--- a/src/config/media/alsa.md
+++ b/src/config/media/alsa.md
@@ -5,6 +5,8 @@ Install the `alsa-utils` package make sure your user is part of the
 
 ```
 # xbps-install -S alsa-utils
+```
+```
 # usermod -a -G audio <username>
 ```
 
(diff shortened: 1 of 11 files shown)


Vaelatern:
This, I think, is a style question more than anything.

the-maldridge:
There's no reason to have separate code blocks unless you're going to also have an individual comment for each of those code blocks explaining what it does.

PR #9, opened 2018-11-13
added autoformat script and ran it on the tree

bobertlo:
This is just a suggestion for autoformatting. It only depends on pandoc. Shell scripts are not my strong point, but it works. I am not strongly attached to this solution, but thought I'd post it.


bobertlo:
I'd love feedback on this. pandoc has a ton of different modes if a different style is preferred. I just got it to a point where it doesn't break `mdbook`

platerza:
Whether this is the chosen solution or not, I think we should avoid the `bash` dependency and stick with `sh`.

bobertlo:
I agree, I forgot I switched it to bash at one point.

CameronNemo:
I sort of like the ````` feature from an editing perspective. Any way to get pandoc to keep those?

bobertlo:
I was looking into that before work and will see if I can get it to switch over, or at least keep existing lines formatted that way.

bobertlo:
I'm closing this because I do not see pandoc as a viable option for maintaining this content.

PR #11, opened 2018-11-13
config: add wpa_supplicant

bobertlo:


src/config/network/wpa_supplicant.md
```diff
@@ -1 +1,103 @@
...
+> dhcpcd.
+
+### Running wpa\_supplicant manually
+
+You may start
+[wpa\_supplicant.conf(5)](https://man.voidlinux.org/wpa_supplicant.conf.5)
+manually, if you are not already running
+[dhcpcd(8)](https://man.voidlinux.org/dhcpcd.8) :
+
+    $ sudo wpa_supplicant -B -i <interface_name> -c <path/to/configuration/file>
```
cvs-is-better: I would suggest avoiding using sudo, and use a # to indicate root shell, to be consistent with the rest of the docs.

src/config/network/wpa_supplicant.md
```diff
@@ -1 +1,103 @@
...
+First, you must create a configuration file for your interface:
+
+    $ cp /etc/wpa_supplicant/wpa_supplicant.conf /etc/wpa_supplicant/wpa_supplicant-<device>.conf
+
+### WEP Configuration
+
+For WEP configuratoin, you can just add the following lines to your
+device's `wpa-supplicant.conf` :
+
+    # Example default configuration file for wpa_supplicant.conf(5).
```
cvs-is-better: minor point, but this config file could probably be ordered the same as in the package: https://github.com/void-linux/void-packages/blob/master/srcpkgs/wpa_supplicant/files/wpa_supplicant.conf

src/config/network/dhcpcd.md
```diff
@@ -7,6 +7,12 @@ can enable the `dhcpcd` service.
 # ln -s /etc/sv/dhcpcd /var/service
 ```
 
+And then start it with [sv(8)](https://man.voidlinux.org/sv.8):
```
Duncaen: `sv start` only works if `runsv` for the directory is already running, which means the service is already running or there is a `down` file in the service directory.
So this is redundant, you don't need to start a service after activating it.

src/config/network/wpa_supplicant.md
```diff
@@ -1 +1,103 @@
...
+    ctrl_interface=/run/wpa_supplicant
+    update_config=1
+    ctrl_interface_group=wheel
+    eapol_version=1
+    ap_scan=1
+    fast_reauth=1
+    # Add your networks here.
+    network={
```
cvs-is-better: This is a duplicate of the WEP config, this should be the correct format

    network={
        ssid="MYSSID"
        #psk="YOUR AP KEY"
        psk=a8c34100ab4e5afac33cad7184d45a29ee0079001577d579bec6b74e4d7b5ac8
    }

src/config/network/wpa_supplicant.md
bobertlo: Good catch. This issue is a great candidate for a style guide, I will keep a note of that.

src/config/network/wpa_supplicant.md
bobertlo: I didn't double check that. These lines were from the wiki. Will do

src/config/network/dhcpcd.md
bobertlo: Oops, I'd been doing that this whole time haha. Thanks

the-maldridge:
Merged in 6608232

PR #12, opened 2018-11-14
contributing: add docs styleguide

bobertlo:
I took a stab at defining a style guide based on issues that have come up so far.

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -39,3 +39,6 @@
 - [Troubleshooting](./troubleshooting/index.md)
   - [XBPS](./troubleshooting/xbps/index.md)
 - [Contributing](./contributing/index.md)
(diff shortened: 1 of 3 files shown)


platerza:
Looking pretty good to me so far.

pltrz:
In my endeavour starting the chroot install guide I realized it might be a good idea to say something in the style guide about the grammatical tense. Like instructions should not be `We then format the filesystem...` but instead `Then, format the filesystem...` etc. I'm no expert on technical writing so I'm not sure what conventions we should stick with but it's something to think about.

the-maldridge:
In general the preference should be for the imperative voice of the present tense.  For example:

Format the Filesystem:
```
$ format c:
```

Install the base system:
```
$ copy *.* c:
```

PR #13, opened 2018-11-15
mirrors: squish links

bobertlo:
Change all links with urls printed as is to be <http://.../> format for maintainability and readability.


Vaelatern:
Sadly I don't understand how this is a change.

bobertlo:
Not that it comes up often, but when I had to change all the links in the install section, it was slightly annoying. It's mainly a style issue, I guess.

PR #15, opened 2018-11-19
Create initial chroot installation guide

pltrz:
This is an initial import for the (chroot) installation guide. There are many different install guides on the old wiki, for very specific use-cases, so I tried to keep all the examples simple and generic. I think effort duplication should be avoided.

Let me know what you guys think and what can be improved!


src/SUMMARY.md
```diff
@@ -8,6 +8,7 @@
     - [Downloading Images](./installation/liveimages/downloading.md)
     - [Preparing Installation Media](./installation/liveimages/prep.md)
     - [Installation Guide](./installation/liveimages/guide.md)
+    - [Installation Guide (chroot)](./installation/liveimages/guide-chroot.md)
```
bobertlo: I think this would be better in the Guides section.

src/installation/liveimages/guide-chroot.md
```diff
@@ -0,0 +1,388 @@
...
+```
+$ ls /sys/firmware/efi
+```
+
+If you know your system supports booting into UEFI mode, but does not appear to
+have done so, check your UEFI/BIOS settings (can usually be accessed by pressing
+a key like `F12` early in the boot process. Look for settings like *UEFI-mode*
+or *Disable Legacy BIOS mode*.
+
+### Connect to the Internet
```
CameronNemo: There is already a network configuration section.

src/installation/liveimages/guide-chroot.md
```diff
@@ -0,0 +1,388 @@
...
+Verify that you are connected to the internet:
+
+```
+ping voidlinux.org
+```
+
+Note that you can terminate a running command by sending a `SIGINT` signal with
+`ctrl-c`.
+
+### <a name="partition-disks"></a>Partition the disks
```
CameronNemo: Partitioning is really variable. I'm tempted to just have a separate partitioning section common to all install guides.

src/installation/liveimages/guide-chroot.md
```diff
@@ -0,0 +1,388 @@
...
+
+Create a password for the root user, and set the correct permissions.
+
+```
+# passwd root
+# chown root:root /
+# chmod 755 /
+```
+
+### Fstab
```
CameronNemo: maybe make a note about optionally adding a tmpfs?

src/installation/liveimages/guide-chroot.md
```diff
@@ -0,0 +1,388 @@
...
+...
+```
+
+Then, force a reconfiguration of the `glibc-locales` package:
+
+```
+# xbps-reconfigure --force glibc-locales
+```
+
+### Boot loader
```
CameronNemo: Same with the partitioning section, the boot loader can very a lot depending on hardware and preferences.

src/installation/liveimages/guide-chroot.md
```diff
@@ -0,0 +1,388 @@
...
+
+Install grub onto the EFI partition:
+
+```
+# grub-install --target=x86_64-efi --efi-directory=/boot/efi --bootloader-id=void-grub --boot-directory=/boot --recheck
+```
+
+## Reboot
```
CameronNemo: I could have sworn you have to generate the initramfs before rebooting, but maybe I'm wrong.

src/installation/liveimages/guide-chroot.md
bobertlo: I was thinking the same thing.

src/SUMMARY.md
pltrz: Good idea. Will move.

src/installation/liveimages/guide-chroot.md
pltrz: Good call. I'll remove the section and instead reference that.

src/installation/liveimages/guide-chroot.md
pltrz: Agreed. I think it's really important for users to understand the options they have and how to make the right choices. Which section should partitioning be moved to?

src/installation/liveimages/guide-chroot.md
pltrz: Sounds good. Btw, the tmpfs entry is already in the `fstab` I should differentiate it somehow from the entries I added

src/installation/liveimages/guide-chroot.md
CameronNemo: For now you could keep it here and somebody could move it on a second pass, or you could make a partitioning subsection in the install section.

src/installation/liveimages/guide-chroot.md
CameronNemo: If it is already there when you do a chroot installation, then you don't have to mention it I guess.

src/installation/liveimages/guide-chroot.md
CameronNemo: Well if you would like to add a section in your PR, that is fine too. But I think it may be easier to leave the partitioning sections as is and combine them with a second pass.

bobertlo: Let's go with that for now.

src/installation/liveimages/guide-chroot.md
pltrz: I've moved it to it's own page and referenced that in this guide. That page will then of course need expansion at some point but I agree users should have the information for all their use-cases.

src/installation/liveimages/guide-chroot.md
pltrz: Forgot that, I don't think it does generate one automatically. Will have to test that :thinking:  anyways, added that in.

src/installation/liveimages/guide-chroot.md
pltrz: Will resolve for now, until we comb it through again.

pltrz:
I'm going to close this as I messed up the branch by merging with void's master. I'll open a new pull request with the changes instead.

src/installation/liveimages/guide-chroot.md
pltrz: I just tested and it looks like it does in fact generate an initramfs. So I'll remove the section but it would be a good thing to add in another part of the docs

PR #17, opened 2018-11-20
Create chroot install guide, and bootloader page.

pltrz:
The bootloader page is info I gathered from the Arch and Void wikis, I'm not sure if I've put the info down correctly but as it was suggested that bootloader installation/configuration gets it's own page, it's a start.

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -5,11 +5,13 @@
 - [Installation](./installation/index.md)
    - [Base System Requirements](./installation/base-requirements.md)
    - [Live Installers](./installation/live-images/index.md)
+      - [Boot Loaders](./installation/live-images/boot-loaders.md)
       - [Downloading Images](./installation/live-images/downloading.md)
       - [Preparing Installation Media](./installation/live-images/prep.md)
       - [Partitioning Notes](./installation/live-images/partitions.md)
       - [Installation Guide](./installation/live-images/guide.md)
    - [Guides](./installation/guides/index.md)
+      - [Installation Guide (chroot)](./installation/guides/chroot-guide.md)
 - [Configuration](./config/index.md)
    - [Post Installation](./config/post-install.md)
    - [Services and daemons](./config/services/index.md)
@@ -48,4 +50,3 @@
 - [Contributing](./contributing/index.md)
    - [Void Docs](./contributing/void-docs/index.md)
       - [Style Guide](./contributing/void-docs/styleguide.md)
-
--- a/src/installation/guides/chroot-guide.md
+++ b/src/installation/guides/chroot-guide.md
@@ -1,0 +1,286 @@
+# Installation Guide (chroot)
+
+## Pre-installation
+
+### Set the keyboard layout
+
+The default console keymap is US. Available layouts can be listed with:
+
+```
+$ ls /usr/share/kbd/keymaps/*/*/*.map.gz | less
+```
+
+To modify the layout, append a corresponding file name to
+[loadkeys(1)](https://man.voidlinux.org/loadkeys), omitting path and file
+extension. For example, to set the Dvorak keyboard layout:
+
+```
+# loadkeys dvorak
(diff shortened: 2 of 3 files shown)


src/SUMMARY.md
```diff
@@ -10,6 +10,7 @@
     - [Partitioning Notes](./installation/liveimages/partitions.md)
     - [Installation Guide](./installation/liveimages/guide.md)
   - [Guides](./installation/guides/index.md)
+    - [Installation Guide (chroot)](./installation/guides/chroot-install.md)
```
CameronNemo: You haven't created the index for the boot loader section.

pltrz: Updated

CameronNemo:
Can you put the partitioning changes in a separate PR, and squash the small edits commit into one of the other commits?

pltrz:
> Can you put the partitioning changes in a separate PR, and squash the small edits commit into one of the other commits?

Yes, I'll get to that sometime tomorrow

src/installation/guides/chroot-guide.md
```diff
@@ -0,0 +1,286 @@
...
+# loadkeys dvorak
+```
+
+### Verify the boot mode
+
+Many new motherboards will support booting in UEFI mode, but some will not,
+instead booting in legacy BIOS mode. You will need to know in which mode your
+system will boot when it is time to partition the disks.
+
+If you have boot via UEFI mode, the directory `/sys/firmware/efi` should exist
```
the-maldridge: When in UEFI mode, the directory /sys/firmware/efi should be present.  If you believe you should be in EFI mode and this directory is not present, verify that the efivars filesystem is mounted.

pltrz:
@the-maldridge Understood. I'll restart and try again. Could you elaborate on some things you think I should be doing differently?

You mention:
> Its full of too many magic commands and no substance as to how to install the system and why each step is happening.

I'm having trouble understanding what you really mean and by extension how to fix it.

I didn't realize it was a bad guide, but I'm committed to getting it right. Let me know of any advice for the redo.

the-maldridge:
I think its a fine guide, but I don't think it is suitable for this repo.  What I mean by magic commands is that in several places it says to run some commands and doesn't really explain why you should run them, or what the end goal is.

I would much prefer a guide that walks through the phases of an install, with examples shown for each, but written in such a way that it can't be blindly copy/pasted, since that would encourage installation by magic, rather than by knowledge.  For reference, I am using sense 1 from here: http://www.catb.org/jargon/html/M/magic.html.  Nothing in the install should be too complicated to explain, and any guide which tries to walk through an install should be able to do so more abstractly.

A concrete example: rather than saying to use a specific partition layout, I would expect a well written guide to recommend characteristics of the disk such as a separate boot, the size of this boot partition, and a short discussion as to whether or not swap space is really necessary, and how much.

If you want to write a guide that's copy/paste/void then that can go in the guides/ subsection perhaps under the title of "Chroot install in N steps" where N is the number of steps.

pltrz:
Thanks @the-maldridge.

It's clearer now, I should be really diving into each section so that the reader is in the driver's seat and making informed decisions based on the guide, and the guide itself is really more general information into the options available at each step, so that the user can decide for himself, instead of just giving commands with brief context on what's happening (*magic*). In a sense it should really be more of a handbook. My understanding is superficial but I'm getting what you mean a bit more.

Thanks for following up; I see now that the guide should be something different than I originally thought. I expect you have a much clearer vision for how this guide should be than I do, and so for that reason I also expect there might be a few revisions until it's done right. I think this puts me more on the right track, though. I'll be starting the guide with a shift in objective.

the-maldridge:
Yeah, that's really what I'm intending, a repository of information that the reader can use to make decisions for themselves.  The guides/ section though is designed to contain copy/paste instructions for specific platforms if that's needed (think installing Void in places that don't use the installer, for example).

PR #18, opened 2018-11-21
mass autoformat

bobertlo:
I ran vmdfmt on every markdown file. Some needed to be fixed in order to process, but now this tree is stable with vmdfmt v0.0.1


bobertlo:
This may be helpful:
```
#!/bin/sh
autoformat() {
    echo $1
    vmdfmt "$1" > "$1.out"
    mv "$1.out" "$1"
}
find . -regex .*.md$ | while read file; do autoformat "$file"; done
```

PR #20, opened 2018-11-25
Try to clean up some of the installation documents

nilium:
This is an editing pass over the installation documentation. All
content should be semantically the same, but hopefully written in such
a way that it's a little bit of a better read.

Summary of content changes:

- Added a howtoread chapter. Currently just a single page, no
  sub-documents. Any decisions about formatting that communicate things
  to a user should be explained here.

- Removed quotes around most uses of "flavor" -- it's fine to direct
  attention to it once or twice, but it's really not necessary to put
  quotes around every reference to a flavor image.

- Split up the verifying integrity write-up into checksums / GPG so
  that it's in smaller chunks overall (should make it easier to know
  when the reader has made progress).

- Split up the prep.md USB / SD card section into two parts (identify
  your device and write the image). Also added example output to
  commands where possible.

### DIFF
--- a/src/installation/live-images/downloading.md
+++ b/src/installation/live-images/downloading.md
@@ -1,40 +1,53 @@
 # Downloading Images
 
-Live images can be downloaded from
-<https://alpha.de.repo.voidlinux.org/live/current/>.
+The most recent live images can be downloaded from
+<https://alpha.de.repo.voidlinux.org/live/current/>. Previous releases can be
+found under <https://alpha.de.repo.voidlinux.org/live/>, organized by date.
 
-### Verifying integrity
+## Verify Images
 
-The image release directories contain a `sha256sums.txt` and a
-`sha256sums.txt.asc` file to verify the integrity of the downloaded images.
+Each image releases's directory contains two files used to verify the image(s)
+you download. First, there is a `sha256sums.txt` file containing image checksums
+to verify the integrity of the downloaded images. Second is the
+`sha256sums.txt.sig` file, used to verify the authenticity of the checksums.
+
+We want to verify both the image's integrity and authenticity, so for this, we
+want to download both files:
 
 ```
 $ wget http://alpha.de.repo.voidlinux.org/live/current/sha256sums.txt{,.sig}
 ```
 
-You can now verify the integrity of downloaded file using
-[sha256sum(1)](https://man.voidlinux.org/sha256sum.1).
+### Verify Image Integrity
+
+You can verify the integrity of a downloaded file using
+[sha256sum(1)](https://man.voidlinux.org/sha256sum.1) with the `sha256sums.txt`
+file we downloaded above. The following sha256sum command will check (`-c`) the
+integrity of only the image(s) you've downloaded:
 
 ```
 $ sha256sum -c --ignore-missing sha256sums.txt
 void-live-x86_64-musl-20170220.iso: OK
 ```
 
-This just makes sure that the file was not corrupted while downloading.
+This verifies that the image is not corrupt.
 
-To verify that the downloaded files are the ones that the Void Linux maintainers
-published and signed you can use pgp.
+### Verify Image Authenticity
+
+To verify that the downloaded `sha256sums.txt` file is the one that the Void
+Linux maintainers published and signed, we use PGP. For this, we need the
+`sha256sums.txt.sig` downloaded above.
 
 The file is signed with the Void Images key:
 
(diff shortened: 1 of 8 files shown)


src/installation/live-images/downloading.md
```diff
@@ -59,3 +72,5 @@ gpg:          There is no indication that the signature belongs to the owner.
 Primary key fingerprint: CF24 B9C0 3809 7D8A 4495  8E2C 8DEB DA68 B482 82A4
 ```
 
+This verifies that the signature for the checksums is authentic. In turn, we can
+assume that the downloaded images are also authentic if their checksums match.
```
the-maldridge: s/assume/assert/

src/installation/live-images/guide.md
```diff
@@ -1,11 +1,11 @@
 # Installation Guide
 
-Once you have [downloaded](./downloading.md) your selected install image and
+Once you have [downloaded](./downloading.md) a Void image to install and
 [prepared](./prep.md) your install media, you are ready to install Void Linux.
 
-> Note: before you begin installation, you should make sure you understand which
-> type of boot process your system will use (BIOS or UEFI) and plan your
-> partitioning appropriately. (See the [partitioning notes](./partitions.md).
+> Note: before you begin installation, you should make sure you whether your
```
the-maldridge: s/make sure/determine/

src/installation/live-images/prep.md
```diff
@@ -1,27 +1,43 @@
-# Preparing installation media
+# Prepare Your Installation Media
```
the-maldridge: s/Your//

src/installation/live-images/prep.md
```diff
@@ -1,27 +1,43 @@
...
+After [downloading a live image](./downloading.md), it must be written to
+bootable media, such as a USB drive, SD card, or CD/DVD.
 
-## Create a bootable USB drive or SD card on Linux
+## Create a Bootable USB Drive or SD Card on Linux
 
-The `dd` command can be used to copy a live image to a storage device.
+### Identify Your Device
 
+Before we write the image, we need to identify the device we want to write it
```
the-maldridge: Please rephrase to be of the form "Before writing the image..."

src/installation/live-images/prep.md
```diff
@@ -30,16 +46,25 @@ Write the live image to the device:
...
 
-Ensure all data is flushed before disconnecting the device:
+dd won't print anything until it's completed (or if it failed), so depending on
+the device, this can take a few minutes or longer.
+
+Finally, ensure all data is flushed before disconnecting the device:
 
 ```
 $ sync
+90+0 records in
```
the-maldridge: This isn't the output from sync, please don't imply it is here.

src/installation/live-images/prep.md
```diff
@@ -30,16 +46,25 @@ Write the live image to the device:
...
+377487360 bytes (377 MB, 360 MiB) copied, 0.461442 s, 818 MB/s
 ```
 
-## Create a bootable USB drive or SD card on Windows
+The number of records, amount copied, and rates will all vary depending on the
+device and the live image you chose.
+
+## Create a Bootable USB Drive or SD Card on Windows
 
 The [USBWriter](https://sourceforge.net/projects/usbwriter/) program is the only
```
the-maldridge: I'm really not sure where this recommendation came from, I certainly don't recommend it and I maintain the images.  Lets either drop it entirely or come up with an actually recommended tool.

nilium:
Getting to these now that I'm off work.

src/installation/live-images/prep.md
nilium: This seems fine to me. I can't really make any recommendations for Windows, so I'm not sure where it came from. Given that there's the Windows Subsystem for Linux now, it may be feasible to just use dd on Windows 10 and later.

src/installation/live-images/prep.md
nilium: That wasn't supposed to be in that spot. My bad.

src/installation/live-images/prep.md
nilium: Rephrased as such. Not sure if it sounds too terse after fitting the sentence to that. Could rewrite it down to just the following if it reads better:

> Before writing the image, identify the device you'll write it to using [fdisk(8)](https://man.voidlinux.org/man8/fdisk.8). After connecting the storage device, identify the device path by running:

PR #25, opened 2018-12-03
about: change init from section 1 to 8

pltrz:
I believe `init(1)` on the *about* page should instead be `init(8)`. That's the section it is for runit's manpage and as far as I know for the traditional SysV-init as well.

Another thing that might be nice but I didn't change in this PR would be a link to the [wikipedia page for init](https://en.wikipedia.org/wiki/Init) or something similar. The passage mentions `init` as if the reader already knows what that means.

### DIFF
--- a/src/about/index.md
+++ b/src/about/index.md
@@ -24,7 +24,7 @@ and time again that they are dedicated to the security, quality, and
 maintainability of this critical library.
 
 A final distinguishing characteristic of Void is our choice of program for
-`init(1)`. Void Linux boots with smarden runit, a very small service supervision
+`init(8)`. Void Linux boots with smarden runit, a very small service supervision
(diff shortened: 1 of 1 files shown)


bobertlo:
I think a link to https://man.voidlinux.org/init.8 would be appropriate and consistent with the rest of the documentation.

pltrz:
Yes it would. Yet I can't help but think, though, that the way it's written the `init` here is not referring to *runit*-init specifically, but instead to `init` as in the program all unix-likes share as the program running under PID 1. I think it should either (a) be reworded to intend runit specifically, then we could link that manpage, otherwise (b) we should point to some documentation more generic to init as a concept rather than runit, specifically.

the-maldridge:
You are correct, in this case I am referring to init in the general case, not in the specific case of void.

PR #26, opened 2018-12-03
repositories: new subsection

bobertlo:
I added a page on non-free and restricted packages.


src/maintenance/packages/non-free.md
```diff
@@ -0,0 +1,37 @@
...
+If you are using a different mirror than the default, change the mirror in the
+configuration file:
+
+```
+/etc/xbps.d/10-repository-nonfree.conf
+```
+
+## Restricted packages
+
+There are some packages that are supported but do not allow redistribution of
```
Vaelatern: `but their licenses do not allow`

Vaelatern: And `A step beyond non-free, there are some packages`

pltrz:
Maybe this page could be more generic and cover how to add any of the other optional repositories like `multilib` and `debug`? Then as for the restricted packages you could just point to where `xbps-src`'s usage is explained. That way, we could avoid having extra pages for those teaching the user essentialy the same thing. Similar to what's on the old wiki [here](https://wiki.voidlinux.org/XBPS#Subrepositories).

bobertlo:
I thought about that, but also thought that nonfree packages is a high traffic question from new users. That was my thinking at least. I think debugging, nonfree, and multilib are diverse enough topics people are going to look for completely independently.

FWIW someone was complaining about this not being documented the other day on irc.

src/maintenance/packages/non-free.md
bobertlo: Good call

CameronNemo:
The book is searchable. So although it wouldn't be in the sidebar / table of contents, the information would still be fairly accessible.

pltrz:
I think if we keep the non-free page, we'll just end up with duplicate information. It would be:

- non-free page: explains how to add non-free subrepository and info on packages still not available without `xbps-src`
- subrepositories page (or similar): explains what other subrepositories are available and how to install them

That page doesn't exist now, but it will afterwards, which would make some of the contents of the non-free page redundant.

bobertlo:
The other page would be multilib, with a duplicated command to install the repo. 

I could pull this page, debugging, and write up multilib in one 'Subrepositories' page though, if that is too many pages for the three different topics. When I have a minute.

the-maldridge:
I like the direction this is going, but I do think this should be a separate page talking about each repo and then a page about how to add stuff.

I think it would also be valuable to include information on why things wind up in nonfree, as this is sometimes not obvious.

A nod to xbps-mini-builder might be worth it on the restricted page.

bobertlo:
So I should get back to this tonight after work. I am wondering if this should just be it's own section (maybe right after packages) called Sub-repositories with an overview of the different repos and a page on each. I don't want to start moving sections until there is some kind of plan. LIke:

5.3. Sub-repositories
   - 5.3.1. Debug
   - 5.3.2. Multilib
   - 5.3.3. Nonfree
   - 5.3.4. Restricted

the-maldridge:
Why not just have a sub section on repositories in general?  Then we could talk about installing from local files as well.

bobertlo:
I need to write a page on multilib still, and haven't thoroughly proofread this and will probably still make changes, but I wanted to put this up before I go to work for feedback. I'm starting to think the mirror section from packages should come in here also.

pltrz:
I think it's structured much more clearly now

Duncaen:
I don't think sub-repositories wording is the right word, they are not different from any other repository.

src/maintenance/repositories/restricted.md
```diff
@@ -0,0 +1,32 @@
+# Restricted packages
+
+A step beyond non-free, there are some packages that are supported but their
```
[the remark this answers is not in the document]
the-maldridge: Am maintainer, did restrict.

Packages can be restricted from distribution by either the upstream author or Void.  Void reserves the right to restrict distribution of any package for effectively any reason, massive size being the most common.

src/maintenance/repositories/mirrors.md
```diff
@@ -0,0 +1,21 @@
+# Selecting mirrors
+
+To substitute a regional mirror for one the official repositories, you may
+create a repository file in `/etc/xbps.d` with the same name as it's default
+file in `/usr/share/xbps.d` to override the default mirror.
+
+For example, check your installed repositories:
+
+```
+$ ls /usr/share/xbps.d
```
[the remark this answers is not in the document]
bobertlo: We need to overload the filenames. The other option I see is copying the directory contents entirely into /etc/xbps.d as is documented somewhere else.

src/maintenance/repositories/restricted.md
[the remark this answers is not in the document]
bobertlo: I didn't realize this, but it makes a lot of sense. I will probably borrow that language.

CameronNemo:
Additional sounds good!

bobertlo:
I just changed the name to Official Repositories on the side bar, because it looks better over there IMO. I was going to contrast with a 'Custom Repositories' section and split the adding repositories page between them.

src/maintenance/repositories/mirrors.md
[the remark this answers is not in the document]
bobertlo: I didn't realize this was already written up in another section. I missed some commits.

bobertlo:
I think I have consolidated all the changes. I want to wait on reviews now and hopefully get this merged. This is getting to be a pretty big chunk again.

bobertlo:
I did touch these a couple times, to fix spelling and Header Casing. I just don't want to make any more editorial changes without feedback.

src/SUMMARY.md
```diff
@@ -39,9 +39,19 @@
    - [Packages](./maintenance/packages/index.md)
       - [General](./maintenance/packages/management.md)
       - [Building](./maintenance/packages/building.md)
-      - [Debugging](./maintenance/packages/debugging.md)
-      - [Mirrors](./maintenance/packages/mirrors.md)
       - [Finding files](./maintenance/packages/files.md)
+   - [Repositories](./maintenance/repositories/index.md)
+      - [Official Repositories](./maintenance/repositories/official/index.md)
+         - [Installing
```
the-maldridge: Is this formatted correctly?

src/maintenance/repositories/custom.md
```diff
@@ -0,0 +1,24 @@
+# Custom Repositories
+
+Void supports user created repositories available either locally, or remotely.
```
the-maldridge: Maybe add some note here about packages not part of the main tree can seriously bork your system and there are very real security questions to answer when trusting a non-official repo?

src/maintenance/repositories/index.md
```diff
@@ -0,0 +1,4 @@
+# Repositories
+
+Repositories are the heart of the xbps package system. Repositories can be
+locally or remotely available.
```
the-maldridge: Maybe define what a repository is in the sense of xbps here?  i.e. that it contains a -repodata file?

src/maintenance/repositories/mirrors/changing.md
```diff
@@ -0,0 +1,22 @@
+# Changing Mirrors
+
+Copy all repository configuration files from `/usr/share/xbps.d` to
```
the-maldridge: This is a [magic](http://www.catb.org/jargon/html/M/magic.html) instruction and needs to be rewritten to explain why this is being done.

src/maintenance/repositories/mirrors/changing.md
```diff
@@ -0,0 +1,22 @@
...
+`/etc/xbps.d` and then change the repository URLs in `/etc/xbps.d`.
+
+```
+# mkdir -p /etc/xbps.d
+# cp /usr/share/xbps.d/*-repository-*.conf /etc/xbps.d/
+# sed -i 's|https://alpha.de.repo.voidlinux.org/current|<repository>|g' /etc/xbps.d/*-repository-*.conf
+```
+
+You can use `xbps-query` to verify that all repositories are changed to the
```
the-maldridge: I haven't tested to be sure, but I'm pretty sure you need an -S before the query will show the new repos.

src/maintenance/repositories/official/index.md
```diff
@@ -0,0 +1,13 @@
+# Official Repositories
+
+Void provides other official repositories, which are maintained by the Void
+project, but not installed in the default configuration:
+
+- [debug](./debug.md): contains debugging symbols for packages
+- [multilib](./multilib.md): contains 32-bit libraries for 64-bit systems (glibc
+   only)
+- [multilib](./multilib.md)/nonfree: contains non-free multilib packages
```
the-maldridge: Intentional that this link looks like this?

src/maintenance/repositories/official/installing.md
```diff
@@ -0,0 +1,12 @@
+# Installing Repositories
+
+Void provides official [sub-repositories](./sub-repositories/index.md) that may
```
the-maldridge: I think it was discussed that they aren't sub- anything.  Please change this language.

src/maintenance/repositories/official/multilib.md
```diff
@@ -0,0 +1,9 @@
+# multilib
+
+The `multilib` repository provides 32-bit packages as a compatibility layer
+inside a 64-bit system and is available through the package
+`void-repo-multilib`. The repository also has a sub-repository containing
```
the-maldridge: again, language change for sub-

src/maintenance/repositories/official/nonfree.md
```diff
@@ -0,0 +1,11 @@
+# nonfree
+
+Void has a `nonfree` repository for packages that don't have free licenses.
+Install the `void-repo-nonfree` package to enable this repository.
+
+Packages can end up in the `nonfree` repository for a number of reasons:
```
the-maldridge: This is fine for now, but it would be really nice to have a policy document for what winds up in what repo.  Maybe @maxice8  would be willing to shepherd a PR through to void-packages for repository policy.

src/maintenance/repositories/restricted.md
```diff
@@ -0,0 +1,37 @@
...
+
+Packages can be restricted from distribution by either the upstream author or
+Void. Void reserves the right to restrict distribution of any package for
+effectively any reason, massive size being the most common. The other most
+common reason is restrictive licensing that does not allow third-party
+redistribution of source or binary packages.
+
+## Building manually
+
+This feature must be enabled explicitly. Run the following in the
```
the-maldridge: Magic command, needs explanation.  Probably also shouldn't be here, but rather a pointer to the corresponding documentation in void-packages.

src/maintenance/repositories/custom.md
bobertlo: Yes, I had been trying to strongly imply that you would create these yourself. I will look into that.

src/maintenance/repositories/official/installing.md
bobertlo: Missed this one. Oops.

src/maintenance/repositories/restricted.md
bobertlo: I would like to thing that the audience could interperet a FLAG=value in a config file. I could break it down further.

src/maintenance/repositories/official/nonfree.md
bobertlo: You guys figure this out.

src/maintenance/repositories/official/multilib.md
bobertlo: My thinking was that the non-free multilib repository depends on the multilib repository.

bobertlo:
For anyone interested the formatting is failing on the removal of files. The current travis is trying to format removed files and creating empty files.

src/maintenance/repositories/mirrors/changing.md
bobertlo: This is correct, btw.

src/maintenance/repositories/official/index.md
bobertlo: Yes, but I will rework how the multilib/nonfree repository is represented.

src/maintenance/repositories/official/nonfree.md
bobertlo: I didn't mean to be rude here, and I'm willing to help with this, but it is way above my head specifying how packages end up in each repo!

src/SUMMARY.md
pltrz: Is it because of the 80 character limit? If so, perhaps we could change the limit for this particular file, since it's full of single-lines anyways, and more than likely we'll get more lines that exceed the 80 characters in this file in the future.

src/maintenance/repositories/restricted.md
```diff
@@ -0,0 +1,37 @@
...
+These packages are marked as restricted and must be built from their
+[void-packages](https://github.com/void-linux/void-packages) template locally.
+
+Packages can be restricted from distribution by either the upstream author or
+Void. Void reserves the right to restrict distribution of any package for
+effectively any reason, massive size being the most common. Another common
+reason is restrictive licensing that does not allow third-party redistribution
+of source or binary packages.
+
+## Building manually
```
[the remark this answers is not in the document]
bobertlo: Oh, I thought I got through all the notes last time, oops. I'm referring people to the void-packages documentation, but still mentioning the `XBPS_ALLOW_RESTRICTED` flag, since it's not very obvious at all in the documentation, but removed the command line to do it .

bobertlo:
I made these changes

the-maldridge:
Looks good.  I'll lint manually and then merge.

PR #31, opened 2018-12-08
various: bulk-process with vmdfmt

pltrz:
I noticed that most files end with a blank line, but some don't. I'm not sure if in our style we wan't them all to have the extra line or none of them, but `vmdfmt` at the current state says no to them, so I bulk-processed all the files with it using:

```sh
for i in $(find src/)
do
    vmdfmt -w $i
done
```


bobertlo:
I just didn't want to submit another change to almost every file again myself. This had bugged me since I can't format the whole tree. I missed the newline bug on my first version.

Btw you can `vmdfmt -w -l src/` to format recursively and list what changed.

pltrz:
> Btw you can `vmdfmt -w -l src/` to format recursively and list what changed.

Didn't realize it ran recursively; nice!

the-maldridge:
Please amend your PR to fix travis to use the new vmdfmt invocation.

pltrz:
@the-maldridge let me know if this is what you mean:

I changed

```sh
while read -r file ; do
    printf 'Formatting %s\n' "$file"
    vmdfmt "$file" > "$file.1"
    mv "$file.1" "$file"
done < /tmp/changed_files
```

to

```sh
while read -r file ; do
    printf 'Formatting %s\n' "$file"
    vmdfmt -w "$file"
done < /tmp/changed_files
```

bobertlo:
Looks good to me.

I tested against the issue I've had where removing files causes CI to fail and this fixes that also.

The other option would be not tracking changed files and just running `vmdfmt -w -l src/` and check the entire tree. This would be simpler, but would then require us to not merge anything that the formatter does not like, or every PR will fail CI.

the-maldridge:
needs rebase.

PR #42, opened 2019-01-23
network/wpa_supplicant.md: Document only how to run wpa_supplicant under

maxice8:
runit.

### DIFF
--- a/src/config/network/wpa_supplicant.md
+++ b/src/config/network/wpa_supplicant.md
@@ -12,15 +12,7 @@ configuration files.
 [wpa_cli(8)](https://man.voidlinux.org/wpa_cli.8) are frontends to manage the
 configuration files and the running service.
 
-## Configuring wpa_supplicant
-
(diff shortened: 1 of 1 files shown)


bobertlo:
I think dhcp hooks are at least worth mentioning, since the base install has dhcpcd enabled, which invokes wpa_supplicant hooks by default so a fresh install will be running wpa_supplicant that way on wireless interfaces.

maxice8:
They will be removed in dhcpcd 7.1.0

On Jan 24, 2019 16:36, "Robert Lowry" <notifications@github.com> wrote:

I think dhcp hooks are at least worth mentioning, since the base install
has dhcpcd enabled, which invokes wpa_supplicant hooks by default so a
fresh install will be running wpa_supplicant that way on wireless
interfaces.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/void-linux/void-docs/pull/42#issuecomment-457307946>,
or mute the thread

PR #44, opened 2019-01-24
config/shell.md: change oksh to loksh

### LINKED ISSUE #43: shells: `loksh` and `oksh`
In the section on shells, `oksh` is listed, however on my system I get a message stating `oksh` is being removed from the repos. It seems to have been superseeded by `loksh`, which is not listed in the docs. This suggests to me that the docs are slightly outdated in this regard. 

I have tried to create a PR to fix this, but something is broken along the line and it doesn't seem to be working.

bobertlo:
fixes #43

### DIFF
--- a/src/config/shell.md
+++ b/src/config/shell.md
@@ -23,7 +23,7 @@ $ chsh -l
 /bin/zsh
 /bin/rc
 /bin/ksh
-/bin/oksh
+/bin/loksh
 /bin/yash
 ```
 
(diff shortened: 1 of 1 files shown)


src/config/shell.md
```diff
@@ -38,7 +38,6 @@ shells.
 - heirloom-sh
 - ksh
 - mksh
-- oksh
```
maxice8: shouldn't be replaced by loksh ?

bobertlo: I am so bad at git sometimes. Thanks!

bobertlo:
Fixed

PR #45, opened 2019-01-28
pulseaudio.md: added info for starting pulseaudio for x sessions.

kartik-ynwa:


Vaelatern:
It's not necessary to do so, because pulse is autostarted.

kartik-ynwa:
You are correct. Sorry about this.

Vaelatern:
6541dd7
I added an explanation so other's don't think as you did.

PR #47, opened 2019-01-30
mirrors: add lysator mirror

andkem:
This is a PR to add Lysator's FTP as a void-linux mirror.

### DIFF
--- a/src/maintenance/repositories/mirrors/list.md
+++ b/src/maintenance/repositories/mirrors/list.md
@@ -12,11 +12,13 @@
 
 ## Tier 2 mirrors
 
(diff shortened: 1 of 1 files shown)


andkem:
I might add that I'm planning to write a wiki-page on how to use xbps with a hidden service. Since xbps supports setting a proxy via the SOCKS_PROXY environment variable, one can use it out of the box with an onion by simply changing the addresses under /usr/share/xbps.d and pointing it to Tor.

Vaelatern:
Please contribute such documentation to https://github.com/void-linux/void-docs

PR #48, opened 2019-02-01
mirrors: add tor instructions

andkem:
Adds instructions on using XBPS with Tor.

### DIFF
--- a/src/maintenance/repositories/mirrors/tor.md
+++ b/src/maintenance/repositories/mirrors/tor.md
@@ -1,0 +1,98 @@
+# Using Tor mirrors
+
+Tor is an anonymising software that bounces traffic via computers all around the
+world. It can provide access to regular sites on the internet or to hidden sites
+only available on the network.
+
+## Using XBPS with Tor
+
+XBPS can be made to connect to mirrors using Tor. These mirrors can be normal
+mirrors, via exit relays, or, for potentially greater anonymity, hidden service
+mirrors on the network.
+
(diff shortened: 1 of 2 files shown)


src/maintenance/repositories/mirrors/tor.md
```diff
@@ -0,0 +1,94 @@
...
+or as a system service:
+
+```
+# ln -s /etc/sv/tor /var/service
+```
+
+By default, Tor will act as a client and open a SOCKS5 proxy on TCP port 9050 on
+localhost.
+
+### Making XBPS connect via the SCOKS proxy
```
Vaelatern: SOCKS

andkem: Fixed

src/maintenance/repositories/mirrors/tor.md
```diff
@@ -0,0 +1,94 @@
...
+### Security consideration
+
+It is advisable to set `SOCKS_PROXY` automatically in your environment if you
+are using an onion. If the setting is missing, a DNS query for the name of the
+hidden service will leak to the configured DNS server.
+
+To automatically set the environment variable, add it to `/etc/profile`:
+
+```
+# echo "export SOCKS_PROXY=\"socks5://127.0.0.1:9050\"" >> /etc/profile
```
the-maldridge: I'd prefer if this went into a file in /etc/profile.d/

andkem: Is the pushed change acceptable? I don't know how easy the here document syntax is to read.

PR #53, opened 2019-02-05
maintenance/repositories: add signing repositories

Duncaen:


src/maintenance/repositories/signing.md
```diff
@@ -0,0 +1,23 @@
+## Signing repositories
+
+Remote repositories are **must** be signed, for local repositories this is not
```
CameronNemo: typo "[repos] are must be"

Vaelatern: If we are here making changes, `local repositories do not need to be signed` is mildly more idiomatic

src/maintenance/repositories/signing.md
```diff
@@ -0,0 +1,23 @@
...
+
+The [xbps-rindex(1)](https://man.voidlinux.org/xbps-rindex.1) tool is used to
+sign repositories. First initialize the repository metadata with signing
+properties (this is only required once).
+
+```
+$ xbps-rindex --sign --signedby "I'm Groot" /path/to/repository/dir
+```
+
+Afterwards sign one or more packages with the following command, newly added
```
bobertlo: Afterwards, ...

And, if I am understanding this correctly, it maybe more clear to say:

... with the following command:
```
...
```

Any packages added in the future will also need to be signed.

Vaelatern: Don't quite agree...

Vaelatern: Actually, this makes sense, saying "Remember that this only signs the packages you already have. Newly made packages, or changed packages, will not be automatically signed."

bobertlo:
Minus the comma, with or without the change of wording, this looks good to me.

PR #54, opened 2019-02-05
Void Theme

Duncaen:
Screenshot


the-maldridge:
What about an actual void theme?

Duncaen:
I tried to add the void navigation to a `header.hbs`, but this doesn't work nice with the upstream stylesheets.
Only alternative would be to divert from their `index.hbs` template, but its rather large. Not sure if its a good idea to maintain a new copy.

Duncaen:
I'm currently playing around with a simplified template which looks like any other void page, I removed theme support and all javascript except for search.

Nothing is fixed/sticky except the chapter navigation arrows if there is enough space.
Sidebar and content share a scroll bar for simplicity.

I think the changes help with both usability and also maintainability.

screenshot 1
screenshot 2

Duncaen:
This PR now contains the theme:

2019-02-06-182724_1500x1000_scrot
2019-02-06-182751_1501x999_scrot
2019-02-06-182847_536x1000_scrot

the-maldridge:
Looks good to me.  I'd like a second person to shipit before merge though.  @Gottox thoughts?  I know a while back you were looking at other theme stuff.

Gottox:
lgtm to me! :+1:

Issue #50, opened 2019-02-04
NetworkManager


bobertlo:
The NetworkManager page is empty. It should be written or removed.

the-maldridge:
I'm inclined to remove.  Writing it would require non-trivial information about things like nmcli and nmtui.

PR #56, opened 2019-02-07
network: write NetworkManager page, and small fixes

bobertlo:


bobertlo:
#50

bobertlo:
@the-maldridge just waiting on a quick look from you, based on your input on #50

src/config/network/networkmanager.md
```diff
@@ -1 +1,38 @@
...
+
+[NetworkManager(8)](https://man.voidlinux.org/NetworkManager.8) is a daemon that
+manages network connections on Ethernet, WiFi, and mobile broadband devices and
+operates as an all-in-one solution to network management. The `NetworkManager`
+package contains the basic utilities to run
+[NetworkManager(8)](https://man.voidlinux.org/NetworkManager.8).
+
+## Starting NetworkManager
+
+To enable the [NetworkManager(8)](https://man.voidlinux.org/NetworkManager.8)
```
nilium: Maybe remove the man page links (and the `(8)`) after the first paragraph? I'm not sure where things stand on style there, but it's a lot of repeating the same link.

Vaelatern: I'm happy with repeated links, once they are in there.

bobertlo: I've just been spamming them to reinforce that the reader should RTFM lol. Here it's probably a little too much. I'd been pretty much using man links and code blocks whenever talking about any program or package.

I'll take some out, but I really don't want a reader to have to scroll up or look into a previous section for a link to a man page either and I fully intend to keep spamming as much as I can pass as acceptable. ;)

nilium: Not saying you need to remove them. Just seems like overkill to me, so if there’s agreement on linking all instances, then let’s keep it.

bobertlo: I like goofy things like this overkill if it makes policy simpler and reminds people to read man pages.

Issue #51, opened 2019-02-04
Definition of scope


bobertlo:
There is no definition of the scope of these documents, as was discussed on IRC. What should be included and what should not?

There seemed to be an agreement that these documents should be a manual for the setup and administration of void systems, and generally excluding how to use upstream software aside from documenting idiosyncrasies presented by void.

I think it would be useful to formalize this. Specifically with a definition of scope near the style guide and probably a less formal note near the how to read section explaining that we are not rewriting upstream application documentation, nor are we compiling a collection of links to said documentation which is easily found by the user.

This will be especially important if we make an open call for contributors. Those of us who have contributed seem to have similar use cases for void and similar opinions on the docs, but new users who can document software we aren't knowledgeable with may have very different ideas of what this manual is.

My goal with this issue is not to define policy, but to open a discussion.

bobertlo:
#62 added an informal description in 'howtoread/index.md'.

I'm leaving this open for now to think about possible style guide additions. I don't see a pressing need to address anything at the moment.

bobertlo:
I think this has been addressed properly for now.

PR #62, opened 2019-02-08
howtoread/index.md: explain scope of the handbook

bobertlo:
This is a follow-up to the issue #51

I would sincerely like input on this. I hope this is not too opinionated or in conflict with anyone else's opinions.

### DIFF
--- a/src/howtoread/index.md
+++ b/src/howtoread/index.md
@@ -1,6 +1,24 @@
 # How To Read The Handbook
 
-## Examples
+This handbook is not an extensive guide on how to use and configure common Linux
+software. The purpose of this document is to explain how to install, configure,
+and maintain Void Linux systems, and to highlight the differences between common
+Linux distributions and Void.
+
+Those looking for tips and tricks on how to configure a Linux system in general
+should consult upstream software documentation. Additionally, the [Arch
+Wiki](https://wiki.archlinux.org/) provides a fairly comprehensive outline of
+common Linux software configuration, and a variety of internet search engines
+are available for further assistance.
+
+## Reading The Manuals
+
+While this handbook does not provide a large amount of copy and paste
+configuration instructions, it does provide links to the [man
+pages](https://man.voidlinux.org/) for the referenced software wherever
+possible.
+
+## Example Commands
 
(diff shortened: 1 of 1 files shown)


src/howtoread/index.md
```diff
@@ -1,6 +1,24 @@
 # How To Read The Handbook
 
-## Examples
+This handbook is not an extensive guide on how to use and configure common Linux
+software. The purpose of this document is to explain how to install, configure,
+and maintain Void Linux systems, including the differences from other common
```
Vaelatern: `Focusing on`, perhaps?

src/howtoread/index.md
```diff
@@ -1,6 +1,24 @@
...
+Linux distributions.
+
+Those looking for tips and tricks on how to configure a Linux system in general
+should consult upstream software documentation. Additionally, the [Arch
+Wiki](https://wiki.archlinux.org/) provides a fairly comprehensive outline of
+common Linux software configuration, as well as the
+[google](https://www.google.com) search engine.
```
Vaelatern: `and a variety of internet search engines exist for further assistance` maybe.

src/howtoread/index.md
```diff
@@ -1,6 +1,24 @@
 # How To Read The Handbook
 
-## Examples
+This handbook is not an extensive guide on how to use and configure common Linux
+software. The purpose of this document is to explain how to install, configure,
+and maintain Void Linux systems, focusing on the differences from other common
```
nilium:
```suggestion
and maintain Void Linux systems, and to explain differences between common Linux distributions and Void.
```

The wording of this seemed a little off.  Maybe explain is the wrong word too. Highlight? Cover? Something like that. I think what I'm trying to get at is that it's not that this guide focuses on the differences, just that there will be differences and this guide documents them.

bobertlo: I did feel iffy about the wording here. I will come up with something better. I always appreciate your reviews and this is very helpful :)

bobertlo:
@nilium I made those changes. I'm not sure this is the perfect wording, but much clearer.

If there are no other comments, I think this is ready to go.

nilium:
I have nothing to add — seems fine to me.

Issue #57, opened 2019-02-07
CUPS pages overlap


bobertlo:
Looking at the PR #41 it seems that the CUPS pages (config/install) are overlapping. I don't think that all of this information belongs on one page, but restructuring this section would be very helpful. There seems to be installation and configuration information mixed between both pages and, in some cases, duplicated.

CameronNemo:
It could go on one page maybe. Indeed it could use a little restructuring. Currently the model is stage 1 install everything you might want, stage 2 configure what you installed. Probably not the best way to go about it.

bobertlo:
I agree, I could have qualified that statement. I didn't mean these necessarily need to be merged into one page.

bobertlo:
Closed by #64

PR #41, opened 2019-01-23
cups: add how-to setup service

go2null:

PR #64, opened 2019-02-08
config/print: split up information more logically

CameronNemo:
Should address #41 and #57


src/config/print/config.md
```diff
@@ -30,15 +17,31 @@ configure a printer using the command line.
...
+
+```
+# xbps-install -S cups-pk-helper
+```
+
+You may run `system-config-printer` from your desktop environment's application
+launcher, or via the command line:
+
+```
+$ system-config-printer
```
bobertlo: This should probably just be `# system-config-printer`. I'd like this a lot better if it was shown as running with root, followed by a note about the need to run as root, and the polkit exceptions and a mention of the `cups-pk-helper` package.

src/config/print/config.md
```diff
@@ -30,15 +17,31 @@ configure a printer using the command line.
...
+```
+
+If you are not using the `polkit` helper mentioned above, the command must be
+invoked as root. Various packages can be used invoke GUI applications as root,
+including `lxqt-sudo` and `kdesu`.
```
bobertlo: I think this is slightly outside of the scope of this page. This handbook has so far been assuming basic proficiency of being able to launch both user and superuser commands from a terminal, and we might be better off sticking to that assumption here.

src/config/print/config.md
```diff
@@ -30,15 +17,31 @@ configure a printer using the command line.
...
 ```
 # xbps-install -S system-config-printer
-# system-config-printer
 ```
 
-This command does not need to be run as root if PolicyKit is in use and
-`cups-pk-helper` is installed.
+Normally this tool would require root privileges to configure printers, but
+unprivileged access through `polkit` can be achieved with a helper. If you are
+interested in this, install the helper:
```
bobertlo: if you are using PolicyKit, you can install the `cups-pk-helper` package to allow unprevileged users access to CUPS configuration.

src/config/print/config.md
CameronNemo: Yeah I waffled a bit on whether to use root or user in the example. Went with user because it is typically bad form to run GUI apps as root. I will change it though. It is just an example, and as you mention below the audience will have some level of basic proficiency.

src/config/print/config.md
CameronNemo: Well one note I would have here is that the CLI tools and web interface would still need root. Just system-config-printer and some other GUI tools take advantage of the policykit helpers.

PR #65, opened 2019-02-08
config/bluetooth: add bluetooth page

pltrz:
I've started on the bluetooth page. Feedback appreciated.

Some extra thoughts:

1. In order for unpriviledged users to use `bluetoothctl`, they must be in either the `bluetooth` group or the `lp` group. The Void Wiki says to use the former while the Arch Wiki says to use the latter. I really don't know much about these two groups and by extension which would be the correct one to suggest to users.

Neither the Arch or Debian wikis mention the `bluetooth` group, but both briefly mention the `lp` group.

[Arch](https://wiki.archlinux.org/index.php/users_and_groups#System_groups):
> Access to parallel port devices (printers and others).

[Debian](https://wiki.debian.org/SystemGroups):
> Members of this group can enable and use printers. (The user lp is not used anymore.)

-------------------------------------------

2. I included the following:

> Note: To use an audio device such as a wireless speaker or headset, you must also use the PulseAudio sound server.

Though, this is only an assumption I made based on my limited experience. If anyone knows whether you can use bluetooth audio devices with `sndio` or plain `ALSA` then please chime in!

### DIFF
--- a/src/config/bluetooth.md
+++ b/src/config/bluetooth.md
@@ -1,0 +1,86 @@
+# Bluetooth
+
+Ensure the bluetooth controller is not blocked. Use `rfkill` to check whether
+there are any blocks and to remove soft blocks. If there is a hard block, there
+is likely either a physical hardware switch or an option in the BIOS to enable
+the bluetooth controller.
+
+```
+$ rfkill
+ID TYPE     DEVICE      SOFT      HARD
+0 wlan      phy0   unblocked unblocked
+1 bluetooth hci0     blocked unblocked
+
+# rfkill unblock bluetooth
+```
+
+## Installation
+
+Install the `bluez` package.
+
+```
+# xbps-install -S bluez
+```
+
+Enable the `bluetoothd` and `dbus` services.
+
+```
+# ln -s /etc/sv/dbus /var/service
+# ln -s /etc/sv/bluetoothd /var/service
+```
+
+Add your user to the `bluetooth` group and restart the `dbus` service, or simply
+reboot the system. Note that restarting the `dbus` service may kill processes
+making use of it.
+
+```
+# usermod -a -G bluetooth $USER
+# sv restart dbus
+```
+
+> Note: To use an audio device such as a wireless speaker or headset, ALSA users
+> need to install the `bluez-alsa` package, while
+> [PulseAudio](./media/pulseaudio.md) users do not need any additional software.
+
+## Usage
+
+Manage bluetooth connections and controllers using `bluetoothctl`. It uses a
+command line interface; to find out what commands are available, enter `help`.
+To exit, enter `exit` or `quit`.
+
+```
+$ bluetoothctl
+
+[bluetooth]# help
+Menu main:
+Available commands:
+-------------------
+advertise             Advertise Options Submenu
+scan                  Scan Options Submenu
+gatt                  Generic Attribute Submenu
+list                  List available controllers
+show [ctrl]           Controller information
+select <ctrl>         Select default controller
+devices               List available devices
+paired-devices        List paired devices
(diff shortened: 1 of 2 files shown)


Vaelatern:
2. is wrong.
```[*] bluez-alsa-1.3.1_6              Bluetooth Audio ALSA Backend```

**Comment by pltrz** [comment]
> 1. is wrong.
>    `[*] bluez-alsa-1.3.1_6              Bluetooth Audio ALSA Backend`

Thanks for the clarification @Vaelatern.

**Comment by bobertlo** [comment]
The group is system specific. A `bluetooth` group is created by the `bluez` package, and it has a `post-install` message: "You need to be in the bluetooth group to be able to modify the bluetooth state!" So, I think we should stay consistent with that.

I don't know if/why the `lp` group allows control of bluetooth devices, but that group is indeed meant for printing and parallel port access. I do not think we should document it that way. :)

File: src/config/bluetooth.md [hunk 1]
```diff
@@ -0,0 +1,84 @@
...
+devices               List available devices
+paired-devices        List paired devices
+...
+```
+
+Consult the [Arch Wiki](https://wiki.archlinux.org/index.php/Bluetooth#Pairing)
+page for an example on how to pair a device.
+
+> Note: To use an audio device such as a wireless speaker or headset, you must
+> additionally use either the [PulseAudio](./media/pulseaudio.md) sound server
```
bobertlo: "... ALSA users need to install the `bluez-alsa` package, while PusleAudio users do not need additional software."

pltrz:
> The group is system specific. A `bluetooth` group is created by the `bluez` package, and it has a `post-install` message: "You need to be in the bluetooth group to be able to modify the bluetooth state!" So, I think we should stay consistent with that.
> 
> I don't know if/why the `lp` group allows control of bluetooth devices, but that group is indeed meant for printing and parallel port access. I do not think we should document it that way. :)

Thanks @bobertlo, this is very helpful to know!

src/config/bluetooth.md
```diff
@@ -0,0 +1,84 @@
...
+```
+
+Enable the `bluetoothd` and `dbus` services.
+
+```
+# ln -s /etc/sv/bluetoothd /var/service
+# ln -s /etc/sv/dbus /var/service
+```
+
+Add your user to the `lp` group and restart the `dbus` service.
```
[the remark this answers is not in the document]
pltrz: I could also say either reboot entirely or restart dbus, and if you choose to restart dbus it might have this unintended side-effect? I'll write that out so you can see how it might look

src/config/bluetooth.md
```diff
@@ -0,0 +1,86 @@
...
+
+```
+# xbps-install -S bluez
+```
+
+Enable the `bluetoothd` and `dbus` services.
+
+```
+# ln -s /etc/sv/bluetoothd /var/service
+# ln -s /etc/sv/dbus /var/service
```
bobertlo: Small thing, but shouldn't dbus be started before bluetoothd?

bobertlo:
I just had that last little thing I saw. Otherwise, this looks good to me. Thanks for this content :)

src/config/bluetooth.md
pltrz: Yes it should. Updated

pltrz:
Now I'm wondering if the PulseAudio/ALSA blurb might be more appropriate under the Installation header?

bobertlo:
Go for it that seems like a good idea. I finally got out my bluetooth headphones to try and test this quick also :)

bobertlo:
I'm not seeing any changes in your commits?

pltrz:
my mistake

PR #66, opened 2019-02-08
config/shell: add some available shells

pltrz:


src/config/shell.md
```diff
@@ -32,15 +32,18 @@ shells.
 
 - bash
 - dash
+- dreampie
```
Vaelatern: No

src/config/shell.md
```diff
@@ -32,15 +32,18 @@ shells.
...
 - elvish
 - es
 - fish-shell
 - heirloom-sh
+- ion
```
Vaelatern: Ok

src/config/shell.md
```diff
@@ -32,15 +32,18 @@ shells.
...
 - ksh
-- mksh
 - loksh
+- mksh
 - posh
 - rc
 - tcsh
+- xonsh
```
Vaelatern: I guess

src/config/shell.md
Vaelatern: This is not an available shell.

pltrz: My mistake. I Just went and installed all the shells I could find and looked at which ones added entries to /etc/shells. This one does not though.

pltrz: Removed it in the latest push

PR #67, opened 2019-02-09
theme/index.hbs revert nav bar to sync with the rest of voidlinux.org

bobertlo:


Duncaen:
I would prefer to remove forum and add the docs to the rest of voidlinux.org.

bobertlo:
I think you should file a request on void-linux.github.io then. I thought the goal was to replace the wiki though. This is admittedly over my head.

bobertlo:
I just mean it should be consistent. Ill just keep this open if you are working on that

pltrz:
I think you are both right. It should be updated to reflect the way @Duncaen describes but in the mean time it really should have taken @bobertlo's change for the sake of consistency. This is kind of "consumer-facing" after all.

bobertlo:
I'm pulling this because of the report on reddit about this.

PR #68, opened 2019-02-13
A trick to solve issue with groups in per-user services

ametisf:
Hi. I'm not sure how much wiki-like you want these docs to be (there are not many pull requests yet), so if this doesn't fit the style feel free to throw this away.

This is just a trick I use to make my user services use all of my users groups.

Also, why do you use the UID variable name for user name? Usually USER variable is used for text username and UID is for users numerical id. Here I used intermediate variable username to try and prevent any colisions.

### DIFF
--- a/src/config/services/user-services.md
+++ b/src/config/services/user-services.md
@@ -26,3 +26,15 @@ One important caveat: if any services you have need group permissions instead of
 just your user permissions, you will want to append those groups in a colon
 separated list to your username, such as `/etc/sv/anon:void1:void2:void3/run`
 instead of just `/etc/sv/anon/run`.
+
+If you want to append all users groups as in `/etc/group`, groups utility can
+solve this for you:
+
+```
+username="$UID"
+
+# GNU coreutils
+usergroups="$( groups "$username" | tr -s ' ' ':' )"
+# other implementations
+usergroups="$UID:$( groups "$username" | tr ' ' ':' )"
+```


Duncaen:
This is broken, this would result in `$username`'s primary group to be used as `username` and the first secondary group as primary group.

ametisf:
At least on my machine, the output of `groups username` is 'username : primarygroup group1 group2' etc. Why would that be broken?

Duncaen:
I edited my comment, this only works with coreutils/gnu `groups(1)`, every other implementation does not prefix the output with a username.

```
$ busybox groups duncan
duncan wheel audio video kvm input xbuilder socklog
```

ametisf:
I changed the snippet to include both implementations. Do you think it's ok like this?

Duncaen:
I would prefer a portable method, but I can't think of a way other than splitting the `groups(1)` output at `:`.

bobertlo:
I'm inclined against including this. It doesn't even seem like best practice to me.

If a cleaner solution is found and people feel strongly I could maybe see this, but it might be better left as an exercise to the reader.

Thank you for contributing though! It's always welcome :)

bobertlo:
Just to clarify my opinion. I think instead of more content to copy and paste into your system, this page could go in the other direction and explain why you are putting that code in and what it is doing.

Is groups coming from the directory name because of a feature of runit? chpst? the script? What is chpst? What is this script actually doing? Not that these are hard questions but it shouldn't read like a puzzle.

To me this is the difference between these docs and the wiki.

the-maldridge:
Seeing as this addition does not really fit with the goals of docs, I'm going to close this issue.  If you'd like to make a new PR with the style that @bobertlo suggests, feel free to do so.

PR #70, opened 2019-02-18
live-images/index.md: change voice and layout of information

bobertlo:


bobertlo:
@nilium @the-maldridge I did a lot of editing on this page for voice/clarity. Do these changes look productive? I would like to keep going through the docs like this (especially in this section) if that is okay. I pretty much copied and pasted these pages from the wiki. I added some information, but the English is not great, especially for ESL readers.

the-maldridge:
Seems okay to me.

src/installation/live-images/index.md
```diff
@@ -1,28 +1,42 @@
...
-> comfortable with a more advanced setup may prefer to install Void from a base
-> image.
+The base images provide only a minimal set of packages to install a usable Void
+system. These base packages are only those needed to configure a new machine,
+update the system, and install additional packages from repositories.
 
-### Comparison of flavor images
+### Flavor images
+
+Void also provides several "flavor" images which each include a full desktop
```
nilium: We've already called out that Void provides flavor images, so we can just skip directly to describing them.
```suggestion
Each of the Void "flavor" images includes a full desktop
```

src/installation/live-images/index.md
```diff
@@ -1,28 +1,42 @@
...
+environment, web browser, and basic applications configured for that
+environment. The only difference from the base images is the additional
+packages/services installed.
```
nilium: Try to favor "and" and "or" over slashes.
```suggestion
packages and services installed.
```

src/installation/live-images/index.md
```diff
@@ -1,28 +1,42 @@
...
+
+The install process for each of the flavor images is the same as the base
+images, except that you **must** select the `Local` source when installing,
+because the `Network` install only installs the base system, pulled fresh from
+the official mirrors.
```
nilium: Maybe something like this, since otherwise this paragraph is one long sentence.

```
The install process for each flavor image is the same as the base images, except
you **must** select the `Local` source when installing. If you select `Network`
instead, this only installs the base system, pulled fresh from the official
mirrors, without any additional packages.
```

PR #72, opened 2019-02-27
config/kernel.md: update query command and available kernels

pltrz:
Not sure what "dbg" was previously matching in querying "linux[34]" but it doesn't seem to match anything now; removed.

On the other hand, "hfsprogs" is matched with the query now. Ignored this match.

I also updated the sample of available kernel versions for this search.

### DIFF
--- a/src/config/kernel.md
+++ b/src/config/kernel.md
@@ -5,17 +5,14 @@
 Void Linux provides many kernel series in the default repository,
 
 ```
-$ xbps-query --regex -Rs 'linux[34]' | grep -Ev '(dbg|headers)'
-[-] linux3.14-3.14.79_1           The Linux kernel and modules (3.14 series)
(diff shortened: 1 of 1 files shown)


Duncaen:
It lists debug packages if you have `void-repo-debug` installed.

Duncaen:
What about `xbps-query --regex -Rs '^linux[0-9.]+-[0-9._]+'`?

pltrz:
> It lists debug packages if you have `void-repo-debug` installed.

My mistake then. Updated.

> What about `xbps-query --regex -Rs '^linux[0-9.]+-[0-9._]+'`?

This is much more correct and robust.

pltrz:
@Duncaen Am I correct in understanding that this regex also removes the need to exclude `dbg`, `headers`, etc matches with `grep`?

Duncaen:
Yes

bobertlo:
Nice Solution.

PR #74, opened 2019-03-01
multi-update

zlice:
-clarify kernel.md file "contents"
-section "enable" seems more fitting networkmanager.mg
-service enable but not start
-service headings for ease of finding information (like void wiki)
-xorg "video" clarification
-xorg input section
-style-guide missing hyphen leads to 404


zlice:
table in xorg

PR #75, opened 2019-03-01
multi-updates

zlice:
-clarify kernel.md file "contents"
-section "enable" seems more fitting networkmanager.mg
-service enable but not start
-service headings for ease of finding information (like void wiki)
-xorg "video" clarification
-xorg input section (added --- table line)
-style-guide missing hyphen leads to 404

### DIFF
--- a/src/config/kernel.md
+++ b/src/config/kernel.md
@@ -53,8 +53,8 @@ There are two different methods to blacklist kernel modules, for the initramfs
 and for the booted system. Some modules are loaded by the initramfs very early
 in the boot process, those have to be blacklisted in the initramfs.
 
-You can blacklist modules with a `.conf` file like `/etc/modprobe.d/radeon.conf`
-with the contents.
+You can blacklist modules with a `.conf` file like
+`/etc/modprobe.d/radeon.conf` with the contents.
 
 ```
 blacklist radeon
--- a/src/config/kernel.md
+++ b/src/config/kernel.md
@@ -40,7 +40,7 @@ is currently booted.
 ### Loading kernel modules at boot
 
 To load kernel modules at boot time, a `.conf` file like
-`/etc/modules-load.d/virtio.conf` can be created.
+`/etc/modules-load.d/virtio.conf` can be created with the contents.
 
 ```
 # load virtio-net
@@ -53,8 +53,8 @@ There are two different methods to blacklist kernel modules, for the initramfs
 and for the booted system. Some modules are loaded by the initramfs very early
 in the boot process, those have to be blacklisted in the initramfs.
 
-You can blacklist modules with a `.conf` file like
-`/etc/modprobe.d/radeon.conf`.
+You can blacklist modules with a `.conf` file like `/etc/modprobe.d/radeon.conf`
+with the contents.
 
 ```
 blacklist radeon
@@ -63,7 +63,7 @@ blacklist radeon
 #### dracut
 
 To blacklist modules from being included in a dracut initramfs you need to
-create a `.conf` file like `/etc/dracut.conf.d/radeon.conf`.
+create a `.conf` file like `/etc/dracut.conf.d/radeon.conf` with the contents.
 
 ```
 omit_drivers+=" radeon "
--- a/src/config/services/index.md
+++ b/src/config/services/index.md
@@ -3,6 +3,8 @@
 Void uses the [runit(8)](https://man.voidlinux.org/runit.8) supervision suite to
 run system services and daemons.
 
+## Enable
+
 Services are enabled by simply linking them into the `/var/service` service
 directory.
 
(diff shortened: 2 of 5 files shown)


src/config/kernel.md
```diff
@@ -40,7 +40,7 @@ is currently booted.
 ### Loading kernel modules at boot
 
 To load kernel modules at boot time, a `.conf` file like
-`/etc/modules-load.d/virtio.conf` can be created.
+`/etc/modules-load.d/virtio.conf` can be created with the contents.
```
pltrz: I don't think this improves clarity

src/config/kernel.md
```diff
@@ -54,7 +54,7 @@ and for the booted system. Some modules are loaded by the initramfs very early
 in the boot process, those have to be blacklisted in the initramfs.
 
 You can blacklist modules with a `.conf` file like
-`/etc/modprobe.d/radeon.conf`.
+`/etc/modprobe.d/radeon.conf` with the contents.
```
pltrz: Same as [1]

src/config/kernel.md
```diff
@@ -63,7 +63,7 @@ blacklist radeon
 #### dracut
 
 To blacklist modules from being included in a dracut initramfs you need to
-create a `.conf` file like `/etc/dracut.conf.d/radeon.conf`.
+create a `.conf` file like `/etc/dracut.conf.d/radeon.conf` with the contents.
```
pltrz: Same as [1]

src/config/services/index.md
```diff
@@ -3,19 +3,33 @@
...
 
+## Disable
+
 To disable them again you just remove the link.
 
 ```
 # rm /var/service/<service name>
 ```
 
+## Disable Autorun
```
pltrz: Maybe this should be reworded a bit, but I think it's a useful addition

src/config/services/index.md
```diff
@@ -3,19 +3,33 @@
 Void uses the [runit(8)](https://man.voidlinux.org/runit.8) supervision suite to
 run system services and daemons.
 
+## Enable
```
pltrz: [2] Personally, I think this page does not need this type of thing in it's own section; it's in the context of the page already.

src/config/services/index.md
```diff
@@ -3,19 +3,33 @@
...
+
 Services are enabled by simply linking them into the `/var/service` service
 directory.
 
 ```
 # ln -s /etc/sv/<service name> /var/service/
 ```
 
+## Disable
```
pltrz: Same as [2]

src/config/services/index.md
```diff
@@ -28,6 +42,8 @@ available and can be used like `sv <command> <services...>`.
 See the [sv(8)](https://man.voidlinux.org/sv.8) manual page for further
 information.
 
+## Status
```
pltrz: Same as [2]

src/config/xorg/index.md
```diff
@@ -51,3 +51,35 @@ Void also provides proprietary video drivers, which are available in the
 | NVIDIA     | Proprietary | 500+                            | `nvidia`       |
 | NVIDIA     | Proprietary | 300/400 Series                  | `nvidia390`    |
 | NVIDIA     | Proprietary | GeForce8/9 + 100/200/300 Series | `nvidia340`    |
+
+## Input Drivers
```
pltrz: I think this is a great addition.

src/config/kernel.md
zlice: saw this had it. https://github.com/void-linux/void-docs/blob/master/src/config/media/alsa.md
since code and files are in the same block it seemed off

zlice:
know what the format.sh is getting exit 2 for? not familiar with travis-ci but the script checks git status and isn't happy inside that vm. is it because i did master-master ?

Vaelatern:
It is giving you format fixes.
Follow them.

zlice:
it literally just wants the code to be a specific way? i was paying attention to the git HEAD detached part. i'll try to get to that

pltrz:
Are you using [`vmd`](https://github.com/bobertlo/vmd)? You'll need it to format correctly. It's in the repository.

pltrz:
See [this](https://docs.voidlinux.org/contributing/void-docs/style-guide.html#formatting).

src/config/services/index.md
bobertlo: I'm actually not too opposed to this, but more details per subsection would be nice if we did this. Like mentioning that the service will automatically be started.

src/config/services/index.md
bobertlo: I think this is a good addition, but maybe we can leave the disabling heading and remove this heading. And a little rewording would be useful. "To disable a service from starting at boot, while still being managed by runit, create a file named `down` in it's service directory." Or something.

src/config/services/index.md
bobertlo: Disabling Services

zlice:
@bobertlo will keep in mind in the future.

@pltrz after looking at that and installing i'm more confused. the format.sh / vmdfmt spits out what i already have.

zlice:
i will never understand how git rebasing works - just making branches like bobertlo suggested

PR #76, opened 2019-03-01
[WIP] config/network/iwd: create page

pltrz:
I've started working on a page for IWD.

There is still expansion to be done around the general configuration and network configuration sections. Some things might have to be shuffled around as well.

I want to keep only the essentials with a few simple examples and refer users to the [upstream documentation](https://iwd.wiki.kernel.org/networkconfigurationsettings), which has good examples for more complex network configurations.

### DIFF
--- a/src/config/network/iwd.md
+++ b/src/config/network/iwd.md
@@ -1,0 +1,98 @@
+# IWD
+
+[IWD](https://iwd.wiki.kernel.org/) (iNet Wireless Daemon) is a wireless daemon
+for Linux that aims to replace [WPA supplicant](./wpa_supplicant.md).
+
+## Installation
+
+Install the `iwd` package.
+
+```
+# xbps-install -S iwd
+```
+
+Enable the `dbus` and `iwd` services.
+
+```
+# ln -s /etc/sv/dbus /var/service
+# ln -s /etc/sv/iwd /var/service
+```
+
+> Note: To use EAP-TLS, EAP-TTLS, and EAP-PEAP based configurations, version
+> ≥4.20 of the kernel is required. Previous kernel versions do not include the
+> necessary cryptographic authentication modules.
+
+## Usage
+
+IWD provides a command line client, `iwctl`. It can be used to add, remove, and
+configure network connections; running it produces an interactive prompt. To
+list available options, enter `help`.
+
+```
+$ iwctl
+
+[iwd]# help
+                               Available commands
+--------------------------------------------------------------------------------
+  Commands                                          Description
+--------------------------------------------------------------------------------
+
+Adapters:
+  adapter list                                      List adapters
+  adapter <phy> show                                Show adapter info
+  adapter <phy> set-property <name> <value>         Set property
+...
+```
+
+> Note: By default, only the root user and those in the `wheel` group have
+> permission to operate `iwctl`.
+
+Some useful commands:
+
+```
+[iwd]# device list                                       # List all Wi-Fi devices
+[iwd]# device <interface> show                           # Display details of a Wi-Fi device 
+[iwd]# station <interface> scan                          # Scan for networks
+[iwd]# station <interface> get-networks                  # List networks
+[iwd]# station <interface> connect <network-name>        # Connect to a network
+[iwd]# station <interface> connect-hidden <network-name> # Connect to a hidden network
+[iwd]# station <interface> disconnect                    # Disconnect from a network
+[iwd]# known-networks list                               # List known networks
+[iwd]# known-networks forget <network-name>              # Forget a known network
+```
+
+> Note: `iwctl` also supports passing commands as arguments.
+> 
+> ```
+> $ iwctl station <interface> get-networks
+> ```
+
+## Configuration
+
+Consult the [upstream
+documentation](https://iwd.wiki.kernel.org/networkconfigurationsettings) for
+options and examples available to configure the daemon and networks as described
+below.
+
+### Daemon configuration
+
+The main configuration file is located in `/etc/iwd/main.conf`. If it does not
+exist, you may create it.
+
+### Network configuration
+
+IWD stores information on known networks, and reads information on
+pre-provisioned networks from network configuration files located in
+`/var/lib/iwd`; IWD monitors the directory for changes. Network configuration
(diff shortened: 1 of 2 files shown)


src/config/network/iwd.md
```diff
@@ -0,0 +1,74 @@
...
+[iwd]# station <interface> scan                          # Scan for networks
+[iwd]# station <interface> get-networks                  # List networks
+[iwd]# station <interface> connect <network-name>        # Connect to a network
+[iwd]# station <interface> connect-hidden <network-name> # Connect to a hidden network
+[iwd]# station <interface> disconnect                    # Disconnect from a network
+[iwd]# known-networks list                               # List known networks
+[iwd]# known-networks forget <network-name>              # Forget a known network
+```
+
+> Note: `iwctl` also supports passing commands from the interactive session as command line arguments.
```
bobertlo: "supports passing commands as arguments" maybe? This is confusing because iwctl seems to function as an interactive session otherwise.

pltrz: Yeah, I really wasn't sure how to word this. Thanks

bobertlo:
I don't want to get into too many third party packages in the docs but a maintained alternative to wpa_supplicant and NetworkManager would be nice to present to users :)

And I hadn't heard of this project before it looks interesting. Wish it didn't need dbus.

pltrz:
> I don't want to get into too many third party packages in the docs

Glad to agree on this. I do see IWD as essential as it looks like it will be the upcoming standard tool.

> maintained alternative to wpa_supplicant and NetworkManager

As I understand it, IWD is more of alternative to wpa_supplicant than to NetworkManager. For example NetworkManager can use IWD or wpa_supplicant as a backend. wpa_supplicant's interactive prompt (wpa_cli) is pretty clunky and has unnatural syntax, so people use NM as a frontend to it. IWD is nice since it's already got a nice interactive prompt so you don't need a second tool to make it useable; and as a plus one of it's design goals is to minimize external libraries as much as possible in favour of existing kernel features.

> Wish it didn't need dbus.

Can I ask the reason? I don't really have an opinion on this btw. Just curious. It does allow IWD to have a decent way of managing permissions and allowing exceptions such as with the [default config](https://git.kernel.org/pub/scm/network/wireless/iwd.git/tree/src/iwd-dbus.conf) and allows for [other possibilities](https://wiki.archlinux.org/index.php/Iwd#Deny_console_(local)_user_from_modifying_the_settings) (I have not experimented with this, but I have also not experimented with dbus in general). As I understand it, it's how other tools are able to get restricted elevated permissions.

bobertlo:
I'm glad it's more of a wpa_supplicant replacement than NetworkManager. As far as dbus: I do have it installed, and sometimes I think spotify starts it, but nothing on my core system depends on it and I guess I just enjoy that feeling. I haven't been actively avoiding it, but have yet to depend on it since using Void. I haven't looked into IWD closely yet, and if dbus lets it work better I'm not personally opposed to this. Also, this is just my personal views on my systems, not related to the docs at all.

My personal (laptop) network management is done with wpa_supplicant and I usually just use wifish to add networks and wpa_cli to manage saved networks. It works but it would be nice if there was a better system.

pltrz:
This page pretty much includes the essentials now. Like I mentioned in the first comment, for more exotic configurations we should point users upstream.

src/config/network/iwd.md
```diff
@@ -0,0 +1,119 @@
...
+```
+[Scan]
+disable_periodic_scan=true
+```
+
+## Network configuration
+
+IWD stores information on known networks, and reads information on pre-provisioned networks from text configuration files located in `/var/lib/iwd`. You can create, modify or remove those files. IWD monitors the directory for changes.
+
+> Note: Files inside the `/var/lib/iwd` directory have permssions set to be read and written only by the root user.
```
pltrz: I do not know if this is necessaryto include. Users can figure this out with an 

```
# ls -l /var/lib/iwd/
-rw------- 1 root root 109 Mar  3 12:58 <ssid>.psk
```

Let me know what you guys think

src/config/network/iwd.md
```diff
@@ -0,0 +1,119 @@
...
+
+A simple example of a WPA2/PSK secured network configuration at `/var/lib/iwd/<ssid>.psk` contains only the pre-shared key and the plaintext password.
+
+```
+[Security]
+PreSharedKey=XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
+Passphrase=<password>
+```
+
+> Note: If the plaintext password is included and the pre-shared key is ommited, the pre-shared key will automatically be calculated from the SSID and password, and written to this file by IWD upon attempting to connect. If the pre-shared key is included and the plaintext password ommited, the network will connect and the plaintext password will not be written to the network configuration file.
```
pltrz: I was not too sure about the clarity of my wording on this note. Open to suggestions on improving this.

pltrz:
I think in the end I will remove some of the details in the config. These should be left as an exercise to the user. It does not quite fit the style of the rest of the docs

pltrz:
Waiting on {bobertlo/vmd/issues/7, void-linux/void-docs/issues/81} to be resolved

bobertlo:
I fixed the formatter, but I'm going to sleep on it before making a release. Sorry about the space left in there, that's really nontrivial to change.

If you want to use master you can `go get github.com/bobertlo/vmd/cmd/vmdfmt`. Travis uses master so it's already updated. I'm stuck vendoring blackfriday now because this broke their HTML output very slightly, but it fixes that bug in the AST output lol

pltrz:
You were quick to fix that! Glad it could be sorted and wasn’t as difficult as it seemed. Great job.

I’ll fetch the latest master and push a new format.

I’m curious, I have not looked into the nature of that bug yet but is the fix you came up with something that could be pushed upstream?

bobertlo:
I sent in a PR, but then realized it broke a test and the very strange handling of this specific situation is for a feature on their end. So, maybe some other time I'll figure out a way to get both working, but it probably won't be pretty.

pltrz:
No sweat!

PR #77, opened 2019-03-03
file vs command clarificaiton

zlice:


pltrz:
I don't think this matches the style of voice we should go for in the docs. "You can" etc. I think we should give clear imperative instructions, so I think these should be reworked to begin with.

If we insist on how it is, though, I would suggest adding "... with the *following* contents:"

My suggestion would be something more along these lines.

> To exclude modules, blacklist them by adding them to configuration files in `/etc/modprobe.d/<module>.conf` with the `blacklist` keyword.
> 
> ```
> blacklist <module>
> ```

By no means does it have to be exactly like that but what I mean is that a different style of voice would fit better.

zlice:
made commas, you have colon example ?

zlice:
"you" wording was there since the init commit it looks like

bobertlo:
I'm sorry I meant colons, I need to stop typing things on my phone!

@zlice if you would be willing to, I think the suggestion on wording change by @pltrz would be a much better solution. Otherwise *colons* (sorry again) and I agree `with the following contents:` would be better.

I would love if you would take a shot at changing the wording, but I don't think we should make every contribution responsible for the entire page touched, and I do think these changes are at least an improvement. So I don't see harm in merging this and letting someone else worry about "voice" changes.

src/config/kernel.md
```diff
@@ -26,9 +26,9 @@ $ vkpurge list
 3.8.2_1
 ```
 
-You can now remove a specific kernel version like `3.8.2_1` or `all` which
-removes all kernels expect the latest kernel of each series and the kernel that
-is currently booted.
+To remove a specific kernel version like `3.8.2_1` or `all` (which removes all
```
bobertlo: This is not exactly a sentence. Maybe something like:

To remove a specific version:
```
vkpurge rm 3.8.2_1
```

To remove all kernels except the latest kernel of each series and the kernel that is curretly booted:
```
# vkpurge rm all
```

bobertlo: And some text above `vkpurge list` would be helpful.

zlice:
i feel like you're messing with me at this point. you could have re-wrote everything but keep tacking on "one more thing"

bobertlo:
I promise I am not messing with you! I only meant the lines you were already touching as far as rewording. If you revert that last section I commented on, or make better changes, I swear that is the last thing. When you go into other sections editing them, then it is up to you to get them acceptable to merge.

I am not trying to nitpick, and yes as far as rewriting this whole page I could do that myself, but I don't have time to merge changes that need immediate re-work. I understand your feelings though, I remember running into these same issues with commits like this :)

zlice:
understood. i'll try to do it later today/night

PR #79, opened 2019-03-03
headers and autostart disable

zlice:


src/config/services/index.md
```diff
@@ -3,19 +3,34 @@
 Void uses the [runit(8)](https://man.voidlinux.org/runit.8) supervision suite to
 run system services and daemons.
 
+## Enabling Service
```
bobertlo: Services

src/config/services/index.md
```diff
@@ -3,19 +3,34 @@
...
+
 Services are enabled by simply linking them into the `/var/service` service
 directory.
 
 ```
 # ln -s /etc/sv/<service name> /var/service/
 ```
 
+## Disabling Service
```
bobertlo: Services

src/config/services/index.md
```diff
@@ -3,19 +3,34 @@
...
+
 To disable them again you just remove the link.
 
 ```
 # rm /var/service/<service name>
 ```
 
+## Disable Autorun
```
bobertlo: "Disabling Autorun", or remove heading and leave this as an addition to "Disabling Services", which I would prefer.

PR #80, opened 2019-03-03
video drivers calfiy. input section

zlice:


src/config/xorg/index.md
```diff
@@ -51,3 +51,35 @@ Void also provides proprietary video drivers, which are available in the
...
+| `xf86-input-libinput` |
+| `xf86-input-mtrack`   |
+| `xf86-input-synaptics`|
+| `xf86-input-vmmouse`  |
+| `xf86-input-wacom`    |
+
+## Xorg config
+
+An Xorg config for a keyboard may look something like a file
+`/etc/X11/xorg.conf.d/30-keyboard.conf` with the contents.
```
Vaelatern: Please emphasize that this is almost never necessary.

bobertlo:
Thank you for splitting these up, but do try to make changes inside of existing PRs. Instead of doing a `git commit` you can just use `git commit --amend` to change your last commit and then `git push -f origin <branchname>` to force update. If you forget to do this and create a new commit you can run `git rebase -i HEAD~2` to merge the last two commits together.

src/config/xorg/index.md
```diff
@@ -51,3 +51,35 @@ Void also provides proprietary video drivers, which are available in the
...
+
+```
+Section "InputClass"
+  Identifier "keyboard-all"
+  Driver "evdev"
+  MatchIsKeyboard "on"
+EndSection
+```
+
+> Note: For more in depth configurations, please seek other documentation.
```
bobertlo: I think this is unnecessary, unless a link to documentation is given. I would lean towards just removing this though.

Vaelatern:
You can then run `git push -f` to force the PR to update

src/config/xorg/index.md
zlice: added that the config is to use a specific driver

src/config/xorg/index.md
zlice: removed

zlice:
i'd love to know why (even with "conflicts") the rebase worked today and wouldn't the other day .-.

src/config/xorg/index.md
bobertlo: Everything else looks great, but I would appreciate a stronger notice that this is almost never needed.

"Although it is almost never necessary, an Xorg config for ..."

PR #82, opened 2019-03-07
New favicon: Remove unreadable Void text

Anachron:
This favicon is way better to read in browsers. 

Compare https://docs.voidlinux.org/ to https://blog.cron.world/ .


Vaelatern:
It's the same favicon we use on all our sites.

Anachron:
Yep. There are a lot of different favicons around:

bobertlo:
Yes, I came here to say this is the wrong place to ask about a change like that. There is a different repository for the main site that a lot more people pay attention to.

bobertlo:
Please make a PR in https://github.com/void-linux/void-linux.github.io

This repo will follow whatever decision is made there so I'm closing this issue.

pltrz:
This was merged in void-linux/void-linux.github.io/pull/41 so we might consider merging this too now

PR #83, opened 2019-03-07
Update dhcpcd.md

fabulvoid:
A slash has been omitted at the end of the "ln -s" command.

### DIFF
--- a/src/config/network/dhcpcd.md
+++ b/src/config/network/dhcpcd.md
@@ -4,7 +4,7 @@ To run [dhcpcd(8)](https://man.voidlinux.org/dhcpcd.8) on all interfaces you can
 enable the `dhcpcd` service.
 
 ```
-# ln -s /etc/sv/dhcpcd /var/service
+# ln -s /etc/sv/dhcpcd /var/service/
 ```
 
 If you want to run dhcpcd just on a specific interface you can use the


bobertlo:
Does anyone have any opinion on this? both work, I count 10 without trailing '/' and 7 with. I almost favor without the slash, because of tab-completion, but I really don't have an opinion, but I guess consistency wouldn't hurt.

 @pltrz ? @nilium ?

@fabulvoid if you would like to modify one of these PR's to include every case of a missing '/' in the meantime that would help your case.

nilium:
Runit’s FAQ uses trailing slashes for that sort of example: http://smarden.org/runit/faq.html#tell

It’s what I personally do when linking services since it feels clearer to me that the service is going under the dir and not replacing the /var/service symlink. I don’t think it matters too much though, other than being unambiguous about the destination.

bobertlo:
Good enough for me.

pltrz:
I agree with @nilium. Trailing slash makes it explicit, rather than implicit. Explicit works better here imo, since implicit assumes the reader knows how runit is structured.

bobertlo:
@fabulvoid thanks for contributing. in the future you can group small things like this into a single PR. I'm going to change all the other instances to this format.

fabulvoid:
Thanks, @bobertlo! I‘ll have to find out how to group passages from several files into a single PR. Will it be ok to just copy the file names and passeges into the comment filed of the PR?
Sorry to have to ask, but these were my very first PRs in my life.

PR #86, opened 2019-03-10
Create iptables.md

fabulvoid:
Describes how to activate the iptables firewall.

### DIFF
--- a/src/config/network/iptables.md
+++ b/src/config/network/iptables.md
@@ -1,0 +1,64 @@
+# Firewall with iptables
+
+The `iptables` package is installed by default on the base system, but not activated.
+
+Check for the `iptables` and `ip6tables` directories:
+
+```
+$ ls /etc/sv
+...
+ip6tables
+iptables
+...
+```
+
+## Firewall rules
+
+Two rulesets are already installed in the `/etc/iptables` directory:
+
+```
+$ ls /etc/iptables
+...
+empty.rules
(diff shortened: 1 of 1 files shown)


Duncaen:
I think the runit services for iptables are bad and shouldn't be promoted.
All runit services start in parallel, your web service might start before the iptables rules are loaded.

Vaelatern:
The alternative is a line in `rc.local` which is the correct course.

fabulvoid:
Thanks for your feedback. I have edited 'iptables.md' according to @Vaelatern's suggestion. Hope my contribution is more usable now.

Piraty:
Void also ships `runit-iptables` for a few month now, which restore iptables during stage1 (core-services). If this can be considered elegant is up to debate, but at least they are activated before network is up...

bobertlo:
@Piraty that looks interesting.

@fabulvoid please rebase this branch into a single commit.

bobertlo:
@Duncaen @Vaelatern does the `runit-iptables` package seem like a better solution to this? It looks clean to me and I think it would be simpler to mention that package and, maybe, also mention `rc.local` as an alternative? It's not a big difference either way, I'd just like opinions.

bobertlo:
Closed by https://github.com/void-linux/void-docs/commit/aa9a6ad2ce51a656866e2868ad2c12cf16d2728c

PR #87, opened 2019-03-25
Iptables

bobertlo:
The other PR seems to be idle. I squashed the two commits and then formatted and changed the `rc.local` script addition slightly. I think it would be worth noting at least that this will block ALL incoming traffic (including ssh.) I'd appreciate any input on that. Otherwise I'd just like review on this.


Vaelatern:
This looks OK.

PR #92, opened 2019-04-19
config/media/pulseaudio: mention alsa-plugins-pulseaudio

maxice8:


src/config/media/pulseaudio.md
```diff
@@ -7,6 +7,12 @@ PulseAudio depends on a `dbus` system daemon, make sure its enabled.
 # ln -s /etc/sv/dbus /var/service/
 ```
 
+PulseAudio also depends on `alsa-plugins-pulseaudio` to work with ALSA.
```
jnbr: Maybe be more specific on what this is needed for. It's not a hard dependency.

maxice8:
On Fri, Apr 19, 2019 at 02:55:18PM -0700, Johannes wrote:
> jnbr commented on this pull request.
> 
> 
> 
> > @@ -7,6 +7,12 @@ PulseAudio depends on a `dbus` system daemon, make sure its enabled.
>  # ln -s /etc/sv/dbus /var/service/
>  ```
>  
> +PulseAudio also depends on `alsa-plugins-pulseaudio` to work with ALSA.
> 
> Maybe be more specific on what this is needed for. It's not a hard dependency.
> 
> -- 
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub:
> https://github.com/void-linux/void-docs/pull/92#pullrequestreview-228837927

Isn't 'to work with ALSA' enough ? as far as i have experienced it i needed it 100% of the time i used any ALSA application.

jnbr:
I think it's a bit vague, would prefer `to wirk with ALSA applications` as it doesn't affect ALSA as a sink afaiu.

maxice8:
Updated

bobertlo:
If I understand this correctly this looks ok. I don't use Pulseaudio at all though.

pltrz:
This looks accurate to me

Duncaen:
> For applications which use ALSA directly and don't support PulseAudio, the `alsa-plugins-pulseaudio` package can be installed to make them use PulseAudio through ALSA.

maxice8:
On Sat, Apr 20, 2019 at 04:47:25AM -0700, Duncan Overbruck wrote:
> > For applications which use ALSA directly and don't support PulseAudio, the `alsa-plugins-pulseaudio` package can be installed to make them use PulseAudio through ALSA.
> 
> -- 
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub:
> https://github.com/void-linux/void-docs/pull/92#issuecomment-485101340

fixed

Issue #63, opened 2019-02-08
Accessibility improvements


Duncaen:
- [x] Skip to content https://github.com/void-linux/void-docs/pull/61
- [ ] unbind mdbooks arrow key bindings? I think this could interference with users who use the keyboard/arrow keys to navigate web pages.
- [ ] sitemap? not supported by mdbook.

pltrz:
When browsing using qutebrowser, which uses insert-mode for typing, typing into the search bar causes it to exit insert-mode after each character typed

Duncaen:
The search functionality comes from mdbook without any modifications, I don't really want to divert from it.
https://github.com/rust-lang-nursery/mdBook/blob/master/src/theme/searcher/searcher.js#L270-L277
https://github.com/rust-lang-nursery/mdBook/blob/master/src/theme/searcher/searcher.js#L333-L334
They do unfocus the input and refocus it after every key event for some reason.

PR #94, opened 2019-04-29
Sitemap generator

jakubvokoun:
refs: #63

A simple shell script could make the job.

### DIFF
--- a/sitemap.sh
+++ b/sitemap.sh
@@ -1,0 +1,56 @@
+#!/bin/sh
+
+set -e
+
+# functions
+fail() {
+    echo "$1" 1>&2
+    exit 1
+}
+
+
+get_last_mod() {
+    if [ ! -f "$1" ]; then
+        date --iso-8601
+    else
+        git log -1 --format="%ci" "$1" | awk '{print $1}'
+    fi
+}
+
+# vars
+book_path="book"
+src_path="src"
+output="$book_path/sitemap.xml"
+changefreq="weekly" # values: always hourly daily weekly monthly yearly never
+url_prefix="https://docs.voidlinux.org"
+
+# check if build directory exists
+test -d $book_path || fail "Directory $book_path does not exist."
+
+# cleanup
+test -f $output && rm $output
+
+# sitemap generator
+exec 1> $output
+
+echo "<?xml version=\"1.0\" encoding=\"UTF-8\"?>"
+echo "<urlset xmlns=\"http://www.sitemaps.org/schemas/sitemap/0.9\">"
+
+tmp=$(mktemp)
+
+find "$book_path" -type f -name "*.html" > "$tmp"
+while IFS= read -r f
+do
+    p=$(echo "$f" | sed "s/$book_path//g")
+    md=$(echo "$p" | sed "s/.html/.md/g")
+    lastmod=$(get_last_mod "$src_path$md")
+    printf "\t<url>\n"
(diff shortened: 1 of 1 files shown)


ericonr:
@jakubvokoun do you think you could put this script inside the CI folder? I think it fits better there.

jakubvokoun:
@ericonr No problem. Should I also add i to `.travis.yml`?

```
script:
  - ci/format.sh
  - ci/sitemap.sh
```

the-maldridge:
Hmm, thought I'd answered this some time ago.  This won't work as written because it won't ever be run on production infrastructure.  There would need to be a fix somewhere in the scriptlets that build the site unless mdbook has suddenly gained a hooks subsystem on build.

ericonr:
@the-maldridge so we'd need to make some modification to the ansible roles for this to run the way it is now? I can look into mdbook a bit further.

ericonr:
@jakubvokoun Adding it to travis wouldn't hurt, and would help verify the script isn't malfunctioning for some reaon. I'm waiting on https://github.com/void-ansible-roles/mdbook/pull/1, after it's merged I can try to add this there as well.

It'd be necessary to add a bit more logic to your script a well. If you add `[mdbook.linkcheck]` to `book.toml` (which I want to do, as it allows checking links in CI), the book output goes into `book/html/` instead of `book/`. Ideally, I'd like to support both of these scenarios.

ericonr:
@jakubvokoun the script needs to support the output directory being in `book/` or `book/html/`. With support for that, we can look into adding it to the ansible role.

flexibeast:
@Duncaen: Do you feel a sitemap is still needed for accessibility purposes? The sidebar seems to essentially be a sitemap, and can be tabbed through (after first tabbing through the navbar).

flexibeast:
In a conversation on IRC, @Duncaen indicated he'd like a page with links to each section of the Handbook, accompanied by when that section was last updated.

i'd like to set aside presentation issues for now, and discuss the generation of this content.

We could manually create the content: each commit involving changes to Handbook content would also have to include not only any relevant changes to `SUMMARY.md`, but also make record the change date in (say) `sitemap.md`.

We could instead try to auto-generate the content. It's easy enough to write a script to go through `SUMMARY.md` and run something like:

```
 git log -n1 --format=format:"%ai" src/about/index.md
```

on each file, to produce dates and times. However, it seems to me that if we try to do this prior to the deployment process for docs.voidlinux.org, we would end up needing two distinct commits: one for the change to the content, and one that uses the log data from the first commit to update `sitemap.md` appropriately.

One way we might be able to work around would be to make `sitemap.md` a placeholder page; its content would be generated during the deployment process by getting `git log` data at that point.

Thoughts/comments on all this?

jakubvokoun:
> @jakubvokoun the script needs to support the output directory being in `book/` or `book/html/`. Do you think you could look into adding that, please?

Now you can set output directory using env variable, default vaule is `book/`.

classabbyamp:
> mdbook has suddenly gained a hooks subsystem on build

sorry for necroposting, but this should now be possible, see [(minimal) docs here](https://rust-lang.github.io/mdBook/format/configuration/preprocessors.html#provide-your-own-command) and [example here](https://github.com/void-linux/void-packages/blob/7519edcbbc5f7fe9fbc43499bd7a289b9898198f/manual/book.toml#L8). The "API" for custom preprocessors is json, so a sitemap preprocessor could use jq/python/etc to generate it

PR #97, opened 2019-05-08
guides: add full disk encryption walkthrough

abenson:


Vaelatern:
I personally would recommend going "Go see how to set up a disk, but keep these things in mind and run these commands maybe" instead of duplicating other sections of the guide.

abenson:
Okay, tonight I'll reduce any duplication based on what already exists in the section.

illiliti:
what about detached header?

bobertlo:
Wow, I completely forgot this was here. I didn't feel qualified to review this since I did not have FDE working. I will review this ASAP.

src/installation/guides/fde.md
```diff
@@ -0,0 +1,254 @@
...
+
+Device      Start       End   Sectors  Size Type
+/dev/sda1    2048    264191    262144  128M EFI System
+/dev/sda2  264192 100663262 100399071 47.9G Linux filesystem
+```
+
+Configure the encrypted volume.
+
+```
+# cryptsetup luksFormat --type luks1 /dev/sda1
```
ericonr: Why did you default to luks1? I think nowadays most kernels in Void support luks2, and it is a better option, if supported. This could perhpas be explained.

src/installation/guides/fde.md
```diff
@@ -0,0 +1,254 @@
...
+
+Once the volume is created, it needs to be opened. Replace voidvm with an
+appropriate name.
+
+```
+# cryptsetup luksOpen /dev/sda1 voidvm      
+Enter passphrase for /dev/sda1: 
+```
+
+Once the LUKS container is opened, create the LVM volume group using that
```
ericonr: It is possible to make an FDE setup without logical volumes. Do you think it would make sense to add that information here?

src/installation/guides/fde.md
```diff
@@ -0,0 +1,254 @@
...
+└─sda1               crypto_LUKS       135f3c06-26a0-437f-a05e-287b036440a4   
+  └─voidvm        LVM2_member       gjOBNB-SXiK-qcGn-uHxp-ZdBZ-14wd-4FIL7p 
+    ├─voidvm-root xfs         root  b4e4aa35-e819-42a4-bbfc-cdcd74f1df8a   /
+    ├─voidvm-swap swap              1f8e9ca3-65bc-45b8-b7d0-6300587dfcf8   [SWAP]
+    └─voidvm-home xfs         home  b6cedb19-d8f6-4629-8519-cd2e7279ec2c   /home
+sr0                                                                           
+```
+
+Edit the `GRUB_CMDLINE_LINUX_DEFAULT=` line in `/etc/default/grub` and add
+`rd.auto=1` and `cryptdevice=UUID=<UUID>:lvm`. Make sure the UUID matches from
```
ericonr: I tried using cryptdevice and failed, perhaps because I wasn't using rd.auto. My setup is using rd.luks.uuid now, and it works well enough :)

ericonr:
Do you think splitting this guide into two pages, one for the part of it that relates to a chroot install, and the other that is specific for FDE, would make sense? I'm thinking of writing an addendum guide for setting up Secure Boot with dracut, and not having to repeat some of the information would make the guides a lot better, I think. I could help with writing.

bobertlo:
@ericonr I think that is eventually the goal, but this is a huge question that needs to be documented, and it's been sitting here almost a year. I'm feeling like we can just merge this at this point and if someone writes a chroot guide we can point to that from here.

src/installation/guides/fde.md
abenson: grub doesn't support luks2 yet. 

Support has been merged, but it hasnt been released

src/installation/guides/fde.md
abenson: Is that using `rd.luks.uuid` without `cryptdevice`?

src/installation/guides/fde.md
abenson: Once it's released it isn't hard to convert to luks2

abenson:
Splitting it was the goal, I just never got around to it.

bobertlo:
One thing I think would be helpful is to not use the same name `voidvm` for both the LUKS volume and the LVM volume group. It doesn't change anything in practice but for clarity would be helpful?

bobertlo:
Well, if noone objects or is working on a chroot guide *presently* I am going to do a final review and merge this as soon as I have time. We can de duplicate when it is duplicated.

bobertlo:
Sure, I think this may need a small amount of love still, and could be split up. But we can't avoid duplication by *not merging either guide* so when someone writes a general chroot install guide then we can deal with that.

Thanks everyone involved in this.

PR #98, opened 2019-05-23
live-images: Fixed typo in MATE wm name

d9a:
I feel like this is the most petty thing ever to fix, but I also wanted to learn the process of a pull request (proper formatting of commits too) before doing something that actually matters.


pltrz:
Do not underestimate the power of a fix, no matter the size!!

Thanks for your contribution! You've made the project that much better. :muscle:

PR #101, opened 2019-06-18
add apulse info on alsa page

xythz:


src/config/media/alsa.md
```diff
@@ -39,3 +39,19 @@ default.
...
+```
+Once installed you can use the apulse wrapper script with this command.
+```
+$ apulse <program-name>
+```
+Environment variables such as `APULSE_CAPTURE_DEVICE` and `APULSE_PLAYBACK_DEVICE` can be used to set which capture and playback device you want apulse to use.
+```
+APULSE_CAPTURE_DEVICE=default APULSE_PLAYBACK_DEVICE=default apulse firefox
+```
+The above example would run firefox with apulse using the default alsa capture and playback devices.
```
ericonr, line 57: You need whitespace between each block code and the text. Also before and after the header.

ericonr:
No idea if this is still active, but if it is:

* reformat the stuff
* make a single commit (search for how to squash them)
* probably only mention the existence of apulse and link to its manpage or [upstream](https://github.com/i-rinat/apulse). The explanation for how to use it isn't Void specific.

flexibeast:
@ericonr: The [current ALSA page](https://docs.voidlinux.org/config/media/alsa.html) mentions `apulse`, but not the env vars. The env vars aren't mentioned on [the man page](https://man.voidlinux.org/apulse.1), but they are mentioned in [the project README](https://github.com/i-rinat/apulse/blob/master/README.md). i think we could close this by just adding a reference to the latter - what do you think?

ericonr:
@flexibeast I might just make a PR for the project's man pages. Then we can just link to the man page and be done with it. What do you think?

ericonr:
https://github.com/i-rinat/apulse/pull/112 now we wait :p

xythz:
Ah, sorry I forgot about this pull request @ericonr @flexibeast Thank you for resolving this, I'll close this now since you guys put up a commit.

ericonr:
@xythz if you want to reopen it to add the link to the man pages, no problem :)

Just need to clean up the commits a bit before we can merge.

flexibeast:
@ericonr: Thanks for submitting that PR to the `apulse` repo. :-)

Hmm, even if that PR gets merged pronto, and then the `apulse` package got immediately updated, istr that the relevant page on man.voidlinux.org wouldn't just automatically become available? @the-maldridge?

the-maldridge:
I wouldn't trust anything is going to work with man.voidlinux.org right now.  It needs to be reworked and migrated.

flexibeast:
@the-maldridge: \*nod\* Thanks.
@ericonr: Given the above, i'd like to suggest that for now we simply add a reference to the `apulse` README. Any objections?

PR #103, opened 2019-06-24
Description of removing default kernels

realiddqd:
Since its impossbile to remove default kernel series with simple xbps-remove -f command, proposed option should be added to free additional space, time and, as a bonus, stress to SSD drives.

### DIFF
--- a/src/config/kernel.md
+++ b/src/config/kernel.md
@@ -41,6 +41,15 @@ that is currently booted
 # vkpurge rm all
 ```
 
+## Completely remove kernel series marked as default
+
+To remove default kernel series execute
+
+```
+# echo "ignorepkg=linux4.19" | sudo tee -a /etc/xbps.d/10-ignore.conf; xbps-remove linux4.19
+```
+Make sure you have installed and tested desired kernel series.
+
 ## Kernel modules
 
 ### Loading kernel modules at boot
--- a/src/config/kernel.md
+++ b/src/config/kernel.md
@@ -46,7 +46,7 @@ that is currently booted
 To remove default kernel series execute
 
 ```
-# echo "ignorepkg=linux4.19" | sudo tee -a /etc/xbps.d/10-ignore.conf; xbps-remove linux4.19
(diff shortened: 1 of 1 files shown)


Chocimier:
command is already called by root, no need for sudo

Vaelatern:
Please learn about `vkpurge`.

realiddqd:
> Please learn about `vkpurge`. How is it not adequate?

Like i did not already. But if pkg marked as meta\default one cant remove whole pkg. Not some single kernel. Its a different thing.

Vaelatern:
I still don't understand why you want this. What's the use-case?

pltrz:
For example, if you have the `linux5.1` package and use that kernel, you might want to remove the default `linux4.19` kernel if you don't use it

Vaelatern:
You know how it's set to default, we will update the default later.

pltrz:
Although, I think we should avoid "one-liners" like these, instead outlining each step specifically and explaining how/why. I know it's a tiny thing but it goes a long way with clear documentation

pltrz:
@Vaelatern has it ever been considered to use the kernel packages with `xbps-alternatives` or something, for this exact reason? On Gentoo there is `eselect`, etc. People often use different kernels than the default.

realiddqd:
Well if someone can put good text with explanation since some of commands might be useful in other cases, then please do.

ericonr:
@flexibeast and I are looking into extending the documentation about xbps capabilities further, to make an operation like the one you described more obvious.

flexibeast:
i'm unclear what, if anything, still needs to be done here. `vkpurge` is mentioned in the Handbook in [the "Kernel" page](https://docs.voidlinux.org/config/kernel.html); is there something else that should be added to the Handbook somewhere?

cc: @Vaelatern

Vaelatern:
I feel this PR won't be accepted. The use of ignorepkg should be documented but not in this fashion. It exists in the man pages, and a guide to local package management could include using ignorepkg.

PR #105, opened 2019-08-09
Add 'TeX Live' section.

flexibeast:
Today i got around to installing TeX Live, and found the wiki page about it helpful. With the wiki going away, i think it would be useful to add this information to the handbook.

### DIFF
--- a/src/config/texlive.md
+++ b/src/config/texlive.md
@@ -1,0 +1,91 @@
+# TeX Live
+
+In Void, the `texlive-bin` package provides a basic TeX installation, including
+the `tlmgr` program. Use `tlmgr` to install TeX packages and package collections
+from CTAN mirrors. Install the `gnupg` package to allow `tlmgr` to verify TeX
+packages.
+
+The `texlive-bin` package contains the latest TeX Live version; however, earlier
+versions, such as `texlive2018-bin`, are also available.
+
+## Configuring TeX Live
+
+After installing TeX Live, update the value of `PATH`:
+
+```
+$ source /etc/profile
+```
+
+Check that `/opt/texlive/<year>/bin/x86_64-linux` (or
+`/opt/texlive/<year>/bin/i386-linux`) is in your `PATH`:
+
+```
+$ echo $PATH
+```
+
+If required, change the global default paper size:
+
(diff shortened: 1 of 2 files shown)


Duncaen:
For commands that need to be run as root, we use `#` as prefix instead of `$` without sudo.
https://docs.voidlinux.org/contributing/void-docs/style-guide.html#commands

I think instructions like `xbps-install textlive-bin` are not necessary, users should already know how to install packages and/or refer to the right documentation which describes how to install packages.
I would also add the `gnupg` requirement to the initial block where it says to install `texlive-bin` and maybe also mention that `texlive-bin` is the latest version and other versions like `texlive2018-bin` are available.

flexibeast:
Thanks for the quick feedback!

Sorry for not having changed the `$ sudo ...` to `# ...` - yes, i'd read about `$` vs `#` in the style guide, but wasn't sure if there was some particular reason the wiki had specified the use of `sudo` instead.

i've now updated the PR with the requested changes. However, unfortunately i forgot to do a `git --amend` before force-pushing, so the PR now has two commits - my apologies. i hope that doesn't cause too much inconvenience.

leahneukirchen:
Looks good.  You could also mention `tlmgr install scheme-full` to install all available packages.

flexibeast:
Thanks for the suggestion; i've updated the PR accordingly.

flexibeast:
Ping?

src/config/texlive.md
```diff
@@ -0,0 +1,94 @@
...
+$ tlmgr search --global <package-name>
+```
+
+and then install it:
+
+```
+# tlmgr install <package_name>
+```
+
+If the error is a missing file and not a package (for example a font), search
```
jnbr: Nitpick: "the error" appears the first time here.

flexibeast:
@jnbr: Thanks for the feedback; i've now fixed that section. Hope all is now in order.

jnbr:
Thank you. Looks good to me now.

bobertlo:
I don't know about TeX Live, but there seems to be a consensus and I don't see any glaring issues otherwise.

PR #107, opened 2019-08-13
Remove installing.md and move it into index.md

codingHahn:
The installing.md file only provided one paragraph, which can be
integrated into the index.md. This makes the structure of the pages more
logical in my opinion. 
Additionally, the repositories are now listed in a table.


src/xbps/repositories/official/index.md
```diff
@@ -3,11 +3,23 @@
...
-The next sections provide information on installing repositories, and summaries
-of the different repositories offered.
+| Package			| Description						| Link					|
+|-------------------------------|-------------------------------------------------------|---------------------------------------|
+| `void-repo-multilib`		| 32-bit compatibility packages 			| [multilib](./multilib.md)		|
+| `void-repo-mulitlib-nonfree`	| 32-bit compatibility packages with nonfree components | [multilib-nonfree](./multilib.md)	|
+| `void-repo-nonfree`		| packages with unfree licenses 			| [nonfree](./nonfree.md)			|
+| `void-repo-debug`		| debug symbols for packages 				| [debug](./debug.md)			|
+
+The next sections provide summaries of the different repositories offered.
```
bobertlo: This is no longer needed

src/xbps/repositories/official/index.md
```diff
@@ -3,11 +3,23 @@
...
-- [multilib/nonfree](./multilib.md): contains non-free multilib packages
-- [nonfree](./nonfree.md): contains software packages with non-free licenses
-
-The next sections provide information on installing repositories, and summaries
-of the different repositories offered.
+| Package			| Description						| Link					|
+|-------------------------------|-------------------------------------------------------|---------------------------------------|
+| `void-repo-multilib`		| 32-bit compatibility packages 			| [multilib](./multilib.md)		|
+| `void-repo-mulitlib-nonfree`	| 32-bit compatibility packages with nonfree components | [multilib-nonfree](./multilib.md)	|
+| `void-repo-nonfree`		| packages with unfree licenses 			| [nonfree](./nonfree.md)			|
```
nilium:
```suggestion
| `void-repo-nonfree`		| packages with non-free licenses 			| [nonfree](./nonfree.md)			|
```

Also, definitely run this through the formatter. Inline tab characters will inevitably break text layout if you're using the wrong tab width.

src/xbps/repositories/official/index.md
```diff
@@ -3,11 +3,23 @@
...
+
+The next sections provide summaries of the different repositories offered.
+
+## Installing
+
+Those repositories may be installed via the packages from the package column. 
+For example, to install the `nonfree` repository, install the package `void-repo-nonfree`:
+
+```
+# xbps-install -S void-repo-nonfree
```
nilium:
```suggestion
# xbps-install void-repo-nonfree
```

The -S would have to happen after installation, not before. If the reason for the -S is that the user should have synced current package data, then I think this goes without saying.

src/xbps/repositories/official/index.md
```diff
@@ -3,11 +3,23 @@
...
+| `void-repo-multilib`		| 32-bit compatibility packages 			| [multilib](./multilib.md)		|
+| `void-repo-mulitlib-nonfree`	| 32-bit compatibility packages with nonfree components | [multilib-nonfree](./multilib.md)	|
+| `void-repo-nonfree`		| packages with unfree licenses 			| [nonfree](./nonfree.md)			|
+| `void-repo-debug`		| debug symbols for packages 				| [debug](./debug.md)			|
+
+The next sections provide summaries of the different repositories offered.
+
+## Installing
+
+Those repositories may be installed via the packages from the package column.
```
nilium:
```suggestion
These repositories can be installed from the packages named in the table above. 
```

src/xbps/repositories/official/index.md
```diff
@@ -3,11 +3,23 @@
 Void provides other official repositories, which are maintained by the Void
 project, but not installed in the default configuration:
```
nilium:
```suggestion
project, but not installed in the default configuration. The following table describes XBPS packages
you can install to enable additional repositories and what those repositories contain.
```
Any time there's a table, there should probably be a line that introduces that table's contents, especially now that the contents of the table are referred to in later writing.

src/xbps/repositories/official/index.md
bobertlo: I didn't notice this. Definitely change this.

src/xbps/repositories/official/index.md
codingHahn: It seems to be common practice to have that -S flag there.

codingHahn:
Done. Should I remove the pages for the individual repos as well?

src/xbps/repositories/official/index.md
bobertlo: I think nilium is technically right on this one.

bobertlo:
I personally think those are fine left as separate pages for now.

Now that I look closer I think the 'package' part of this table could just have links to the other pages. I don't think the extra column is necessary.

Sorry if the review process seems nit-picky. Keeping the quality of each commit up really helps the quality of the whole project.

src/xbps/repositories/official/index.md
nilium: Hold off on changing anything here while I check on IRC.

bobertlo: This is a separate issue we don't need to deal with in this commit.

codingHahn:
No problem. I like the process that way.
My reasoning for having an extra column is the ability to double click to highlight the package names for easier copy and paste.

bobertlo:
Honestly, this table is just not working for me. If it could go back to something more like the original list, I would consider merging it.

Otherwise, I am looking at this section though and wondering if it could all just fit on the one page like you suggested. The debug page has a lot of content that doesn't really fit with the rest, but everything else could just be described under a single heading. Not sure how to handle the debug info. I can't think of anywhere else it would belong though.

ericonr:
@codingHahn would you like to work on @bobertlo 's suggestions? I generally agree with them, most of the pages in that section could fit under a single page. Regarding repositories themselves, I think you could mention them by the name of the package, but also require note that they only exist for glibc installations.

codingHahn:
Good suggestions. I will try to look at this next weekend. If you hear nothing from me, feel free to ping me.

ericonr:
@codingHahn ping?

codingHahn:
So I finally looked at it again (Sorry for the wait). My idea would be to have some kind of quick reference at the top, like the table here, and then explain the repos in subsequent headings.

I quite like the table as a formatted quick reference. In my eyes, a simple list would not look nearly as good, because we would need at least three pieces of information per list item:
- package name
- brief description
- link to subheading
It just feels cleaner in a table.

ericonr:
@codingHahn no problem!

Why couldn't the links be something like:

```
- [`void-repo-mulitlib-nonfree`](./multilib.md): <insert repo description>
```

src/xbps/repositories/official/index.md
```diff
@@ -1,13 +1,26 @@
...
+These repositories can be installed from the packages named in the table above.
+For example, to install the `nonfree` repository, install the package
+`void-repo-nonfree`:
+
+```
+# xbps-install -S void-repo-nonfree
+```
+
+> Note: These packages only install a repository configuration file in
+> `/usr/share/xbps.d`.
```
ericonr, line 26: We have updated the style guide a bit:

- Quotes shouldn't be used for notes. If you can add this in a way that fits the text flow better that would be awesome.
- We want to avoid explicit commands, especially when they are something like `xbps-install`.

ericonr:
What do you think of @bobertlo's suggestion regarding folding all the repo pages into this one, too? And we should probably also mention that you need to sync the repos after installing a repository package.

src/xbps/repositories/official/index.md
codingHahn: Ok. I'll try to rewrite the note. About the explicit command: should I just write something like "Install the package ..."?

codingHahn:
Yeah, I can do that

src/xbps/repositories/official/index.md
ericonr: Yes, just that c:

codingHahn:
> @codingHahn no problem! Sorry for my delay too.
> 
> Why couldn't the links be something like:
> 
> ```
> - [void-repo-mulitlib-nonfree](./multilib.md): <insert repo description>
> ```

To copy the repo name you need to press ALT and doubleclick instead of just doubleclicking. I understand that this is mostly a non-issue.

ericonr:
Hi! Flexibeast ended up doing some of the work of merging the files, but I am still interested in having the list in the format mentioned above. We could think of a way of making it less repetitive, by perhaps removing explicit mentions of each package in their paragraphs, and instead introducing the list like "Other repositories can be added by installing their respective packages. These are the packages for the other official repositories:"

src/xbps/repositories/official/index.md
```diff
@@ -1,13 +1,73 @@
...
+## Debug
+
+Void Linux packages come without debugging symbols, if you want to debug
+software or look at a coredump you will need the debugging symbols. These
+packages are contained in the debug repo. Install the `void-repo-debug` package
+to enable this repository.
+
+### Finding debug dependencies
+
+The `xtools` package contains the `xdbg` utility to retrieve a list of debug
```
ericonr, line 60: I would make `xbdg` a man page link.

ericonr:
Thanks for the contribution!

I believe most of the intended changes have already been done by other commits in the repo, so I'm closing this. If you'd like to make any new changes to this page, please open a new PR c:

PR #108, opened 2019-08-13
Kernel.md

codingHahn:
I updated the kernel.md page. I added a bit of context regarding vkpurge and kernel modules. Additionally, I added the missing mkinitcpio explanation. I hope all is right.

### DIFF
--- a/src/config/kernel.md
+++ b/src/config/kernel.md
@@ -36,14 +36,14 @@ $ vkpurge list
 3.8.2_1
 ```
 
-To remove a specific kernel version like `3.8.2_1`
+To remove a specific kernel version like `3.8.2_1`:
 
 ```
 # vkpurge rm 3.8.2_1
 ```
 
 To remove `all` kernels except the latest kernel of each series and the kernel
-that is currently booted
+that is currently booted:
 
 ```
 # vkpurge rm all
@@ -88,7 +88,7 @@ omit_drivers+=" radeon "
 ```
 
 Now initramfs needs to be regenerated to make the changes take effect on the
-next reboot.
+next reboot:
 
 ```
 # dracut --force
@@ -104,7 +104,7 @@ blacklist radeon
 ```
 
 Now initramfs needs to be regenerated to make the changes take effect on the
-next reboot.
+next reboot:
 
 ```
 # mkinitcpio -p linux
--- a/src/config/kernel.md
+++ b/src/config/kernel.md
@@ -21,13 +21,15 @@ modules.
 
 ## Removing old kernels
 
-When updating the kernel, old versions are left behind in case it is neccessary to roll back to an older version.
-Over time, many old kernel version accumulate and make updating dmks modules take a long time.
-Thus, it may be advisable to clean old kernels from time to time.
+When updating the kernel, old versions are left behind in case it is neccessary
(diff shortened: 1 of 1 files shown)


src/config/kernel.md
```diff
@@ -21,28 +21,40 @@ modules.
 
 ## Removing old kernels
 
-To list kernels that can be purged
+When updating the kernel, old versions are left behind in case it is neccessary to roll back to an older version.
+Over time, many old kernel version accumulate and make updating dmks modules take a long time.
```
bobertlo: This could be shortened to something just like 'As old kernels accumulate, you may wish to remove versions that are no longer necessary.'

src/config/kernel.md
```diff
@@ -76,15 +88,27 @@ omit_drivers+=" radeon "
...
+next reboot:
 
 ```
 # dracut --force
 ```
 
 #### mkinitcpio
 
-XXX: add example of blacklisting kernel modules for mkinitcpio
+To blacklist modules from being included in a mkinitcpio initramfs
```
bobertlo: This is all duplicated from a previous section. You could just reference it, like: 

'After configuring modules to be blacklisted, some may still be loaded by the `initramfs`, in which case a new `initramfs` needs to be generated for the changes to take effect:'

bobertlo: Or I guess modules could be *added* also, but something like that.

src/config/kernel.md
```diff
@@ -21,28 +21,40 @@ modules.
...
+that is currently booted:
 
 ```
 # vkpurge rm all
 ```
 
 ## Kernel modules
 
+The Linux kernel has modular components which are called modules.
+Kernel modules are typically drivers for devices or filesystems.
```
bobertlo: I don't think these first two lines are necessary for the target audience, and especially not the first one. This can be really shortened or removed.

src/config/kernel.md
codingHahn: In my opinion the duplication justified, because those using mkinitcpio would just skip over the dracut part.

src/config/kernel.md
codingHahn: I agree that the first sentence is superfluous.

src/config/kernel.md
```diff
@@ -76,15 +90,27 @@ omit_drivers+=" radeon "
...
 ```
 
 #### mkinitcpio
 
-XXX: add example of blacklisting kernel modules for mkinitcpio
+To blacklist modules from being included in a mkinitcpio initramfs a `.conf`
+file needs to be created like `/etc/modprobe.d/radeon.conf` with the contents:
+
+```
+blacklist radeon
```
bobertlo: This is just literally a copy of 2 headings ago. Could we just compromise and say something like:

To blacklist modules in a mkinitcpio initramfs, create a `.conf` file in `/etc/modprobe.d/` as above, then regenerate the image:

```
# mkinitcpio -p linux
```

The changes will take effect on the next reboot.

bobertlo:
I just have the one last comment. Other than this it looks good but please be sure to squash your commits into one for the pull request. A guide has been added to the contributing section of the docs.

bobertlo:
Sorry I thought the cpio section was duplicated but I was confused. I'm squashing the commits and merging.

PR #110, opened 2019-08-21
create macbook.md

nsklaus:
this helps macbook users for a few tweaks needed post-install

### DIFF
--- a/src/installation/guides/macbook.md
+++ b/src/installation/guides/macbook.md
@@ -1,0 +1,48 @@
+# Macbookpro 12,1 (early 2015)
+
+Everything is well supported on this laptop and install will work without problems. Post-installation, after having booted to your newly installed system, there is however a few points that need a little bit of work.
+
+## keymap are wrong
+This is a problem inherited by a few distros, including archlinux.
+tty keymaps for the macbooks (mac-fr, mac-us, mac-de) are completely broken and *you should not attempt to use them*, or it will result in an unusable keyboard and you will be forced to do a hard reset on your system. as a workaround generic keymaps
+work well (fr, us, de ..) you can use those for starter. later on you can find working ones and install/use them. here i've found a working one for the french macbook keyboards: [forums.archlinux.fr](https://forums.archlinux.fr/viewtopic.php?t=8679)
+
+```
+cd /usr/share/kbd/keymaps/mac/all
+# wget ftp://ftp.linux-france.org/pub/macintosh/mac-fr-ext_new.map.gz
+# localectl set-keymap mac-fr-ext_new
+```
+if that works for you you can add it to the `/etc/rc.conf` modify that line accordingly:
+```
+# Keymap to load, see loadkeys(8).
+KEYMAP=mac-fr-ext_new
+```
+
+- fonts are too small
+you need to install the `terminus-font` package, then use it like shown below:
+```
+sudo xbps-install -Su terminus-font
+setfont ter-u32b
+```
+if that works well for you add it to the `/etc/rc.conf` file:
+```
+# Console font to load, see setfont(8).
+FONT="ter-u32b"
+
+```
+- mouse pointer only move up and down
+after having Xorg installed, the first time you will start it you will noticed your pointer only moves up and down when using the touchpad. this is because the usbmouse module is getting loaded and used instead of the bcm5974 device (the touchpad). as a workaround you can do this:
+
+```
+echo "blacklist usbmouse" | sudo tee /etc/modprobe.d/usbmouse.conf
+sudo dracut --force --add-drivers bcm5974 --omit-drivers usbmouse
+```
+
+then reboot and try to start xorg again. if that works well for you, you can make that permanent like so:
+
+```
+echo 'omit_drivers+="usbmouse"' | sudo tee /etc/dracut.conf.d/disable-usbmouse.conf
+echo 'drivers+="bcm5974"' | sudo tee /etc/dracut.conf.d/bcm5974.conf
+```
+
+so when you get your kernel updated you won't have to run `sudo dracut --force --add-drivers bcm5974 --omit-drivers usbmouse` each time


src/installation/guides/macbook.md
```diff
@@ -0,0 +1,53 @@
+# Macbookpro 12,1 (early 2015)
```
nilium: Is this specific to only one model?

Also, trailing whitespace here (if you end up with two spaces, it is significant in Markdown, so it should be avoided).

src/installation/guides/macbook.md
```diff
@@ -0,0 +1,53 @@
...
+
+Everything is well supported on this laptop and install will work without problems. Post-installation, after having booted to your newly installed system, there is however a few points that need a little bit of work.
+
+## keymap are wrong
+This is a problem inherited by a few distros, including archlinux.
+tty keymaps for the macbooks (mac-fr, mac-us, mac-de) are completely broken and *you should not attempt to use them*, or it will result in an unusable keyboard and you will be forced to do a hard reset on your system. as a workaround generic keymaps
+work well (fr, us, de ..) you can use those for starter. later on you can find working ones and install/use them. here i've found a working one for the french macbook keyboards: [forums.archlinux.fr](https://forums.archlinux.fr/viewtopic.php?t=8679)
+
+```
+cd /usr/share/kbd/keymaps/mac/all
```
nilium: Code snippets should follow the style guide: https://docs.voidlinux.org/contributing/void-docs/style-guide.html

Use `$` for unprivileged commands, `#` for privileged commands. In this case, there's also a question of whether the user should be modifying /usr/share or if /usr/local/share is the appropriate place for this.

src/installation/guides/macbook.md
```diff
@@ -0,0 +1,53 @@
...
+wget ftp://ftp.linux-france.org/pub/macintosh/mac-fr-ext_new.map.gz
+localectl set-keymap mac-fr-ext_new
```
nilium: I don't belive localectl is available on Void. This is part of systemd, which Void doesn't currently provide.

src/installation/guides/macbook.md
```diff
@@ -0,0 +1,53 @@
...
+Everything is well supported on this laptop and install will work without problems. Post-installation, after having booted to your newly installed system, there is however a few points that need a little bit of work.
+
+## keymap are wrong
+This is a problem inherited by a few distros, including archlinux.
+tty keymaps for the macbooks (mac-fr, mac-us, mac-de) are completely broken and *you should not attempt to use them*, or it will result in an unusable keyboard and you will be forced to do a hard reset on your system. as a workaround generic keymaps
+work well (fr, us, de ..) you can use those for starter. later on you can find working ones and install/use them. here i've found a working one for the french macbook keyboards: [forums.archlinux.fr](https://forums.archlinux.fr/viewtopic.php?t=8679)
+
+```
+cd /usr/share/kbd/keymaps/mac/all
+wget ftp://ftp.linux-france.org/pub/macintosh/mac-fr-ext_new.map.gz
```
nilium: If this is something we want to support, I think you need to submit package templates to provide Mac keymaps. We shouldn't have documentation asking users to fetch files from an arbitrary FTP server. Otherwise, this entire section can't be accepted.

src/installation/guides/macbook.md
```diff
@@ -0,0 +1,53 @@
...
+```
+
+if that works for you you can add it to the `/etc/rc.conf` modify that line accordingly:
+
+```
+# Keymap to load, see loadkeys(8).
+KEYMAP=mac-fr-ext_new
+```
+
+## fonts are too small
```
nilium: This isn't really particular to MacBooks or Void, it's just the nature of high-density screens.

src/installation/guides/macbook.md
```diff
@@ -0,0 +1,53 @@
...
+on macbookpro the hidpi screen makes the font really small and hard to read. the default resolution is 2560x1600. you need to install the `terminus-font` package, then use it like shown below:
+
+```
+sudo xbps-install -Su terminus-font
```
nilium: 1. Please follow the style guide here.
2. Is it possible, instead of telling users to install a specific font, to suggest that they look at setting a new terminal font with suggestions? We also likely don't need to tell them how to install a package here.

src/installation/guides/macbook.md
```diff
@@ -0,0 +1,53 @@
...
+```
+# Console font to load, see setfont(8).
+FONT="ter-u32b"
+```
+
+## mouse pointer only move up and down
+after having Xorg installed, the first time you will start it you will noticed your pointer only moves up and down when using the touchpad. this is because the usbmouse module is getting loaded and used instead of the bcm5974 device (the touchpad). as a workaround you can do this:
+
+```
+echo "blacklist usbmouse" | sudo tee /etc/modprobe.d/usbmouse.conf
```
nilium: Instead of using a weird sudo/tee combination, maybe just tell users to blacklist usbmouse by creating a .conf file with `blacklist usbmouse` under /etc/modprobe.d. Stuff like this should be discouraged when it can be replaced with a clear, simple explanation.

src/installation/guides/macbook.md
```diff
@@ -0,0 +1,53 @@
...
+sudo dracut --force --add-drivers bcm5974 --omit-drivers usbmouse
+```
+
+then reboot and try to start xorg again. if that works well for you, you can make that permanent like so:
+
+```
+echo 'omit_drivers+="usbmouse"' | sudo tee /etc/dracut.conf.d/disable-usbmouse.conf
+echo 'drivers+="bcm5974"' | sudo tee /etc/dracut.conf.d/bcm5974.conf
```
nilium: Ditto on just telling the user to create the file.

src/installation/guides/macbook.md
```diff
@@ -0,0 +1,53 @@
+# Macbookpro 12,1 (early 2015) 
+
+Everything is well supported on this laptop and install will work without problems. Post-installation, after having booted to your newly installed system, there is however a few points that need a little bit of work.
+
+## keymap are wrong
+This is a problem inherited by a few distros, including archlinux.
```
nilium: This line isn't important.

src/installation/guides/macbook.md
```diff
@@ -0,0 +1,53 @@
...
+tty keymaps for the macbooks (mac-fr, mac-us, mac-de) are completely broken and *you should not attempt to use them*, or it will result in an unusable keyboard and you will be forced to do a hard reset on your system. as a workaround generic keymaps
```
nilium: This could be condensed to "The `mac-` prefixed keymaps may not work and using them may require you to replace them later." I'm not sure what you mean by "hard reset" here.

src/installation/guides/macbook.md
```diff
@@ -0,0 +1,53 @@
...
+work well (fr, us, de ..) you can use those for starter. later on you can find working ones and install/use them. here i've found a working one for the french macbook keyboards: [forums.archlinux.fr](https://forums.archlinux.fr/viewtopic.php?t=8679)
```
nilium: The documentation doesn't have an "I," so "I've found" isn't the tone that documentation should have. Random forum posts aren't a great source of information here either. If asking people to install something, it really needs to be packaged.

src/installation/guides/macbook.md
nsklaus: i don't have many laptop, at least that's what i must do on mine to have it working.
it is also similar to what was described on the old wiki 'macbook' page. i believe he had a slightly different model than mine, but we got the same 'mouse problem'.
(pointer only moving up and down).

about the formating, here's my comment: it's a pain in the put, these modern tools.

src/installation/guides/macbook.md
nsklaus: thanks for providing a link to the style guide.
i'll modify the snippets, here i had used # as a comment

src/installation/guides/macbook.md
nsklaus: yes, i'll modify that it's not localectl but instead: loadkeys.
i mixed it, sorry for the confusion.

src/installation/guides/macbook.md
nilium: I've encountered the mouse issue on 2015 models, but what this document should maybe be is a series of problem/solution sections instead of targeting a specific model. Models can be called out if it's truly model-specific.

src/installation/guides/macbook.md
nsklaus: i'd be glad if you updated all these broken keymaps. this situation has been going on since 2015, at least on arch.

basicaly all mac-us, mac-fr, mac-de tty keymaps are broken and loading them make the keyboard unusable. hard-reset required.
by search and luck i did find this one i'm linking to. it works perfectly.

i have no experience in package making. i can learn and contribute. but don't wait for me, someone has to replace these broken keymaps before the new year 2016 ..oh wait it's 2019 now ;)

src/installation/guides/macbook.md
nsklaus: well, since i was at it and it was needed anyway, i thought it wouldn't hurt to add this to the page so the person reading it won't have to jump back and forth between pages.

src/installation/guides/macbook.md
nsklaus: please do so. i'm just struggling to make it work nicely, and i just thought about sharing my progress in the docs, but packaging working keymaps would be indeed better.

ericonr:
Hey there! For a PR affecting mostly a single file, you need to have a single commit. For this, you need to squash the commits, which you can do using `git rebase -i`.

You also need to run `ci/format.sh` to format the files so they pass the checks :)

These are some basic guidelines, then we can look into working on the rest of the PR.

ericonr:
Ping @nsklaus 

Hi there! Are you still interested in working on this? I would be interested in helping you out with formatting and wording. I think not a lot of info here is Void specific, and the page probably fits better outside the configuration section, but there is some information that seems hard to gather from other places, and I think it would make sense to include.

What do you think?

src/installation/guides/macbook.md
```diff
@@ -0,0 +1,69 @@
...
+```
+# Console font to load, see setfont(8).
+FONT="ter-u32b"
+```
+
+## mouse pointer only move up and down
+after having Xorg installed, the first time you will start it you will notice
+your pointer only moves up and down when using the touchpad. this is because the
+usbmouse module is getting loaded and used instead of the bcm5974 device (the
+touchpad). as a workaround you can do this:
```
ahesford, line 53: The symptom I saw was a completely dead trackpad, not restricted movement. Change to something like, "By default, the `usbmouse` kernel module may be loaded at boot and bind to your trackpad, rendering it unusable. If you experience problems with your trackpad, blacklist the `usbmouse` module [...]"

nsklaus:
> Ping @nsklaus
> 
> Hi there! Are you still interested in working on this? I would be interested in helping you out with formatting and wording. I think not a lot of info here is Void specific, and the page probably fits better outside the installation section, but there is some information that seems hard to gather from other places, and I think it would make sense to include.
> 
> What do you think?

i reply to clarify the situation. your offer is nice thank you for that.
but it's a little late now. i am not using void linux anymore at this moment.
i don't wish to bash void linux, i think it is a nice distro, but i had too much little problems with it.
i think it's the combination of packages versions and my specific hardware that caused them.
anyway since i'm not on void, i cannot tell if current situation improved or not.
i cannot tell if parts of the problems i reported are now solved or not.
for what it's worth, i tested manjaro linux lately and i can say, the "mac-fr" and "mac-us" tty keyboard inherited from arch linux are still completely broken, so the "mac-fr-ext_new.map.gz" file i pointed to, at least still seems necessary.
you'd be better off grabbing the content of this PR and make a new one from it.
almost one year went by since the submission of this.
everything needs to be re-tested again with void's current packages offering and compatibility with that particular hardware ("macbookpro12,1" early 2015 model).

maybe it would have been better at the time if what i had made was just accepted as it was, and then being improved upon by anyone seeing necessity to do so, instead of blocking it until it is perfect for inclusion. since: unperfect doc is better than no doc.

but again, i'm not bashing on void, i still think it is a nice distro and i might test it again at some point. until that time come though, i cannot finish this attempt at filling documentation for it.

ericonr:
That's unfortunate, but I understand. Thanks for letting me know, I will see if I can move the contents somewhere! Take care c:

ericonr:
Another note, seems like some drivers are broken (for now?) with `wpa_supplicant`, but work with `iwd`.

nsklaus:
> Another note, seems like some drivers are broken (for now?) with `wpa_supplicant`, but work with `iwd`.

for what it's worth, i'm using artix at the moment, and everything works for my particular hardware.
my wifi driver is brcmfmac.

also i have found the issue about bad keymap for tty: it is files missing in kbd package.
on devuan in console-data package, it have additional keymap that are missing in kbd.
"mac-macbook-fr" (there is both "de" and "fr" but unfortunately no "us").
after i picked up the "mac-macbook-fr" tty keymap from devuan and put it in artix, my tty keymap problem was instantly gone. all distro using kbd misses these.

ericonr:
Oh, interesting. Thanks!

PR #111, opened 2019-08-21
Create rosetta-stone.md

Wychmire:
Add the rosetta stone page to the XBPS section and update SUMMARY.md to include a link to that page

I did my best to find a reasonable way to include the two footnotes (both are in the last section, Building Packages) but if the way I chose doesn't work, let me know


pltrz:
To be honest, I'm not quite sure a rosetta stone is a great fit for this handbook. I know it was in the wiki, but some things on there fall out of the scope that we're aiming to keep on this handbook.

Perhaps this contribution might be useful [in this project](https://github.com/blalor/package-manager-rosetta-stone), though

Wychmire:
I reference it frequently as a method of looking up commands, but if it's out of scope then it shouldn't be included. I was thinking of trying the Arch Wiki, but since XBPS isn't widely used it might not be worth it. I'll take a look at the project you linked.

For future reference, is there somewhere that mentions what _is_ in scope?)

Duncaen:
I see it similar but because of the demand I wouldn't oppose.

Maybe its worth it to start the page with a paragraph that the shown commands are just a small guide and that reading the real manual pages is suggested because package manager do things different and sometimes translating an action from one to another is not always optimal.

Wychmire:
I've added a paragraph which should do what you want.
Before (and if) this gets merged it'd be great if someone or multiple someones could do a quick look-over to make sure I'm not missing anything or have incorrect info. I don't _think_ I do, but it'd be nice to be sure

flexibeast:
@Wychmire: Thanks for your work on this! A few thoughts:

* Typo in the opening paragraph: should be 'consistent' rather than 'consistant'.

* Re. the entry for "Display files provided by local package", the 'Debian/Ubuntu' column should possibly have `dpkg-query -L`, rather than `dpkg -L`. (The `dpkg(1)` [man page](https://linux.die.net/man/1/dpkg) lists the `-L` option under "dpkg-query options".)

* Re. the entry for "Displays packages which provide the given exp. aka reverse provides", with `apt-file search` as the Debian/Ubuntu equivalent: in these contexts, i have been using `xbps-query -o` (if i'm looking for which *installed* package owns a file) or `xlocate` (if i'm trying to find which *uninstalled* package owns a file). Also, for consistency with other entries, i suggest changing the text of the first column to "*Display* packages which ..." rather than "Displays".

* Re. the entry for "List the files that the package holds. Again, this functionality can be mimicked by other more complex commands": i would suggest `xbps-query -f` be added as the equivalent of `dpkg-query -L` here, and that the sentence "Again, this functionality ..." therefore be removed from the first column.

Wychmire:
- Fixed the spelling of "consistent" and "separate" in the opening paragraph
- Done
- Done
- Done
- Also fixed a couple spots where piping something (i.e. to grep)  got made into a table row + removed the xlint footnote and replaced it with "xlint (from package xtools)" to match the Red Hat / Fedora entries

bobertlo:
I feel similarly also about this page. I'm not sure it's a perfect fit but also there is maybe enough demand to include it. I will try to review this properly as soon as I catch up more on my backlog.

flexibeast:
i haven't seen much demand on r/voidlinux or #voidlinux for this sort of document. @void-linux/doc-writers, should we just close this?

Wychmire:
I'm no longer interested in making sure this stays up to date, it's probably a good idea to close it so it doesn't get included and end up woefully out of date.

ericonr:
I will close it, then.

@Wychmire thanks for the contribution, though! If you want to contribute to some Void specific tooling, we have some work ongoing on #260 and #258, at the very least. I'm also thinking that we can try to document non obvious operations a bit more as well, so if you have some specific ideas, we are open to new issues or PRs!

PR #114, opened 2019-09-03
Add note about handling pre-existing partition layout.

flexibeast:
i recently found [this section of the old wiki](https://wiki.voidlinux.org/Installation#Cfdisk_does_not_give_me_a_choice_of_partition_table_type) helpful, so would like this information added to the Handbook. :-)


Duncaen:
Maybe mention that `wipefs(1)` can be used before starting the installer if the sole purpose is to throw away the previous partition table.

flexibeast:
@Duncaen: Thanks for the suggestion; i've amended the PR accordingly.

bobertlo:
Looks good to me.

PR #119, opened 2019-09-17
xbps/mirrors: merge list into index; cleanup

bobertlo:


bobertlo:
@the-maldridge is the geo-ip functioning right now? IIRC it was up in the air when this was written and I'm not sure it was ever put into production?

bobertlo:
@nilium I'd like to take a pass at a lot of the sections and I really respect your English skills. Do you have spare time to proofread to do some proofreading? I'd rather not just post changes without someone reviewing. If you're willing to do reviews I'll wait on them.

bobertlo:
I asked on IRC at a more decent hour and verified the mirrors are not geographically routed.

PR #120, opened 2019-09-27
theme: Add cursor: pointer to .icon-button

jwkicklighter:
The sidebar-toggle didn't have pointer-style cursor, even though it's a clickable element. I imagine all `.icon-button` elements should have this, so I stuck it there.


Duncaen:
Thanks, merged.

PR #126, opened 2019-10-30
Add section about downgrading packages.

flexibeast:
Having recently had to downgrade `dhcpcd`, i thought it might be useful to port the wiki information about downgrading across to the handbook. :-)

### DIFF
--- a/src/xbps/packages/downgrading.md
+++ b/src/xbps/packages/downgrading.md
@@ -1,0 +1,35 @@
+# Downgrading to a specific package version
+
+## Via xdowngrade
+
+The easiest way to downgrade is to use `xdowngrade` from the `xtools` package,
+specifying the package version you wish to downgrade to:
+
+```
+# xdowngrade /var/cache/xbps/pkg-1.0_1.xbps
+```
+
+## Via XBPS
+
+XBPS can be used to downgrade to a package version that is no longer available
+in the repository index.
+
+If the package version had been installed previously, it will be available in
+`/var/cache/xbps/`. If not, it will need to be obtained from elsewhere; for the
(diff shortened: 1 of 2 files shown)


src/xbps/packages/downgrading.md
```diff
@@ -0,0 +1,20 @@
+# Downgrading to a specific package version
+
+The easiest way to downgrade is by using `xdowngrade` from the `xtools` package.
+
```
Vaelatern: Please make a clear separation, making it clear this is the end of this subsection about `xdowngrade`. Maybe there is a bit of usage you can duplicate into here too?

flexibeast: Thanks for the feedback - PR updated.

src/xbps/packages/downgrading.md
```diff
@@ -0,0 +1,28 @@
...
+The easiest way to downgrade is to use `xdowngrade` from the `xtools` package,
+specifying the package version you wish to downgrade to:
+
+```
+# xdowngrade /var/cache/xbps/pkg-1.0_1.xbps
+```
+
+## Via XBPS
+
+To downgrade using just XBPS, you must first add the desired package version to
```
bobertlo: To be clear, this is adding a locally cached older version of the package which is no longer contained in the repository index, correct? (or could be used if you got an older version somewhere else too.)

bobertlo:
I think that one point could use clarification for the reader. Otherwise LGTM

src/xbps/packages/downgrading.md
flexibeast: Thanks for the feedback - PR updated.

Issue #122, opened 2019-10-21
runit could use a more general explaination


bobertlo:
@Duncaen mentioned this in irc and I think it is a good idea. Right now we have a short quick usage guide in the header of the `Configuration > Services and Daemons` section. I think a possible layout could be:

```
- Configuration
   - Runit
      - Starting/Stopping Services
      - Configuring Services
   - Services and Daemons
      - Cron
      - ...
```

With an explaination of what runit is and why it is used in the heading, but I am very open to discussion on that point. There's probably a third and better solution?

Especially in the context of people viewing Void Linux as an "anti-systemd" distro, I think we could mention how we are more "pro-runit" and dropped systemd promarily over compatibility issues.

bobertlo:
Now that I look at the docs again, I think a good writeup about runit might be better suited to go in the initial `About` section.

flexibeast:
i can do the 'Runit' subsection of 'Configuration'; i assume the more general writeup about why Void is pro-Runit would be best left for someone on the Void dev team.

flexibeast:
So should i simply port [the 'Runit' page on the wiki](https://wiki.voidlinux.org/Runit)?

bobertlo:
I just took a quick look at the wiki page. That seems like a good start. I do generally like the layout of the wiki page more, but some things could be organized better. I'd really like if that information could be split into at least a couple pages. Maybe something like:

- runit
  - Basic Usage
  - Creating/Modifying Services

Or there could be `managing services` and `configuring services`?

Basically it looks like if we get everything from both the docs and the wiki merged into a more organized layout we're in good shape. I'm just brainstorming.

flexibeast:
So the stuff currently on [the "Services and daemons" page of the handbook](https://docs.voidlinux.org/config/services/index.html) would be removed, with that information instead being available via "Basic Usage" and "Creating/Modifying Services" pages under a "runit" (or maybe "runit: services and daemons"?) subsection of the handbook? Into which we could add anything from the wiki not already included?

bobertlo:
Right, that is what I was thinking. `Services and daemons` would stay, but for clarity would be a separate section from runit.

flexibeast:
With #127 now merged, should this issue be closed, or should it remain open until

> we could mention how we are more "pro-runit" and dropped systemd promarily over compatibility issues

has been addressed?

bobertlo:
I think #131 should take care of the rest of this.

PR #127, opened 2019-10-31
Add 'runit' section.

flexibeast:
This is a draft PR for issue #122, intended to provide something concrete for discussion; it reflects my current (possibly incorrect) understanding of @bobertlo's suggestions.

Some things that came to mind when putting this together:

* Should the existing ["Per-User Services" page in the handbook](https://docs.voidlinux.org/config/services/user-services.html) be removed, and the contents added as a subsection of [this PR's "Configuring Services" page](https://github.com/void-linux/void-docs/pull/127/files#diff-29a8d9af5e4959af8c958c6948b03c9f)?

* Further to the opening comment of #122:

> we could mention how we are more "pro-runit" and dropped systemd promarily over compatibility issues

i have, as a starting point, described some of the positive aspects of runit in [this PR's `runit/index.md` page](https://github.com/void-linux/void-docs/pull/127/files#diff-d39d11209fae69d339e13300801e9fbb), based on [the "Benefits" page of the runit site](http://smarden.org/runit/benefits.html), but this will need to be expanded by Void devs with the relevant knowledge and experience.

* i was recently trying to specify the timeout value for one specific service to be considered 'up' on boot, but was unable to work out how to use either the `SVWAIT` environment variable or the `-w` option to `sv(8)` for this. Might it be helpful to include such information in the handbook?


src/config/runit/configuring.md
```diff
@@ -0,0 +1,44 @@
+# Configuring Services
+
+## Editing services
```
bobertlo: I like this, but we should probably include a section before this one that mentions the conf files, since most services have options for necessary configuration included in their run files. In that case, the user can just modify that file and keep using the service directory provided by the package manager.

src/config/runit/configuring.md
```diff
@@ -0,0 +1,44 @@
...
+After editing, the old service should then be stopped, disabled and replaced
+with the new one:
+
+```
+# sv down service_name
+# rm /var/service/service_name
+# ln -s /etc/sv/service_name_edited /var/service/
+```
+
+## Service directory layout
```
bobertlo: I think this could go in the index page? (Or *maybe* an 'overview' page?) I like having this overview of how runit works, but I don't think it belongs in either of the two current subsections.

src/config/runit/configuring.md
flexibeast: Thanks for the feedback - PR updated.

src/config/runit/configuring.md
flexibeast: Thanks for the feedback - PR updated.

bobertlo:
Looks good to me. Thanks for your work!

Issue #123, opened 2019-10-22
Section about date/time and ntp daemons


Duncaen:

flexibeast:
i'd like to work on this. i presume this would go in the 'Services and Daemons' section of the Handbook? What sort of information should be included?

nilium:
For NTP daemons, probably just very basic setup of the main NTP daemons (chrony, openntp, and ntpd). Specifically, package names and services to link. Don't think any of them require additional configuration. It wouldn't have to go into detail on more advanced usage, given that both have their own documentation.

Date/time is pretty broad, so I'm assuming the intent is to just cover how to set it, if one wanted to, and keep it synchronized with the daemons.

flexibeast:
With the merging of #132, can this now be closed?

Duncaen:
I feel like we are still missing something about rc.conf `HARDWARECLOCK` and timezones.

zenny:
> I think we are still missing something about rc.conf `HARDWARECLOCK` and timezones.

@Duncaen I guess so, too because the time shown is incorrect until ntp servers are used. Even after ntp servers deployed with correct `HARDWARECLOCK` and `TIMEZONE` in rc.conf, the time displayed is 5 minutes ahead of any other atomic clocks (like smartphones)!

flexibeast:
@zenny: My understanding is that if the hardware clock has drifted too far from the time reported by time servers, one needs to use something like `ntpdate` to set the hardware time; the ntp servers can then maintain the clock from there.

@Duncaen: If what i wrote above is correct, should this information be added to the Handbook?

bobertlo:
I think with the addition of https://github.com/void-linux/void-docs/pull/154 covers the last issue.

I just saw that these are duplicated between `Date & Time` and `rc files`, but that is a separate issue. I'm closing this.

PR #132, opened 2019-11-06
Add 'Date and Time' section.

flexibeast:
This PR addresses #123.

### DIFF
--- a/src/config/services/date-time.md
+++ b/src/config/services/date-time.md
@@ -1,0 +1,38 @@
+# Date and Time
+
+To maintain accuracy of your system's clock, you can use the [Network Time
+Protocol](https://en.wikipedia.org/wiki/Network_Time_Protocol) (NTP).
+
+Void provides packages for three NTP daemons: NTP, OpenNTPD and Chrony.
+
+Once you have installed an NTP daemon, you can [enable the
+service](../runit/managing.md).
+
+## NTP
+
+NTP is the official reference implementation of the Network Time Protocol.
+
+The `ntp` package provides NTP and the `isc-ntpd` service.
+
+For further information, visit [the NTP site](http://www.ntp.org/).
+
+## OpenNTPD
+
+OpenNTPD focuses on providing a secure, lean NTP implementation which "just
+works" with reasonable accuracy for a majority of use-cases.
+
+The `openntpd` package provides OpenNTPD and the `openntpd` service.
+
+For further information, visit [the OpenNTPD site](http://www.openntpd.org/).
+
+## Chrony
+
+Chrony is designed to work well in a variety of conditions; it can synchronise
+faster and with greater accuracy than NTP.
+
+The `chrony` package provides Chrony and the `chronyd` service.
+
+The Chrony site provides [a brief overview of its advantages over
+NTP](https://chrony.tuxfamily.org/faq.html#_how_does_code_chrony_code_compare_to_code_ntpd_code),
+as well as [a detailed feature comparison between Chrony, NTP and
+OpenNTPD](https://chrony.tuxfamily.org/comparison.html).
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -20,6 +20,7 @@
       - [Cron](./config/services/cron.md)
       - [Logging](./config/services/logging.md)
       - [Console](./config/services/console.md)
+      - [Date and Time](./config/services/date-time.md)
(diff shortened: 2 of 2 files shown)


src/config/services/date-time.md
```diff
@@ -0,0 +1,61 @@
...
+
+To maintain accuracy of your system's clock, you can use the [Network Time
+Protocol](https://en.wikipedia.org/wiki/Network_Time_Protocol) (NTP).
+
+Void provides packages for three NTP daemons: NTP, OpenNTPD and Chrony.
+
+Once you have installed an NTP daemon, it can be enabled with:
+
+```
+# ln -s /etc/sv/ntpd /var/service/
```
nilium: Because people will verbatim copy/paste things without looking, maybe just say that you should enable the service you chose and link to the doc for enabling services. I don't know if we need to repeat the steps for enabling a service here. Documenting the `ntpd` symlink is still important, but should include some mention of it being part of package alternatives, and that users need to verify which alternative is chosen before enabling `ntpd` instead of using the actual service name, such as `chronyd` or `busybox-ntpd`.

Thinking about that, it might be better to tell people the real service name of each daemon below and not have them rely on an ntpd symlink. This is fairly important given that you can also install something like busybox and this will provide another `ntpd` alternative. This could confuse folks who install something and don't expect it to provide an ntpd symlink (confusion is probably the norm for packages using alternatives).

bobertlo: I would also like to add that talking about enabling the service (and information on how to) can just be done once, and does not need to be repeated after the description of each different package.

flexibeast: @nilium, @bobertlo: Thank you both for your feedback. :-)

Since i didn't mention enabling/disabling the service in the description for each of NTP / OpenNTPD / Chrony, i presume you mean that i should just remove the enabling/disabling section beginning with the sentence "Once you have installed an NTP daemon ...", since enabling and disabling services is now described in runit/managing?

Then, instead of talking about the symlinks from `ntpd` to `<service>`, i should simply note the service name for each of NTP / OpenNTPD / Chrony?

nilium: re: Enabling / disabling, yes. You can just tell the users to enable the service of their choice and link to the doc on enabling services, and call out the names of each service below. The doc isn't comprehensive (doesn't include busybox, for example), but that's fine as long as the common choices are covered.

flexibeast: Thanks - PR updated.

nilium:
I've been bogged down with work, so haven't had time to look at this, but it looks correct to me at a glance. Assuming it's fine by @bobertlo, it can be merged.

bobertlo:
The use of `xbps-install` code blocks is less than ideal but otherwise this looks good to me. I'll pull this soon if no one else objects. I'm really busy myself and this isn't strictly above the standards in the existing docs.

flexibeast:
@bobertlo: Thanks for the feedback. :-) For future reference, would you prefer that things like:

> To install OpenNTPD:
> ```
> # xbps-install openntpd
> ```
> This will install the `openntpd` service.

instead be something like:

 > The `openntpd` package provides OpenNTPD and the `openntpd` service.

bobertlo:
@flexibeast I do think that would be more in line with our style guide :)

I can't check that right now but I wonder if there is documentation of this in the style guide??

flexibeast:
Hm, i can't see anything in [the style guide](https://docs.voidlinux.org/contributing/void-docs/style-guide.html) related to this .... Could the principle perhaps be described as "Don't write code blocks showing how to perform actions documented elsewhere, such as installing packages or enabling services"?

(i'm also happy to change the wording in this PR to the alternate version i suggested in my previous comment, unless people just want to merge this as-is for now.)

bobertlo:
Right, I didn't think it was in the style guide, but it has come up repeatedly. Upon looking at the style guide itself I do believe the examples could certainly be changed to *anything* other than xbps-install, but that would be a separate pull request. :)

Also, I would appreciate the wording change you suggested in your previous comment. We are discouraging any repeated commands that could just be described by 'install the `blank` package' in general.

flexibeast:
Thanks, PR updated in https://github.com/void-linux/void-docs/commit/46b7de8dc1a75a5bfd3d689becb7a8cce2afa5a2.

i'll create a separate PR for changes to `style-guide.md`. :-)

bobertlo:
Looks good. Thanks :)

PR #134, opened 2019-11-12
Added eduroam configuration entry.

leonardlorenz:
I was talking to bobertlo about this issue. I found it especially hard to use eduroam with voidlinux since I didn't know which packages and services were needed to use the scripts they provide.


leonardlorenz:
I slimmed it down and added a brief description. Is there anything I can do to further improve it?

bobertlo:
So, I just asked for a review in IRC (since I don't use Eduroam) and @Duncaen suggested it just be added as a section to Network Manager. I didn't realize you could connect to Eduroam with other tools (like wpa_supplicant) but now, realizing this, I have to agree. Maybe this could all be summarized in a section in the `NetworkManager` page instead?

Sorry for more changes. Otherwise this looks good to me.

leahneukirchen:
FWIW, eduroam works fine without NetworkManager or dbus, it just requires an appropriate wpa_supplicant.conf (using WPA Enterprise).

leonardlorenz:
I agree with @Duncaen, it does make sense to move it into networkmanager.md. The issue is NetworkManager specific after all.

bobertlo:
Oops, I didn't see you sent that last commit :)

Issue #129, opened 2019-11-03
Document intel-ucode and other firmware packages


Duncaen:

bobertlo:
Can this be closed?

ericonr:
@Duncaen can we close this one? Or is there something we still need to include?

Duncaen:
I think it would be good to make a firmware section and move the microcode stuff into it.
There are a few other firmware packages that are not installed by default and I think because the question in how to get rid of firmware packages installed by default it would make sense to add a paragraph about that and link to `ignorepkg` documentation for xbps.

flexibeast:
#368 merged. Closing.

PR #137, opened 2019-11-24
kernel.md: added microcode installation instructions to kernel section

manneorama:
Part of issue #129 
Unsure what is meant by "other firmware", also unsure if kernel.md is the correct place but it felt like the most suitable without creating a new section.

### DIFF
--- a/src/config/microcode.md
+++ b/src/config/microcode.md
@@ -1,0 +1,50 @@
+# Microcode
+
+Microcode is loaded onto the CPU or GPU at boot by the BIOS, but can be replaced
+later by the OS itself. An update to microcode can allow a CPU's or GPU's
+behavior to be modified to work around certain yet to be discovered bugs,
+without the need to replace the hardware.
+
+## Installation
+
+### Intel
+
+The Intel microcode package is in the non-free repo, so that has to be enabled
+first.
+
+```
+# xbps-install -S void-repo-nonfree
+```
+
+Install the `intel-ucode` package.
+
+```
+# xbps-install -S intel-ucode
+```
+
+Rebuild the initramfs matching the kernel. This only has to be done once, the
+first time the `intel-ucode` package is installed. If the current bootloader
+configuration contains more than one kernel version, the initramfs has to be
+rebuilt for each of them.
+
+```
+# xbps-reconfigure -f linuxX.X
+```
+
+### AMD
+
+The AMD package contains microcode for both AMD CPUs and GPUs. AMD CPUs and GPUs
+will automatically load the microcode, no further configuration required.
+
+```
+# xbps-install -S linux-firmware-amd
+```
+
+## Verification
+
+Performing the following command after a reboot can be used as verification that
+the update worked
+
+```
+$ grep microcode /proc/cpuinfo
+```
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -46,6 +46,7 @@
       - [Installing Printing Drivers](./config/print/drivers.md)
       - [Configuring a New Printer](./config/print/config.md)
    - [Manual Pages](./config/man.md)
+   - [Microcode](./config/microcode.md)
 - [XBPS Package Manager](./xbps/index.md)
    - [Updating](./xbps/updating.md)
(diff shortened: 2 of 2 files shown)


src/config/kernel.md
```diff
@@ -51,6 +51,55 @@ that is currently booted:
 # vkpurge rm all
 ```
 
+## Microcode
+
+Microcode is loaded onto the CPU or GPU at boot, by the BIOS, but can be
+replaced later in the boot stage by the OS itself. An update to microcode can
+allow a CPUs or GPUs low level behavior to be modified to work around certain
+yet to be discovered bugs, without needing to replace the hardware.
```
Piraty: suggestion:

> <...> or GPU at boot by the BIOS, but <...> replaced later by the OS itself. <...> CPU's or GPU's behavior <...>, without the need to replace the hardware.

src/config/kernel.md
```diff
@@ -51,6 +51,55 @@ that is currently booted:
...
+# xbps-install -S linux-firmware-amd-<date>
+```
+
+#### Verifying
+
+Performing the following command before and after a reboot can be used as
+verification that the update worked
+
+```
+$ cat /proc/cpuinfo | grep microcode
```
Piraty:
```
$ grep microcode /proc/cpuinfo
```

src/config/kernel.md
```diff
@@ -51,6 +51,55 @@ that is currently booted:
...
+# xbps-reconfigure -f linux-X.X.X_X
+```
+
+#### AMD
+
+The AMD package contains microcode for both AMD CPUs and GPUs. AMD CPUs and GPUs
+will automatically load the microcode, no further configuration required.
+
+```
+# xbps-install -S linux-firmware-amd-<date>
```
Piraty: the package is named `linux-firmware-amd` (subpackage of [linux-firmware](https://github.com/void-linux/void-packages/tree/aa0d17eab9b00bf193d2bcf6ff6475b7ad731c78/srcpkgs/linux-firmware/template)), no need to append the version.

src/config/kernel.md
```diff
@@ -51,6 +51,55 @@ that is currently booted:
...
+
+```
+# xbps-install -S intel-ucode
+```
+
+Then reconfigure the linux kernel. This only has to be done once, the first time
+the `intel-ucode` package is installed.
+
+```
+# xbps-reconfigure -f linux-X.X.X_X
```
Piraty: the kernel package is named `linuxX.Y`, no need to append the revision

src/config/kernel.md
```diff
@@ -51,6 +51,55 @@ that is currently booted:
...
+# xbps-install -S void-repo-nonfree
+```
+
+Install the `intel-ucode` package
+
+```
+# xbps-install -S intel-ucode
+```
+
+Then reconfigure the linux kernel. This only has to be done once, the first time
```
Piraty: The intend is to rebuild the initramfs matching the given kernel, you could mention that so that it's clear all existing initramfs files (for other installed kernel packages) need to be rebuilt in order to properly load the microcode

Piraty:
I think microcode might deserve it's own page, because i don't see it fit on the kernel page...

manneorama:
I agree. I’ll create a new page under config.

src/config/microcode.md
```diff
@@ -0,0 +1,51 @@
...
+# xbps-install -S intel-ucode
+```
+
+Rebuild the initramfs matching the kernel. This only has to be done once,
+the first time the `intel-ucode` package is installed. If the current bootloader
+configuration contains more than one kernel version, the initramfs has to be
+rebuilt for each of them.
+
+```
+# xbps-reconfigure -f linux-X.X
```
Duncaen: this should be `linuxX.X`, with the `-` its re configuring the meta package, which does nothing.

manneorama:
Sorry for the tardiness. Life got in the way.

Didn't know whether to squeeze the microcode section in between existing sections or not so I added it to the end of the config section.

xtraeme:
It's OK buddy.

PR #140, opened 2019-12-28
config/graphics/optimus: add to SUMMARY.md

st3r4g:
@xtraeme 
The new section is not showing up, forgot to add to `SUMMARY.md`


xtraeme:
@st3r4g please also document wayland (just briefly, required mesa stuff for intel/amd and then nvidia)

xtraeme:
@st3r4g also NVIDIA optimus is too specific, DRI PRIME is generic and works with intel, amd and nvidia... so it's not really tied to intel+NVIDIA, could intel+amd, or amd+nvidia rofl

xtraeme:
well, maybe we can come up with better PRs later on... ty

PR #145, opened 2020-01-24
Update user-services.md

gt7-void:
Changes the example run script:
 - take username from the service name
 - export USER and HOME

### DIFF
--- a/src/config/services/user-services.md
+++ b/src/config/services/user-services.md
@@ -13,7 +13,13 @@ As an example, create a service `/etc/sv/runsvdir-<username>` with the following
 ```
 #!/bin/sh
 
-exec chpst -u "<username>:$(id -Gn <username> | tr ' ' ':')" runsvdir /home/<username>/service
+export USER="${PWD##*-}"
+export HOME="$(getent passwd "$USER" | cut -d: -f6)"
+
+groups="$(id -Gn "$USER" | tr ' ' ':')"
+svdir="$HOME/service"
+
+exec chpst -u "$USER:$groups" runsvdir "$svdir"
 ```
 
 In this example [chpst(8)](https://man.voidlinux.org/chpst.8) is used to start a
--- a/src/config/services/user-services.md
+++ b/src/config/services/user-services.md
@@ -13,8 +13,8 @@ As an example, create a service `/etc/sv/runsvdir-<username>` with the following
 ```
 #!/bin/sh
 
-export USER="${PWD##*-}"
-export HOME="$(getent passwd "$USER" | cut -d: -f6)"
+export USER="<username>"
+export HOME="/home/<username>"
 
 groups="$(id -Gn "$USER" | tr ' ' ':')"
 svdir="$HOME/service"
@@ -27,7 +27,8 @@ new [runsvdir(8)](https://man.voidlinux.org/runsvdir.8) process as the specified
 user. [chpst(8)](https://man.voidlinux.org/chpst.8) does not read groups on its
 own, but expects the user to list all required groups separated by a `:`. The
 `id` and `tr` pipe is used to create a list of all the user's groups in a way
-[chpst(8)](https://man.voidlinux.org/chpst.8) understands it.
+[chpst(8)](https://man.voidlinux.org/chpst.8) understands it. Note that we
+export `$USER` and `$HOME` because some user services may not work without them.
 
 The user can then create new services or symlinks to them in the
 `/home/<username>/service` directory. To control the services using the


ericonr:
The recommended way according to http://smarden.org/runit/faq.html#userservices is very close to the current one. I'd perhaps change the `group` determination part from the current one, at most.

gt7-void:
The important point of my suggestion is to export USER and HOME, because otherwise these variables are either unset or set to root. I don't remember what but something broke for me without this.

The rest of the proposed change is to replace the one-liner:
```
exec chpst -u "$USER:$(id -Gn "$USER" | tr ' ' ':')" runsvdir "$HOME/service"
```
by
```
groups="$(id -Gn "$USER" | tr ' ' ':')"
svdir="$HOME/service"
exec chpst -u "$USER:$groups" runsvdir "$svdir"
```
which makes the whole script easier to understand IMHO.

The trick of taking the username from the script directory is taken from `agetty`. Since the document already suggests placing the service under `/etc/sv/runsvdir-<username>`, this means there's no need to edit the `run` script at all.

src/config/services/user-services.md
```diff
@@ -13,7 +13,13 @@ following `run` script:
 ```
 #!/bin/sh
 
-exec chpst -u "<username>:$(id -Gn <username> | tr ' ' ':')" runsvdir /home/<username>/service
+export USER="${PWD##*-}"
```
Duncaen, line 16: I would prefer to not derive the user from PWD as this adds more moving parts to it, users might mess up.

ahesford: I think picking a username from PWD is OK as long as it doesn't replace the login-exported USER and isn't exported to the environment. This lets one generic `run` script be used for multiple users, the way the `agetty` services all link back to `agetty-generic/run`.

Thus, I would call the variable `user` or `_user` and do not export.

Duncaen:
> The trick of taking the username from the script directory is taken from `agetty`. Since the document already suggests placing the service under `/etc/sv/runsvdir-<username>`, this means there's no need to edit the `run` script at all.

This is more in our control, this "works" and "make sure" the services are created with the right pattern.

As example users with `-` would already be broken with the current pattern.

src/config/services/user-services.md
Duncaen: The whole point of the change is to export USER.

I think a generic runsvdir script is ok, but this requires more documentation for the details and I don't think its really worth it.

ahesford:
> > The trick of taking the username from the script directory is taken from `agetty`. Since the document already suggests placing the service under `/etc/sv/runsvdir-<username>`, this means there's no need to edit the `run` script at all.
> 
> This is more in our control, this "works" and "make sure" the services are created with the right pattern.
> 
> As example users with `-` would already be broken with the current pattern.

I overlooked this problem. Maybe hard-coding the username should be the recommended pattern, possibly with a brief mention about how this could be made user-agnostic in a fashion like `agetty-generic`.

src/config/services/user-services.md
gt7-void:
```suggestion
export USER="<username>"
```

gt7-void: Here's another, more robust, alternative:
```suggestion
export USER="${PWD##*/runsvdir-}"
```
Because of the `/` this works in any case. Keep the `##` so it works even if the path contains a directory named `runsvdir-*`.

Would that be ok? I can amend the commit with the version you prefer.

Duncaen: I personally would prefer hardcoded username because explaining generic services feels out of scope and unnecessary for most users.

gt7-void: I agree, done that.

src/config/services/user-services.md
```diff
@@ -13,7 +13,13 @@ following `run` script:
 ```
 #!/bin/sh
 
-exec chpst -u "<username>:$(id -Gn <username> | tr ' ' ':')" runsvdir /home/<username>/service
+export USER="<username>"
+export HOME="$(getent passwd "$USER" | cut -d: -f6)"
+
+groups="$(id -Gn "$USER" | tr ' ' ':')"
+svdir="$HOME/service"
```
ericonr, line 20: I feel like these are adding some unecessary indirection. Wouldn't the user who has a home directory with a different name from their username know what they need here? Exporting USER and HOME seems reasonable, though.

ericonr:
@ahesford's suggestion could be a great addition as well!

gt7-void:
Hi guys! I appreciate all the suggestions, and I understand that you want the documentation to be good and correct, and that's why it's a git repo and not a wiki anymore. I agree with all of that.

However good and correct is not the same as perfect, and yes, I'm sure this can be improved in some objective way (even though some of the points raised are subjective).

But the point is that what is now in the manual is *incorrect* because not exporting `USER` and `HOME` breaks things. Please concede to me that what I proposed is at least *less incorrect*.

Regarding hardcoding vs. obtaining the data using `id` or `getent`, I couldn't care less, this is just an example in the manual, it works and whoever reads it can decide to hardcode `HOME`, `groups` and `svdir` in whatever way they want.

I guess if one harcodes `USER`, there is no harm in also hardcoding `HOME`, while `groups` is more dynamic.

src/config/services/user-services.md
gt7-void:
```suggestion
export USER="<username>"
export HOME="/home/<username>"

groups="$(id -Gn "$USER" | tr ' ' ':')"
svdir="$HOME/service"
```

ericonr:
Your current version is very good, and it helps in documenting what each thing is, due to the variable names. I hadn't properly appreciated that not having HOME and USER set could be issues, either.

Would you be okay with adding a small explanation regarding the need to export HOME and USER? Like, literally just mention in the end of the paragraph explaining this script that they are exported because stuff can break without them. I can make an additional commit for that when merging later, if you don't have the time.

gt7-void:
> Your current version is very good, and it helps in documenting what each thing is, due to the variable names. I hadn't properly appreciated that not having HOME and USER set could be issues, either. Thanks for the contribution!
> 
> Would you be okay with adding a small explanation regarding the need to export HOME and USER? Like, literally just mention in the end of the paragraph explaining this script that they are exported because stuff can break without them. I can make an additional commit for that when merging later, if you don't have the time.

Done, I rebased everything into two commits (my original patch + changes after review).

ericonr:
Great, thank you!

PR #149, opened 2020-02-22
README.md: Greatly increased verbosity

markmcb:
This includes a more descriptive view of void-docs. It also contains a placeholder for a new "Guides" repository (i.e., the target for the wiki replacement docs). Moves "Void Docs" sections from Handbook to here as this seems the more appropriate place for them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,11 +1,202 @@
-# Void Docs
+# Void Documentation
 
-Welcome to the Void documentation. This repository contains the source data
-behind <https://docs.voidlinux.org/>. Contributing to this repository follows
-the same protocol as the packages tree.
+Welcome to void-docs, the documentation repository for the [Void Linux Project](https://github.com/void-linux). This repository contains the source data behind <https://docs.voidlinux.org/>. Contributing to this repository follows the same protocol as the packages tree.
 
-## Contributing
+## Documentation Types
 
-See the [Submitting
-Changes](https://docs.voidlinux.org/contributing/void-docs/submitting.html) page
-on the docs.
+Before contributing, ensure you understand the intended purpose and guidelines of each type documentation.
+
+### Handbook 
+
+The goal of the Void Handbook is to step through the most critical elements of Void Linux that most every user will want to consider while installing or operating Void. It is terse and directive in tone.
+
+When proposing new content, ensure it meets these criteria or it will be rejected:
+
+* applies to most or all Void users
+* focuses on critical setup and configuration
+* requires few or no dependencies beyond the base-system
+
+If you can't meet those criteria but still have useful information, then Guides (described below) are likely a better consideration.
+
+The source for the [Void Handbook](https://docs.voidlinux.org/) is [void-linux/void-docs](https://github.com/void-linux/void-docs/) on GitHub.
+
+### Guides
+
+The goal of Guides is to cover specific scenarios that may not be of interest to all users. Examples of Guide documents might include, "full disk encryption," "dual-booting with another OS," etc. 
+
+While Guides are curated, they are the least formal documentation in terms of both tone and directiveness. They offer advice more than truth, i.e., the reader must exercise judgment to decide if it is right for a specific instance of Void Linux. For example, "Updating Packages with XBPS" is more of a truth in Void and belongs in the Handbook, whereas "Booting from Mirrored Drives" is a judgment call and belongs in a Guide as not all users will want to do this.
+
+The source for [Void Guides](https://guides.voidlinux.org/) is [void-linux/void-guides](https://github.com/void-linux/void-guides/) on GitHub.
+
+### Manual Pages (outside the scope of void-docs)
+
+The goal of a manual page is to document atomic components of an operating system. Executable programs, system calls, libraries, configuration files, and kernel routines are a few of the things likely to have a manual page.
+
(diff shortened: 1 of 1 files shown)


markmcb:
As a side note, this is a result of a conversation with @Duncaen in IRC. I generally think there should be a clear delineation between "truth" documents, e.g., the handbook, and advice documents, e.g., wiki-like guides. I'm not tied to the exact approach I've laid out, but hopefully this highlights what I think would be a useful split and result in the creation of void-linux/void-guides

Vaelatern:
Does not seem distasteful to me, I'm OK with this general idea. Thank you for this work! I'll be waiting for others to consider the details before I want to merge it (with a spell-checked commit message)

bobertlo:
Sorry, I have just not been around to review changes to the docs until very recently. I am not against a separate guides section, but I believe that part belongs in a separate PR, if not an issue. I would be happy to merge everything else besides the section describing a non-existing repository.

bobertlo:
I tried to rebase this, and remove the section about external guides, but I messed up git. I will try again later lol

ericonr:
@markmcb don't you think some of this could fit better under the Contributing section of the documentation? We could definitely link from the README to it, but currently there is already a little bit of info there, and I feel like having two places to look for docs can get confusing.

markmcb:
@ericonr, maybe. 

I submitted this based on an IRC conversations and my first impressions when joining the Void community and upon hearing the wiki was deprecated. It felt like (and still feels like) there is a missing documentation strategy. That strategy should outline what documentation goes where, including this one. This attempts to start defining types of documentation. If you buy into the text in my commit, then I don't think this would go in the handbook as it's too detailed and it's not likely that every void user wants to contribute to documentation. 

I think the handbook should remain terse and link to details like this. Some for the handbook something short like this would be my suggestion, "The Void Linux project provides different sets of documentation. You are reading the handbook, which is intended to provide information every Void user needs. We also provide Guides for detailed use case walk-throughs, and the manual pages maintained by upstream projects. If you're interested in contributing to any of these, or want to know more about the differences between them, visit our documentation project in GitHub."

ericonr:
@markmcb I was thinking about this section and the pages inside it https://docs.voidlinux.org/contributing/void-docs/index.html . It is only about contributing to the documentation, so anyone who wishes to contribute should take a peek at them.

I can see what you mean

README.md
```diff
@@ -1,11 +1,202 @@
...
+And also not:
+
+```
+vi /etc/fstab
+```
+
+Command code-blocks should not be used to describe routine tasks documented
+elsewhere in this Handbook. For example, when writing documentation for the
+`foo` package, do not provide a command code-block stating that one should
+install it via `xbps-install foo`.
```
ericonr, line 100: Most of this part is already in the `contributing/void-docs` section, iirc. So we can try to port what's missing to that instead.

README.md
```diff
@@ -1,11 +1,202 @@
...
+
+The source for the [Void Handbook](https://docs.voidlinux.org/) is [void-linux/void-docs](https://github.com/void-linux/void-docs/) on GitHub.
+
+### Guides
+
+The goal of Guides is to cover specific scenarios that may not be of interest to all users. Examples of Guide documents might include, "full disk encryption," "dual-booting with another OS," etc. 
+
+While Guides are curated, they are the least formal documentation in terms of both tone and directiveness. They offer advice more than truth, i.e., the reader must exercise judgment to decide if it is right for a specific instance of Void Linux. For example, "Updating Packages with XBPS" is more of a truth in Void and belongs in the Handbook, whereas "Booting from Mirrored Drives" is a judgment call and belongs in a Guide as not all users will want to do this.
+
+The source for [Void Guides](https://guides.voidlinux.org/) is [void-linux/void-guides](https://github.com/void-linux/void-guides/) on GitHub.
```
ericonr, line 29: This part doesn't exist currently, and we are trying to fit the necessary guides inside of `advanced guides`.

README.md
```diff
@@ -1,11 +1,202 @@
...
+Pull requests should only contain a single commit. If a change is made after the
+initial commit `git add` the changed files and then run `git commit --amend`.
+The updated commit will need to be force pushed: `git push --force`.
+
+If multiple commits are made they will need to be squashed into one with `git
+rebase -i HEAD~X` where X is the number of commits that need to be squashed. An
+editor will appear to choose which commits to squash. A second editor will
+appear to choose the commit message. See
+[git-rebase(1)](https://man.voidlinux.org/git-rebase.1) for more information.
+The updated commit will need to be force pushed: `git push --force`.
```
ericonr, line 202: I think whatever part of this is missing from `contributing/void-docs` should go there. It might even be worth it to create a new page there too.

ericonr:
Hi there! Are you still interested in making these changes?

markmcb:
Hi @ericonr, I think I'm out. If there's ever a new interest in a wiki, or any system where I can solely write content, then I'm interested in contributing. The git workflow kind of kills it for me. I submitted the original content 5 months ago. It was a month before someone looked at it. It's too much overhead for someone like me who would prefer to work in bursts.

ericonr:
I see, I'm sorry you feel that way. Would you mind if I took some of your contributions and used them in other files? I will of course credit you in commits.

markmcb:
Steal with pride my friend. :) Use it however you see fit.

ericonr:
Thanks!

PR #151, opened 2020-03-05
Docker: Moved page from wiki and added to it.

Frick-David:
Moved Docker info from wiki to void-docs. Added a little about void-linux docker images.


Frick-David:
Not sure what failed. Tell me what I need to fix and I will

flexibeast:
Following the "Details" link next to "The Travis CI build failed", it looks like you might not have run the file through `vmdfmt`, as per [the Handbook style guide](https://docs.voidlinux.org/contributing/void-docs/style-guide.html).

Frick-David:
@flexibeast Thanks, I i will fix it. I was using mdbook to view it and assume mdbook build ran it thourh vmdfmt as well.

Frick-David:
My bad on needing to force it twice. Forgot to git add the changes. Should look a lot better now too. Let me know what you think.

flexibeast:
Firstly, thanks for taking the time to work on porting content from the wiki!

i'm not a member of the Void team, myself, so i'm not in a position to make an 'official' review, but for what it's worth, here are some initial thoughts:

1. There are several typos in the PR: 'deplay' rather than 'deploy', 'reosurces' rather than 'resources', 'folllowing' rather than 'following', 'adiition' rather than 'addition', 'diplayed' rather than 'displayed', 'docker' rather than 'Docker' (when not referring to the `docker` command or group), 'void' rather than 'Void', 'linux' rather than 'Linux'.

2. i'm not sure the text:

> It has become increasingly popular because it can be used to
deplay large number of containers quickly, easily, and scales well. It also uses
less reosurces than a typical VM.

should be included; it sounds more like promotional material than technical documentation.

3. As per the style guide, references to standard calls to `xbps-install`, as on line 17, are not needed.

4. The text:

> To install docker, and optionally, docker-compose run the folllowing command.
Docker Compose is a tool to create and run multi-container Docker apps. After
the install, be sure to enable and start the service before trying to use it. If
it is not enabled, it will not work. In adiition, you will need to add your user
to the docker group.
>
>
> `# xbps-install -S docker docker-compose`
> `# ln -s /etc/sv/docker /var/service`
> `# usermod -aG docker $USER`

could be shortened to:

> After installing the `docker` package (and optionally, the `docker-compose` package), enable the `docker` service:
>
> `# ln -s /etc/sv/docker /var/service`
>
> and add your user to the `docker` group:
>
> `# usermod -aG docker $USER`

5. i'm not sure it's necessary for the documentation to include checking whether the `docker` command is available post-installation, but if so, this could all be shortened down to `$ docker -v`, rather than including `docker` help output.

Obviously the typos can be corrected immediately, but the other suggestions i made should wait on any comment(s) by a member of the Void team.

Frick-David:
I will go ahead and make the changes you pointed out as I agree with all accounts. Thanks! I'll check back in after the changes!

Frick-David:
Think it looks a lot better per your suggestions. Thanks

flexibeast:
* Line 4 has 'visualisation' instead of 'virtualisation'.
* Line 8: the first instance of `docker` should be in backticks (\`) rather than straight quotes.
* More generally, there are still several instances of uncapitalised 'Docker': lines 21, 23, 75, 78, 85.
* Line 76: "docker-compose file" here should probably be "Docker Compose file".

Frick-David:
oh man I had autocorrect spell. Let me fix these. Thanks.

Frick-David:
Okay, I fixed everything you pointed out. Sorry for the trouble :(

Quick question, is it virtualization or virtualisation? Void Linux Is Europe-based and I am pretty sure an s is used whereas in the United States where I am, we usually use a z.  So should I put virtualisation? I know my computer marks it as misspelled with an s, and I am not going to lie, it does look a little odd to me, but I have seen civilisation in Europe and I know we spell it civilization. So I do want to double check that is how it is spelled in Europe. :)

flexibeast:
> Quick question, is it virtualization or virtualisation?

Well, i'm in Australia, and as a member of the Commonwealth, we tend to prefer British spellings to US spellings, so 'virtualisation' rather than 'virtualization'. That said, i don't know if there's a policy on this; a search for 'zation' in the Handbook returns no results, and a search for 'sation' in the Handbook shows 'customisation' used in `config/services/configuring.md`, but that was added by me. :-)

Frick-David:
Well, guess I will just go with the Commonwealth spelling for consistency. Somehow I always run into the some odd edge cases in my work ;/  Let me push it now with that change.

ericonr:
@Frick-David do you have any podman knowledge (mentioning it here because it's related to docker)? I couldn't get it working last time I tried (on musl, though).

Frick-David:
@ericonr I do not, but maybe I can help. Are you getting any sort of error message?

Frick-David:
@ericonr So I ran `xbps-install -S podman`
When I run `podman` I get a systemd Cgroup error, which happens for quite a bit of programs that were designed for GNU/Linux.
I then ran the following commands:
```
# mkdir /sys/fs/cgroup/systemd
# mount -t cgroup -o none,name=systemd cgroup /sys/fs/cgroup/systemd
```
which then solved that issue and podman appears to work fine.

ericonr:
@Frick-David I did try setting up the `/sys/fs/cgroup/systemd` folder, and that worked (btw, what a weird folder it is).

And obviously now that I tried getting tech support it worked perfectly. I was getting some weird access issues while trying to run without root. I'm not sure what caused them, but they were a while ago, so some update might have managed to fix my issue in some indirect manner. Or perhaps enought reboots after enabling the cmanager and ConsoleKit2 services was enough.

Thanks a lot for the help, anyway!

Frick-David:
That is good to hear! Yeah, generally when I have an issue it tends to be either a permissions issue as well. Python 3 also seems to be really picky about that!.
Any chance I can convince you to add podman documentation to the void-docs :) ?
Regardless, best of luck. Podman seems pretty cool! I might start using it myself.

src/config/docker.md
```diff
@@ -0,0 +1,96 @@
...
+```
+
+You then can issue other docker commands or use the image in a Dockerfile. Note,
+the Void Linux images also can be used in a docker-compose file. If you need
+clarification, please refer to the documentation. In addition, there are plenty
+of guides on working with docker on-line.
+
+## Troubleshooting
+
+### cgroups bug
```
bobertlo, line 82: I'm not an expert, but this directory already exists on my system, leading me to believe this is outdated?

src/config/docker.md
```diff
@@ -0,0 +1,96 @@
...
+# ln -s /etc/sv/docker /var/service
+```
+
+and add your user to the `docker` group:
+
+```
+# usermod -aG docker $USER
+```
+
+You will know you have properly installed docker if you can run it in the
```
bobertlo, line 21: From here down to the `voidlinux` docker images could all be considered 'how to use docker' and I don't think that fits the criteria for inclusion in the docs. I think the 'for more information' link could stay.

src/config/docker.md
```diff
@@ -0,0 +1,96 @@
+# Docker
+
+Docker is a platform of products to build isolated containers for programs using
+visualization.
+
+## Getting Started with Docker.
+
+After installing the 'docker' package (and optionally, the `docker-compose`
```
bobertlo, line 8: This is a nitpick, but I'd really prefer if this said just 'install the `docker` package and enable the docker service.

I'm also not sure we need to advertise the `docker-compose` package. It's package name is very intuitive.

ericonr:
>  Any chance I can convince you to add podman documentation to the void-docs :) ?

@Frick-David what do you think about making this a single page about containerization? Then Docker and Podman could be subheaders inside it. From my brief experience with podman, all it required was mounting the cgroups fs, so that step could be listed as relevant for both of them. When this gets merged I can add the additional Podman info.

bobertlo:
@ericonr can you confirm that the cgroups is still needed then? I installed the docker package and started the service and all was well. Can't dig into it right now myself.

src/config/docker.md
ericonr: my system doesn't have the `/sys/fs/cgroup/systemd` directory, but podman is working cleanly right now. @Frick-David do you have elogind installed? It's possible that it makes mounting this directory unecessary.

ericonr:
Yeah, neither docker nor podman is asking for any cgroups stuff any more here.

src/config/docker.md
Frick-David: I did have elogind I believe.

Frick-David:
So I will make the changes requested. 

Was the package changed such that the cgroups section should be removed?
Seems Buildah, Podman, and Skopeo are tools often used with Docker on the side, might be worth adding after all changes are approved.

ericonr:
@Frick-David do you have elogind enabled? If you don't, I believe your system will still require the cgroups section. Try to install and enable it to verify if the cgroups section is necessary then.

@bobertlo if it works cleanly with elogind do you think it's valid to only list that or should we list both options?

bobertlo:
Oh I do have elogind running that could be it. New to running it so I didn't consider that.

src/config/docker.md
```diff
@@ -0,0 +1,96 @@
...
+  trust       Manage trust on Docker images
+  volume      Manage volumes
+
+Commands:
+  ...
+  version     Show the Docker version information
+  wait        Block until one or more containers stop, then print their exit codes
+
+Run 'docker COMMAND --help' for more information on a command.
+```
```
ericonr, line 49: I'd prefer we not include command output, makes the page more crowded for little benefit.

Duncaen:
I don't really use docker, but I know that there should be a note about the `docker` group that its equivalent to becoming root.
See [the warning in archs wiki](https://wiki.archlinux.org/index.php/docker#Installation).

src/config/docker.md
```diff
@@ -0,0 +1,96 @@
...
+Void Linux with musl and an image for Void Linux with glibc. All images can be
+found on [Dockerhub](https://hub.docker.com/u/voidlinux).
+
+To download the musl-based image run:
+
+```
+docker pull voidlinux/voidlinux-musl
+```
+
+To download the glib-based image run:
```
Duncaen, line 69: Its glibc.

src/config/docker.md
```diff
@@ -0,0 +1,96 @@
+# Docker
+
+Docker is a platform of products to build isolated containers for programs using
+visualization.
```
PaperMountainStudio, line 4: visualization? Either I misunderstood this sentence, which means it should be probably worded differently, or `visualization` should be replaced with containarization  or virtualization?

Frick-David:
So I reread the documentation and one of the beginning pages says this wiki is for configuring the system only. So docker, nor my other commit, steam are necessary. However, I have seen people talk about a miscellaneous section. Can I have some direction on this as I do feel adding these pages to the 'core' documentation seems iffy. It is a different format than the wiki.

Frick-David:
@ericonr It should be good on the cgroups like you said.

src/config/docker.md
Frick-David: It should be virtualization

src/config/docker.md
Frick-David: Indeed

flexibeast:
@ericonr: Reviewing this PR again, i'm not sure of the extent to which it needs to be a separate page. From what i can tell, the critical parts are:

* You need to be a member of the `docker` group to use it; the `docker` group is equivalent to becoming root.
* If you're not using `elogind`, you need to do cgroups stuff.

In [this comment](https://github.com/void-linux/void-docs/pull/151#issuecomment-612665649), you suggested that we instead have a "Containerization" section, in which this information could be included. If you still think that's the way to go, i'm fine with that.

Frick-David:
@flexibeast, what are your thoughts on this now? Should we put docker into the external applications page?

flexibeast:
@Frick-David: i'm thinking that maybe we could have a "Containers and VMs" section, which would mention not only Docker, but also e.g. Podman, QEMU and libvirt. i'm going to check with more senior members of the Void team if they have any objections to such a section being included.

flexibeast:
@Frick-David: After asking on IRC, there don't seem to be any immediate objections to such a section, so i've just opened #350, so that Void team members not currently around on IRC can comment if necessary.

Frick-David:
That would be pretty sweet as that is a growing area in tech. That's one of the reasons I originally included using the void linux distro in the docker section as it seems a lot of people rely on alpine for small docker images but I think void is better suited. If a docker user does not need a glibc linked package they only need to tack musl onto the void container name and they have a much reduced container size. And if they do not glibc they can just remove the musl part. Compare this to having to use Debian or Ubuntu for glibc stuff and alpine for musl stuff. Really seems to me that alpine has a lot of unexplored potential in the container space.

-------- Original Message --------
On Jun 22, 2020, 11:13 PM, Alexis < notifications@github.com> wrote:
@Frick-David: i'm thinking that maybe we could have a "Containers and VMs" section, which would mention not only Docker, but also e.g. Podman, QEMU and libvirt. i'm going to check with more senior members of the Void team if they have any objections to such a section being included.
—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub, or unsubscribe.￼
[ { "@context": "http://schema.org", "@type": "EmailMessage", "potentialAction": { "@type": "ViewAction", "target": "[https://github.com](<a href=)/void-linux/void-docs/pull/151#issuecomment-647881561">https://github.com/void-linux/void-docs/pull/151#issuecomment-647881561", "url": "[https://github.com](<a href=)/void-linux/void-docs/pull/151#issuecomment-647881561">https://github.com/void-linux/void-docs/pull/151#issuecomment-647881561", "name": "View Pull Request" }, "description": "View this Pull Request on GitHub", "publisher": { "@type": "Organization", "name": "GitHub", "url": "https://github.com" } } ]

Frick-David:
Just saw the update on the new issue. Should I keep this one open then and wait to hear back?

-------- Original Message --------

flexibeast:
@Frick-David: Yes, please keep this open for now.

Frick-David:
Will do. Let me know!

flexibeast:
@Frick-David: There have been no comments on #350 in the last few days, so please free to convert this PR to one addressing that issue.

Frick-David:
@flexibeast for the containers and VM section, should it be its own overall section or a subsection under configuration? For now, I have it under configuration. Will move if told!

Frick-David:
Under this containters and vms section, I added my docker doc, and a podman doc which I will create as well. I edited the docker doc as discussed above (I think it looks a lot better so thanks for the feedback). I reviewed our discussion above and have a couple questions:

1) We agreed to delete the cgroups bug section entirely correct? I should just sad a portion at the top to make sure elogin is installed and enabled correct? Should the cgroups section be kept but just say to enable and install elogin?

2) Did we say to get rid of the void linux docker images section in the docker doc (move it to its own file maybe) ?

3) Did we say to get rid of the adding the user to the docker group entirely? I think a user still has to do that, correct? but that there should be a note that the adding a user to the docker group will give them sudo privaleges and the user should be wary of this?

flexibeast:
> For now, I have it under configuration.

Yeah, i think that's the place for it. But i'm open to it going elsewhere.

> We agreed to delete the cgroups bug section entirely correct? I should just sad a portion at the top to make sure elogin is installed and enabled correct? Should the cgroups section be kept but just say to enable and install elogin?

If Docker can work without installing the `elogind` package, then we should describe how to do so, i.e. say something along the lines "you can either install `elogind`, or do the following".

> Did we say to get rid of the void linux docker images section in the docker doc (move it to its own file maybe) ?

i think mention of Void Docker images should stay, unless someone else can suggest a better place?

> Did we say to get rid of the adding the user to the docker group entirely? I think a user still has to do that, correct? but that there should be a note that the adding a user to the docker group will give them sudo privaleges and the user should be wary of this?

i don't know if a user needs to be in the `docker` group in order to be able to use Docker; please check that. If they don't, i don't think we should suggest making use of the `docker` group, given that they'll get `root` privileges as a result. If they do, then there should be a warning about this, i.e.:

    **Warning**: Becoming a member of the `docker` group is equivalent to having `root` privileges.

Frick-David:
@flexibeast regarding the docker group: I saw this link on doing some investigating which seems to check out.
https://techoverflow.net/2017/03/01/solving-docker-permission-denied-while-trying-to-connect-to-the-docker-daemon-socket/
I get the error listed there if I run a docker command without sudo directly. But adding the user to the docker group eseentially gives them sudo anyways so that is an odd thing. So it comes down to, do i recommend to run every docker command with sudo or add the user to the docker group? Thoughts? I also will add that above error into the troubleshooting section at the bottom!

Frick-David:
I will say, barring the docker group issue. I think the doc is done and looks pretty nice. Formatting it now.

flexibeast:
i think i'd prefer to get @Duncaen's thoughts on what to do re. the privileges issue.

(Just as an aside, a point of terminology: it's important to distinguish between 'superuser' privileges, and `sudo`, a tool for controlling who has privileges to do what. A `sudo` configuration doesn't have to involve granting superuser privileges.)

Frick-David:
Okay gotcha. I will follow his advice. I will keep the docker group part but add the warning as you stated.

Frick-David:
uhhhhh. I was running the commands to amend the commit and my laptop unexpectedly shut down (hoping it is not about it fail, it is pretty old). and now I am getting some error messages from git I have never seen before. Really any command returns the following.
```
[david@Void-002 void-docs]$ git status
error: object file .git/objects/cc/fcce65a2005f850ae06c042a15df5a4840304f is empty
error: object file .git/objects/cc/fcce65a2005f850ae06c042a15df5a4840304f is empty
error: object file .git/objects/cc/fcce65a2005f850ae06c042a15df5a4840304f is empty
fatal: loose object ccfcce65a2005f850ae06c042a15df5a4840304f (stored in .git/objects/cc/fcce65a2005f850ae06c042a15df5a4840304f) is corrupt
```
Oh boy, this is gonna be fun to figure out.

ericonr:
I would first copy the repo somewhere else before trying to run commands. Might be able to do something with checkout or trying to continue a previous operation. `git reflog`  might contain a previous state of the repo as well.

Frick-David:
I will probably get a new laptop heat pipe and fan and do exactly what you said. I had the finished doc on that laptop (thought it read pretty well too) so I will try recover it from there. I will try and get this done ASAP with that considering. The laptop seems to flicker on and off. What bad timing :( In the meantime, I am starting the podman doc on another computer.

Frick-David:
Okay, hey again. So loooking over it. I think that podman section is probably short enough to include in the same page as docker as we discussed. That being said, I see that the containers and VMS has a couple other sections (say libvirt). Is the plan that all of them should be one page or multiple ones?

flexibeast:
@Frick-David: Having a "Docker and Podman" page might be reasonable, given that https://podman.io/whatis.html says:

> What is Podman? Simply put: `alias docker=podman`

i was originally thinking that we might be able to fit all the "Containers and VMs" stuff in a single section, but putting together LXC info in #448, it's become apparent that's not going to be practical. So in general, apart from this "Docker and Podman" section, i think we should start by having individual sections (which will also facilitate different people working on the different sections at the same time, without treading on each other's toes). We can easily consolidate things later.

Finally, a note about the "cgroups bug" section in this PR. The `lxd` [service](https://github.com/void-linux/void-packages/blob/master/srcpkgs/lxd/files/lxd/run) does the directory creation and mounting itself if necessary (and i've followed the same model when creating [the `lxd` 66 service](https://github.com/mobinmob/void-66-services/blob/devel/usr/share/66/service/lxd). So this looks to me like an issue that needs to be fixed in the `docker` package. @ericonr, do you agree? If so, i'll add the 'blocked' and 'packaging' labels to this PR.

ericonr:
Huh, that service feels kida racey... What happens if someone has the `lxd` and the `docker` daemons running at the same time? Something like that should be fixed in `void-runit` (or with a separate core-service), if fixes are in fact needed.

flexibeast:
@ericonr: Good point. i just installed `docker` in my LXC VM and enabled the service, so that both `lxd` and `docker` were enabled. Both seemed to work okay, including after reboot, but my testing was very light.

@ahesford: Thoughts?

ahesford:
Regarding the cgroups mount in the `lxd` run service: this is the wrong way to handle things. If the `/sys/fs/cgroup/systemd` is universal and important enough to be mounted, it should either be mounted in `/etc/runit/core-services/00-pseudofs.sh` like the rest of the `cgroup` mounts (and, since the last release, the `cgroup2` `unified` mount) or users should be instructed to mount that in `/etc/fstab` or `/etc/rc.local`.

I don't know if there's a compelling case to make the `systemd` mount a global `pseudofs` action like the other `cgroup` mounts; I put the `cgroup2` mount in there because it is ultimately supposed to be the "path forward" for cgroups and we're already mounting the original hierarchy.

Even if mounting this in the `lxd` service *were* the right action, my belief is that any service taking responsibility for mounting something in `run` ought to be sure to unmount it in `finish`.

flexibeast:
@ahesford: Okay, thanks. Should changing the `lxd` service be made part of https://github.com/void-linux/void-packages/pull/24185 ?

ahesford:
Maybe. I've taken up discussion in the LXD issue.

CameronNemo:
@ahesford I do not particularly care how the cgroup gets mounted as long as it works. My concern is that if both lxd and docker have the systemd mount in their run and finish scripts, then the unmount in one of the finish scripts could mess up the other running service. also the pseudofs core service does not unmount any filesystems -- why should the lxd run script? unmounting accomplishes very little because it is a pseudo filesystem.

src/config/docker.md
Piraty: namespaces have nothing to with virtualization.

Issue #148, opened 2020-02-15
Undocumented /etc/rc.conf


xtraeme:
The void handbook lacks information about `/etc/rc.conf`, `/etc/rc.local` and `/etc/rc.shutdown`.

I assume this is so because english/american keyboards don't need additional tweaks... but it's really important in all other cases.

Please document them properly!

flexibeast:
i'd like to work on this. Would you like an overview of the purpose of each of these files, mentioning that `rc.conf` in sourced in stage 1 and stage 3, `rc.local` in stage 2, and `rc.shutdown` in stage 3? 

> I assume this is so because english/american keyboards don't need additional tweaks... but it's really important in all other cases.

So the description of `rc.conf` should mention setting KEYMAP? Would you like other variables mentioned as well, e.g. FONT?

xtraeme:
Yes, all variables should be documented.

bobertlo:
Closed by #154

PR #154, opened 2020-03-07
Add section about rc files.

flexibeast:
Initial draft to address #148, for feedback.

In documenting variables that can be set in `rc.conf`, this duplicates information about `HARDWARECLOCK` currently contained in the "Date and Time" section. i'm not sure if DRY is a concern here.

The information about `KEYMAP` doesn't appear elsewhere in the Handbook - i imagine many users won't necessarily think to look in this section for information about how to specify the console keymap?

### DIFF
--- a/src/config/rc-files.md
+++ b/src/config/rc-files.md
@@ -1,0 +1,62 @@
+# rc.conf, rc.local and rc.shutdown
+
+The files `/etc/rc.conf`, `/etc/rc.local` and `/etc/rc.shutdown` can be used to
+configure certain parts of your Void system. `rc.conf` is often configured by
+`void-installer`.
+
+## rc.conf
+
+Sourced in runit stages 1 and 3. This file can be used to set variables,
+including the following:
+
+### KEYMAP
+
+Specifies which keymap to use for the Linux console. Available keymaps are
+listed in `/usr/share/kbd/keymaps`. For example:
+
+```
+KEYMAP=fr
+```
+
+For further details, refer to
+[loadkeys(1)](https://man.voidlinux.org/loadkeys.1).
+
+### TIMEZONE
+
+Specifies which timezone to use. Available timezones are listed in
+`/usr/share/zoneinfo`. For example:
+
+```
+TIMEZONE=Europe/Berlin
+```
+
+### HARDWARECLOCK
+
+Specifies whether the hardware clock is set to UTC or local time.
+
+By default this is set to `utc`. However, Windows sets the hardware clock to
+local time, so if you are dual-booting with Windows, you need to either
+configure Windows to use UTC, or set this variable to `localtime`.
+
+For further details, refer to [hwclock(8)](https://man.voidlinux.org/hwclock.8).
+
+### FONT
+
+Specifies which font to use for the Linux console. Available fonts are listed in
+`/usr/share/kbd/consolefonts`. For example:
+
+```
+FONT=eurlatgr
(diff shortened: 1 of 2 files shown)


src/config/rc-files.md
```diff
@@ -0,0 +1,61 @@
+# rc.conf, rc.local and rc.shutdown
+
+The files `/etc/rc.conf`, `/etc/rc.local` and `/etc/rc.shutdown` can be used to
+configure certain parts of your Void system.
```
Vaelatern: Perhaps mention that `/etc/rc.conf` is often configured by void-installer ?

src/config/rc-files.md
```diff
@@ -0,0 +1,61 @@
...
+Specifies which font to use for the Linux console. Available fonts are listed in
+`/usr/share/kbd/consolefonts`. For example:
+
+```
+export FONT=eurlatgr
+```
+
+For further details, refer to [setfont(1)](https://man.voidlinux.org/setfont.1).
+
+## rc.local
```
Vaelatern: Mention that these are assumed to be shell scripts, while rc.conf is normally just the variables to set?

flexibeast:
@Vaelatern: Thanks, updated.

bobertlo:
I would have liked `rc.local` / `rc.shutdown` info to be in the `runit` section, but that got merged back into services/demons.

I wonder if it would make sense to separate the 'configuration' file from script files?

Would there be more appropriate places to place the variables in the manual? I think just describing them here and pointing to this page from mentions of `rc.conf` elsewhere in the manual may be clearer. I'm tempted to say if they don't belong somewhere else we don't need them here. Everything *is* documented in the file itself. Perhaps we could just refer readers to there for more information?

I'm definitely not telling you what to do here I think this is very open to discussion.

flexibeast:
Fair enough. i'd be interested in @xtraeme's thoughts on this.

src/config/rc-files.md
```diff
@@ -0,0 +1,61 @@
...
+
+For further details, refer to [hwclock(8)](https://man.voidlinux.org/hwclock.8).
+
+### FONT
+
+Specifies which font to use for the Linux console. Available fonts are listed in
+`/usr/share/kbd/consolefonts`. For example:
+
+```
+export FONT=eurlatgr
```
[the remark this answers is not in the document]
xtraeme: This is still unfixed. In `/etc/rc.conf` there's no need to use `export`

src/config/rc-files.md
flexibeast: Added in https://github.com/void-linux/void-docs/commit/d010d217ab8788d2feffb6711fa2825f53fbe4c3.

src/config/rc-files.md
flexibeast: Added in https://github.com/void-linux/void-docs/commit/d010d217ab8788d2feffb6711fa2825f53fbe4c3.

xtraeme:
Looks ok to me right now, thanks!

src/config/rc-files.md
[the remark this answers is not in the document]
flexibeast: Thanks, i've now removed `export` from the examples. Unless you meant that i should remove the examples themselves?

bobertlo:
As an alternative to my earlier suggestion, other pages that reference settings in this file could link *here* without duplicating content. I'm merging this. Thanks :)

Issue #121, opened 2019-10-17
Document default groups


Duncaen:
This came  up yesterday in the IRC channel, there are a number of [default groups](https://github.com/void-linux/void-packages/blob/master/srcpkgs/base-files/files/group) that void ships with that are currently not really documented.

I think a section in the docs that describes what each group does would be good to have.

flexibeast:
i'd like to work on this. Whereabouts in the Handbook should this new section be added?

Duncaen:
I think "Configuration > Users & Groups" and then maybe a "Default Groups" or just as section in the "Users & Groups" page.
I think it only makes sense to split this into its own page if the "Users & Groups" page would get too long.

I do think we were too eager to split sections like "Repositories" and some other categories into too many sub pages, which really doesn't help readability.

bobertlo:
Closed by #155

PR #155, opened 2020-03-07
Add 'Users and groups' section.

flexibeast:
Initial draft to address #121, for feedback.

This draft is based on a combination of information on the Debian wiki, the Arch wiki, and calls to `find / -group`. :-) The 'Description' field for `mail` has text in square brackets to indicate that i'm particularly unsure about it, but hopefully the other descriptions aren't too far off-base.

### DIFF
--- a/src/config/users-and-groups.md
+++ b/src/config/users-and-groups.md
@@ -1,0 +1,39 @@
+# Users and groups
+
+## Default groups
+
+Void Linux defines a number of groups by default.
+
+| Group      | Description                                                   |
+|------------|---------------------------------------------------------------|
+| `root`     | Complete access to the system.                                |
+| `bin`      | Unused - present for historical reasons.                      |
+| `sys`      | Unused - present for historical reasons.                      |
+| `kmem`     | Ability to read from `/dev/mem` and `/dev/port`.              |
+| `wheel`    | Elevated privileges for specific system administration tasks. |
+| `tty`      | Access to TTY-like devices:                                   |
+|            | `/dev/tty*`, `/dev/pts*`, `/dev/vcs*`.                        |
+| `tape`     | Access to tape devices.                                       |
+| `daemon`   | System daemons that need to write to files on disk.           |
+| `floppy`   | Access to floppy drives.                                      |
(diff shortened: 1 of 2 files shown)


src/config/users-and-groups.md
```diff
@@ -0,0 +1,39 @@
...
+| `audio`    | Access to audio devices.                                      |
+| `video`    | Access to video devices.                                      |
+| `utmp`     | Ability to write to `/var/run/utmp`, `/var/log/wtmp`          |
+|            | and `/var/log/btmp`.                                          |
+| `adm`      | Unused - present for historical reasons. This group was       |
+|            | traditionally used for system monitoring, such as viewing     |
+|            | files in `/var/log`.                                          |
+| `cdrom`    | Access to CD devices.                                         |
+| `optical`  | Access to DVD/CD-RW devices.                                  |
+| `mail`     | [Mailboxes in /var/mail?]                                     |
```
bobertlo: is this a question?

bobertlo:
Aside from that question mark this looks good to me.

src/config/users-and-groups.md
flexibeast: Sorry, yes. The information i found suggested this was what the `mail` group is for, but i wasn't sure whether this is true on Void, or whether it's actually unused and is only present for historical reasons.

bobertlo: I'm really not sure on this one. It's definitely historical, but could be still used by some mail packages? I have no idea.

flexibeast: @bobertlo: Fair enough. :-)
@Duncaen: Are you aware of the `mail` group being used by any Void package(s)?

Duncaen: Looks like dma is the only package that uses the mail account, all others setup their own dedicated users.

flexibeast: @Duncaen: Thanks, PR updated accordingly.

bobertlo:
d8df59ebd8de0789ad6b7e99cbef22ff119c1b6e

PR #156, opened 2020-03-07
contributing: add information about infradocs

iio7:


src/contributing/void-docs/infradocs.md
```diff
@@ -0,0 +1,9 @@
+# InfraDocs
+
+[InfraDocs](https://infradocs.voidlinux.org/) is the manual for Void systems
```
the-maldridge: This sounds like infradocs is a manual for managing void systems in general.  This is in fact not true, is a meta-manual for managing void itself.

src/contributing/void-docs/infradocs.md
```diff
@@ -0,0 +1,9 @@
...
+management.
+
+The manual addresses the technical infrastructure managed by different systems
```
the-maldridge: Once the paragraph above has been corrected this paragraph is redundant.

src/contributing/void-docs/infradocs.md
```diff
@@ -0,0 +1,9 @@
...
+and how the project works on a human level.
+
+Knowledge of Void or base Linux administration is assumed.
```
the-maldridge: This is assumed across most if not all of Void, and probably doesn't need to be explicitly stated.

src/SUMMARY.md
```diff
@@ -75,3 +75,4 @@
    - [Void Docs](./contributing/void-docs/index.md)
       - [Style Guide](./contributing/void-docs/style-guide.md)
       - [Submitting Changes](./contributing/void-docs/submitting.md)
+   - [InfraDocs](./contributing/void-docs/infradocs.md)
```
the-maldridge: I'm not convinced that /contributing is the best place for this.  It might be better to link from /about/meta, as the docs on infradocs aren't really about contributing.

iio7:
Please let me know if I did this wrong.

the-maldridge:
The edits look ok, but you need to squash your PR down to one commit.  `git rebase` can do this, which you can find documentation about in the git book.   Preferably you would use the "fixup" mode during squash.

iio7:
Have never used rebase before, hope I got it right.

I would imagine many users not used to Git could perhaps give up at this stage. Is there an "easier" way or perhaps extend the "Submit changes" part of the manual with a rebase example? I'm thinking that perhaps a user isn't well versed in Git, or at least in rebasing, but is still providing valuable contributions.

the-maldridge:
Generally we try not to include documentation in void's manuals that is easy to find elsewhere.  For git, the general reference is [the git book](https://git-scm.com/book/en/v2).

iio7:
I thought as much, but I think it would help a lot with an example like the one already in the manual. 

The Git book is really good, but even though the rebase chapter is well explained, I had to clone the repo and do some trial and error in order to get it right because I couldn't "get how to apply the information to the situation in my case", and I almost gave up in advance because I didn't feel like reading the huge chapter on rebasing.

This is not meant as a complaint, but just as a suggestion, it would make it easier to contribute since Void is depending on the rebase function of Git.

the-maldridge:
Oh you made your initial edit with the web UI.  I'm not sure if there is
any way to do a rebase from there.  Its not meant for large edits and in
my experience isn't really that usable.

iio7:
No, I didn't use the web UI (as far as I recall), I did all the work in the terminal.

PR #159, opened 2020-03-14
Steam: New PR with minor changes. Ported from wiki.

Frick-David:
Sorry for closing other PR and making a new one. My git skills are not top notch in terms of debugging after a mistake. I figured it would be better to do it this way.
I will continue to port documets over from the wiki, with best quality!


bobertlo:
Please see #188. 

We have re-discussed this (it comes up time and again) and are still not interested in pages like this in general for installing third party software.

I think the recommended way to install Steam is through flatpak at this point. The package still exists but we cannot try to support every possible need of arbitrary third party software, installed through a third party service, which is itself closed source.

In light of these things I don't see a reason to include this page, since it would only be a stub pointing to a flatpak section.

bobertlo:
PS: Sorry for the delayed response :)

ericonr:
@Frick-David I think the current external applications page is the most we'd like to have regarding this sort of stuff. Some of the solutions seem useful and I think if you use them or want others to find them, you should host a document for it, but as it is, they don't really fit in the current structure of void-docs.

flexibeast:
What about having a small "Steam" entry on the "External Applications" page? There seem to be regular questions on IRC about getting it working satisfactorily, and i don't think it would necessarily be any more out of place than the existing"Octave" and "MATLAB" entries.

Frick-David:
@flexibeast So I agree, the best section is external applications. But that could quickly become a monster of a document for A LOT of external programs. Does the void team plan on making that a section with sub pages by chance? IMO (up the the team though of course), making that a section with sub documents for each program but making it collapsible so it does not eat up the left menu on eh docs page is the way to go. I will probably put both docker and steam in there if I can get a green light from someone?

flexibeast:
@Frick-David: We've discussed this on IRC, and we'd like Steam to be mentioned in the "External Applications" page, but it should simply tell people to use the Flatpak. Could you please convert this PR accordingly?

Your comment about the potential for the "External Applications" page to become monstrous is a good one. My personal feeling is that, as much as possible, we should indeed try to avoid documenting individual applications in the Handbook, but that there might need to be occasional exceptions. In my experience, people ask about Steam on IRC regularly enough that i think it might be helpful to provide the canonical answer of "Use the Flatpak" in the Handbook. i'm not sure about Docker; we can discuss it further in #151.

ericonr:
Weighing in on this one, I think we can mention Steam by name in the flatpak section. Trying to get a `steam-deps` metapackage into void-packages would potentially be much more helpful. Then the INSTALL message for `steam` could just be `install steam-deps-32bit`.

flexibeast:
@ericonr: That sounds like a neat approach! Have you discussed this idea with other Void team members?

ericonr:
I've mentioned it on IRC once, but never actually discussed it. The biggest issue with it is actually because of the graphics provider. Until we have a proper virtual package for OpenGL, people will have to choose`mesa-dri-32bit` or `nvidia-libs-32bit` manually, and I believe both packages pull in practically everything Steam needs, making a metapackage useless. Have to actually check it, though.

Frick-David:
Would you still like me to put steam in the external applications page, or hold off if y'all discuss the `steap-deps` ?

ericonr:
`steam-deps` is a bit separate from that. I prefer Flatpak as the definitive solution.

Frick-David:
@ericonr So go ahead and edit the commit to make a text edit to the external applications page saying to get steam as a flatpak, correct?

ericonr:
@Frick-David yes, exactly! Thanks c:

Frick-David:
@ericonr can you help me? I am having a lot of trouble with stuff this week. Just fixed my laptop to have my mobo partially break. I am super tired and making a fool of myself in git.

ericonr:
@Frick-David wow, that's some bad luck :/

What can I help you with?

Frick-David:
@ericonr  yeah :( Okay, walk me through these steps. I had the steam PR wayyyy behind master (because it has taken a quite a bit of time). External applications was added to the repo more recently which I was to ad steam as an external application to. What commands would I run to get the PR the newest files to then build off of. Is that a rebase?

Need to really build my git skills again ;/

Secondly, not entirely sure what I ended up doing with git here? Is there a way you can guide me through how to fix this PR with git? I would be super grateful. I want help out with the void project as much as possible, but I am finding my git skills are holding me back ;/

ericonr:
@Frick-David 

> What commands would I run to get the PR the newest files to then build off of. Is that a rebase?

Yes. You can try to do it with `git rebase upstream/master` (assuming you have set up this repo as upstream remote already), but it might find a few conflicts.  Given that you were mostly messing with a separate file, the conflicts shouldn't be too big and solvable in the command line - basically: conflict appears, open file, find `========` line, leave it in a good state, then run `git rebase --continue`. After the rebase is done, you can `git push --force origin master` to update your branch in GH, which is probably necessary for the "Reopen PR" button to start working.

(if you haven't set up the upstream remote, you should do `git remote add upstream https://github.com/void-linux/void-docs.git`, then `git fetch upstream`, then the rebase commands)

Regarding fixing this PR, you probably did something weird with your master branch and pushed it. I believe it should be solved by the force push above.

If you want to erase your current commits, something like `git rebase -i HEAD~3` will give you a commit list, and you can drop commits from there. Do be careful, they will be deleted :p 

> I want help out with the void project as much as possible, but I am finding my git skills are holding me back ;/

Glad to have help here c: . I will see if we can improve contribution documentation more too. If my instructions were lacking, please let me know.

Frick-David:
@ericonr I see that there is a steam package now.Does that affect the doc?

Issue #130, opened 2019-11-05
Add 'Musl' section?


flexibeast:
Might it be useful to port [the Musl page on the wiki](https://wiki.voidlinux.org/Musl) (or at least some of it) across to the Handbook, and mention things like the `gcompat` package?

rubyFeedback:
This would be pretty cool. I have been searching for an oldschool veteran *nix distribution and musl in combination with busybox, and perhaps llvm for a small core system would be very interesting to me - so any documentation here is very useful. I did succeed in building/compiling a custom musl-toolchain and it works too, but past this point I am a bit lost, as in ... "what to do next". So I am missing some steps.

Thus, this would be very useful in general - any documentation improvement here.

bobertlo:
@rubyFeedback some of those topics seem outside the scope of this documentation. You may want to take a look at alpine linux for a busybox musl system. There are good sources on the musl wiki for examples of other toolchains as well. Also, the void bootstrap toolchain. Linux From Scratch is also a great guide for these off topic ideas. IIRC they now even have a musl based guide?

flexibeast:
i'm starting work on this.

bobertlo:
closed by #163

PR #163, opened 2020-03-14
Add "musl" section.

flexibeast:
Initial draft to address #130, for feedback.

i wasn't sure where to put this section in the Handbook, so i've just ended up putting it under the "Configuration" section. i've also added a link to this section from the "Base Requirements" page, with the idea of making people aware that a musl system might not be appropriate for their needs.

Also, would it be correct to say that those who need locale support should use glibc Void, rather than musl Void? (Locale support is certainly one of the reasons i use glibc rather than musl.)

### DIFF
--- a/src/config/musl.md
+++ b/src/config/musl.md
@@ -1,0 +1,52 @@
+# musl
+
+[musl](https://musl.libc.org/) is a libc implementation which strives to be
+lightweight, fast, simple, and correct.
+
+Void officially supports musl by using it in its codebase for all target
+platforms (although binary packages are not available for i686). Additionally,
+all compatible packages in our official repositories are available with
+musl-linked binaries in addition to their glibc counterparts.
+
+Currently, there are nonfree and debug subrepositories for musl, but no multilib
+subrepo.
+
+## Incompatible software
+
+Musl practices very strict and minimal standard compliance. Many commonly used
+platform-specific extensions are not present. Because of this, it is common for
+software to need modification to compile and/or function properly. Void
+developers work to patch such software and hopefully get portability/correctness
+changes accepted into the upstream projects.
+
+Proprietary software rarely supports non-glibc libc implementations, although
+sometimes these applications are available as [flatpaks](https://flatpak.org/),
+which provide their own libc in the image.
+
+### glibc chroot
+
+Software requiring glibc can be run in a glibc chroot.
+
+Create a directory that will contain the chroot, and install a base system in it
+via the `base-voidstrap` package. If network access is required, copy
+`/etc/resolv.conf` into the chroot; `/etc/hosts` may need to be copied as well.
+
+Several directories then need to be mounted as follows:
+
+```
+# mount -t proc none <chroot_dir>/proc
+# mount -t sysfs none <chroot_dir>/sys
+# mount --rbind /dev <chroot_dir>/dev
+# mount --rbind /run <chroot_dir>/run
+```
+
+Use [chroot(1)](https://man.voidlinux.org/chroot.1) to change to the new root,
+then run glibc programs as usual. Once you've finished using it, unmount the
+chroot using [umount(8)](https://man.voidlinux.org/umount.8).
+
+#### PRoot
+
+An alternative to the above is [proot(1)](https://man.voidlinux.org/proot.1), a
+user-space implementation of chroot, mount --bind, and binfmt_misc. By
+installing the `proot` package, unprivileged users can utilise a chroot
+environment.
--- a/src/installation/base-requirements.md
+++ b/src/installation/base-requirements.md
@@ -14,6 +14,9 @@ following minimums for most installations:
 
 Void is not available for i386, i486, or i586 architectures.
 
+Before installing musl Void, please read [the "musl" section](../config/musl.md)
+of this Handbook, so that you are aware of software incompatibilities.
+
 It is highly recommended to have a network connection available during install
 to download updates, but this is not required. ISO images contain installation
 data on-disc and can be installed without network connectivity.
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -12,6 +12,7 @@
       - [Partitioning Notes](./installation/live-images/partitions.md)
       - [Installation Guide](./installation/live-images/guide.md)
 - [Configuration](./config/index.md)
+   - [musl](./config/musl.md)
    - [Post Installation](./config/post-install.md)
    - [Services And Daemons](./config/services/index.md)
       - [Managing Services](./config/services/managing.md)


ericonr:
> Also, would it be correct to say that those who need locale support should use glibc Void, rather than musl Void? (Locale support is certainly one of the reasons i use glibc rather than musl.)

At least on my side, the existence of locales has only made life more difficult with glibc, without any benefits. What I expected to be solved by going to glibc didn't actually work.

src/config/musl.md
```diff
@@ -0,0 +1,99 @@
...
+solutions and old PCs, but it may not be worth the effort to use on a modern
+desktop computer.
+
+## Possible problems
+
+Some programs (mostly graphical applications) will work incorrectly, or
+segfault, when run under musl. This may be due to programs expecting some
+glibc-specific behavior. However, currently most breakages are caused by default
+stack size differences — if you are a software developer, you can try
+rebuilding musl with a bigger stack size and see if that helps.
```
ericonr: This is not necessary, you can use a shell to launch your program and run `ulimit -s hard` on said shell. This completely avoids the issues with limited stack size on musl. We could include a short script that launches something like qutebrowser (which crashes due to the stack issue) through bash.

A script like this:

```shell
#!/usr/bin/env bash

ulimit -s hard
"$@"
```

flexibeast: Ah okay, thanks for pointing this out - i've update the PR accordingly.

q66: @ericonr this is wrong, the musl stack size limitation is thread stack size, not main stack size, and that is *not* influenced by process limits. The main stack size tends to be 8M or so regardless of the libc you run.

src/config/musl.md
```diff
@@ -0,0 +1,99 @@
+# musl
+
+[musl](https://musl.libc.org/) is a libc implementation which strives to be
+lightweight, fast, simple, and correct.
+
+Void officially supports musl by using it in its codebase for all target
+platforms except i686. Additionally, all the packages in our official
+repositories are available with musl-linked binaries in addition to their glibc
+counterparts.
```
q66: i686-musl works the same as anything else, there are just no binary packages for it

src/config/musl.md
```diff
@@ -0,0 +1,99 @@
...
+solutions and old PCs, but it may not be worth the effort to use on a modern
+desktop computer.
+
+## Possible problems
+
+Some programs (mostly graphical applications) will work incorrectly, or
+segfault, when run under musl. This may be due to programs expecting some
+glibc-specific behavior. However, currently most breakages are caused by default
+stack size differences - try running the program from a shell that has called
+`ulimit -s hard`.
```
q66: i would just leave the entire paragraph out in favor of something more vague along the lines of "In some cases software compatibility may be worse with musl". Crashing is not expected behavior, and most of the stack size issues have been ironed out anyway.

src/config/musl.md
```diff
@@ -0,0 +1,98 @@
...
+
+Proprietary software rarely supports non-glibc libc implementations, although
+sometimes such implementations may be available via
+[flatpak](https://flatpak.org/).
+
+The `gcompat` package, a compatibility layer to allow running glibc binaries on
+musl systems, can sometimes assist in getting glibc-based software to work.
+However, if that doesn't work, try creating a glibc chroot.
```
q66: gcompat almost never works except for the simplest of cases (for real world things you can pretty much expect it to *never* work), I wouldn't bother mentioning it at all

src/config/musl.md
flexibeast: Ah okay - PR updated.

src/config/musl.md
```diff
@@ -0,0 +1,94 @@
...
+platforms except i686. Additionally, all the packages in our official
+repositories are available with musl-linked binaries in addition to their glibc
+counterparts.
+
+Currently, there are nonfree and debug subrepositories for musl, but no multilib
+subrepo.
+
+Generally speaking, a musl based system is a good choice for servers, embedded
+solutions and old PCs, but it may not be worth the effort to use on a modern
+desktop computer.
```
q66: this is a subjective opinion and shouldn't be present in the handbook either

i don't see what makes musl any more suited for servers than desktops

on old PCs the claim makes some sense because of lower memory footprint, but yeah, I wouldn't bother

src/config/musl.md
```diff
@@ -0,0 +1,94 @@
...
+Install a base system in it:
+
+```
+# xbps-install -S -R https://alpha.de.repo.voidlinux.org/current -r ~/void-glibc base-voidstrap
+```
+
+Mount the appropriate directories:
+
+```
+# mount --bind ~/void-glibc ~/void-glibc # some programs work incorrectly without this
```
q66: what is this even about?

src/config/musl.md
```diff
@@ -0,0 +1,94 @@
...
+# cp -L /etc/resolv.conf ~/void-glibc/etc/resolv.conf # for network
+# mount --rbind /proc ~/void-glibc/proc
+# mount --rbind /sys ~/void-glibc/sys
+# mount --rbind /dev ~/void-glibc/dev
+# mount --rbind /run ~/void-glibc/run
```
q66: `/proc`, `/sys` should not be bind-mounted

you can easily do:

```
mount -t proc none /path/to/proc
mount -t sysfs none /path/to/sys
```

src/config/musl.md
flexibeast: i don't know; this was just part of what i ported across from the wiki. Should i just delete this line?

src/config/musl.md
```diff
@@ -0,0 +1,94 @@
...
+
+```
+# xbps-install -S -R https://alpha.de.repo.voidlinux.org/current -r ~/void-glibc base-voidstrap
+```
+
+Mount the appropriate directories:
+
+```
+# mount --bind ~/void-glibc ~/void-glibc # some programs work incorrectly without this
+# cp -L /etc/resolv.conf ~/void-glibc/etc/resolv.conf # for network
```
q66: i think `/etc/hosts` is also necessary

src/config/musl.md
flexibeast: So in this specific example, it should be e.g. `mount -t proc none ~/void-glibc/proc`?

src/config/musl.md
q66: i can't think of any real cases where this would matter (I never had issues chrooting without creating a new mountpoint) but hard to say for sure

src/config/musl.md
q66: yeah

`/dev` must be bound though, because it's managed by udev

src/config/musl.md
flexibeast: Removed.

src/config/musl.md
flexibeast: Removed.

src/config/musl.md
flexibeast: Added.

src/config/musl.md
flexibeast: Updated.

src/config/musl.md
flexibeast: Thanks, updated.

src/config/musl.md
flexibeast: Okay, i've removed it - it can always be added back in later if it turns out to be necessary.

src/config/musl.md
flexibeast: @q66: So should i put the original text back in?

q66: no, you already removed the paragraph

flexibeast: Okay, thanks.

ericonr: @q66 sorry for the wrong information. Using ulimit seemed to have solved one issue I had, which was opening any reddit page on qutebrowser while logged into reddit, causing a seg-fault. I'm not sure what led me to believe it had fixed the issue, but testing it now showed me the issue is still present.Weird, but ok.

src/config/musl.md
```diff
@@ -0,0 +1,67 @@
...
+
+Void officially supports musl by using it in its codebase for all target
+platforms (although binary packages are not available for i686). Additionally,
+all compatible packages in our official repositories are available with
+musl-linked binaries in addition to their glibc counterparts.
+
+Currently, there are nonfree and debug subrepositories for musl, but no multilib
+subrepo.
+
+## Possible problems
```
bobertlo: ## Incompatible Software ??

src/config/musl.md
```diff
@@ -0,0 +1,67 @@
...
+
+In some cases, software compatibility may be worse with musl.
```
bobertlo: Something like this maybe?

Musl practices very strict and minimal standard compliance. Many commonly used platform specific extensions are not present. Because of this, it is common for software to need modification to compile and/or function properly. Void developers work to patch such software and hopefully get portability/correctness changes accepted into the upstream projects.

src/config/musl.md
```diff
@@ -0,0 +1,67 @@
...
+sometimes these applications are available as [flatpaks](https://flatpak.org/),
+which provide their own libc in the image.
+
+### glibc chroot
+
+Software requiring glibc can be run in a glibc chroot.
+
+Create a directory that will contain the chroot, and install a base system in it
+via the `base-voidstrap` package. If network access is required, copy
+`/etc/resolv.conf` and `/etc/hosts` into the chroot.
```
bobertlo: I'm not sure `/etc/hosts` should be recommended here. There are a lot of chroot applications where this would be unnecessary and/or undesireable.

src/config/musl.md
```diff
@@ -0,0 +1,67 @@
...
+# mount -t sysfs none <chroot_dir>/sys
+# mount --rbind /dev <chroot_dir>/dev
+# mount --rbind /run <chroot_dir>/run
+```
+
+Use [chroot(1)](https://man.voidlinux.org/chroot.1) to change to the new root,
+then run glibc programs as usual. Once you've finished using it, unmount the
+chroot using [umount(8)](https://man.voidlinux.org/umount.8).
+
+To seamlessly use graphical applications, install the `xhost` package and run:
```
bobertlo: I'm still not sure this is a complete solution? Wouldn't you at least need to export a DISPLAY variable?

src/config/musl.md
```diff
@@ -0,0 +1,67 @@
...
+before chrooting.
+
+#### PRoot
+
+An alternative to the above is [proot(1)](https://man.voidlinux.org/proot.1), a
+user-space implementation of chroot, mount --bind, and binfmt_misc. By
+installing the `proot` package, unprivileged users can utilise a chroot
+environment.
+
+For example:
```
bobertlo: I wonder if we would be better just referring readers to the manual. I don't think this quick rundown is quite enough information, and I also don't think we need more information about it in this page.

src/installation/base-requirements.md
```diff
@@ -14,6 +14,9 @@ following minimums for most installations:
 
 Void is not available for i386, i486, or i586 architectures.
 
+Before installing musl Void, please read [the "musl" section](../config/musl.md)
+of this Handbook, so that you are aware of possible issues.
```
bobertlo: aware of software incompatibilities.

bobertlo:
This looks really close to me. Thanks for your continued work on this.

src/config/musl.md
flexibeast: i added this following @q66's comment [here](https://github.com/void-linux/void-docs/pull/163#discussion_r395958839), but i'm certainly happy to remove it if @q66 is fine with that. :-)

src/config/musl.md
flexibeast: Fair question! That seems like a distinct possibility; this was information i merely ported from the wiki. Should i test this in my musl Void VM?

src/config/musl.md
flexibeast: Changed.

src/config/musl.md
flexibeast: Changed.

src/config/musl.md
flexibeast: i've now removed "For example" and the text following that.

src/installation/base-requirements.md
flexibeast: Changed.

src/config/musl.md
bobertlo: I'm inclined to think we can just skip this, and maybe come back later.

src/config/musl.md
bobertlo: Oh, sorry i didn't catch that. /etc/hosts isn't strictly required for network access. I don't feel super strongly about this. I guess I would be fine if `/etc/hosts` was qualified.

flexibeast: Updated.

src/config/musl.md
flexibeast: Removed.

PR #165, opened 2020-03-17
locales: Add section.

flexibeast:


ericonr:
What do you think about mentioning `/etc/locale.conf`, as well? It's where the actual locale is selected.

flexibeast:
@ericonr: Good suggestion, thanks! PR updated.

bobertlo:
I think a layout like the following might be more helpful:

# Locales

For a list of currently enabled locales run:

```
locale -a
```
## Enabling locales

To enable the locale `xxxx` un-comment or add the proper line in `/etc/default/libc-locale.conf` and run

```
# xbps-reconfigure -f glibc-locales
```

## Setting the System locale

Then set `LANG=xxxx` in `/etc/locale.conf`

flexibeast:
@bobertlo: Thanks, updated.

PR #168, opened 2020-03-20
wpa_supplicant: add information about wpa_cli usage.

flexibeast:


bobertlo:
I'm not sure whether we want to add on sections like these for tools that are well documented. I don't feel super strongly either way. If we did include this I would also mention that settings will take effect immediately but will be lost on reboot unless you run the `save` command. ;)

flexibeast:
@bobertlo: Well, the reason i want to add this is that i've noticed a few people, including myself, get bitten in the last couple of weeks when one *doesn't* specify an interface on the command line. The `wpa_cli(1)` man page describes `-i` with "By default, choose the first interface found with a control socket in the socket path"; but it certainly wasn't obvious to me that this implies `dev-p2p-wlp2s0` would get chosen rather than `wlp2s0`. :-)

So that's my reasoning. If other members of the Void team have stronger feelings against this, i'll let it go, but otherwise, i'd be happy to add the extra information you mentioned. :-)

bobertlo:
I don't care too much about the other info. This seems like a legitimate concern. Merging.

PR #169, opened 2020-03-24
config/wayland: add page

ericonr:
Add a Wayland page, loosely based on what's available on the wiki. I can add/remove/correct information according to suggestions and opinions here.

### DIFF
--- a/src/config/wayland.md
+++ b/src/config/wayland.md
@@ -1,0 +1,92 @@
+# Wayland
+
+This section details the manual installation and configuration of Wayland
+compositors and related services and utilities.
+
+## Installation
+
+Unlike [Xorg](./xorg/index.md), Wayland implementations combine the display
+server, the window manager and the compositor in a single application.
+
+### Desktop Environments
+
+Both GNOME and KDE Plasma have Wayland sessions. For GNOME, Wayland is already
+the default session, while Plasma still has a few issues to iron out, so their
+default remains a Xorg session. While running these desktop environments,
+applications built with Gtk+ will automatically choose the Wayland backend, but
+Qt applications might require the proper environment variable in GNOME. This is
+explained [below](#native-applications).
+
+### Stand-alone compositors
+
+Void Linux currently packages the following Wayland compositors:
+
+- Weston: reference compositor for Wayland
+- Sway: an i3-compatible Wayland compositor
+- Wayfire: 3D Wayland compositor
+- Cage: a Wayland kiosk
+
+### Video drivers
+
+KDE Plasma has an implementation for the proprietary NVIDIA API used for
+Wayland. The other Wayland compositors require open source drivers that
+implement EGL. The main driver for this purpose is provided by the `mesa-dri`
+package. The [graphics section](./graphics/index.md) has more details for
+setting up graphics in different systems.
+
+### Native applications
+
+Qt5 based applications require the installation of the `qt5-wayland` package and
+setting the environment variable `QT_QPA_PLATFORM=wayland-egl` to enable their
+Wayland backend. Some KDE specific applications might also require the
+installation of the `kwayland` package. Gtk+ applications should do it
+automatically, but setting `GDK_BACKEND=wayland` can force the Wayland backend
+(however, this can break applications such as Chromium).
+
+#### Web browsers
+
+Mozilla Firefox ships with a Wayland backend which is disabled by default.
+Launching Firefox with `MOZ_ENABLE_WAYLAND=1` is required to enable it.
+
+The Midori browser, which has a Gtk+ interface, should also work on Wayland.
+
+Qt5 based browsers, such as qutebrowser, also work on Wayland.
+
+#### Terminal emulators
+
+- Alacritty (it's the standard terminal emulator for Sway)
+- Kitty
+- Sakura
+
+#### Media applications
+
+Both mpv and VLC work on Wayland, with hardware acceleration (if available and
+enabled).
+
+For image viewing, imv works natively on Wayland.
+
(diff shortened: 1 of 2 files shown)


ericonr:
A little update based on feedback from IRC.

src/config/wayland.md
```diff
@@ -0,0 +1,106 @@
+# Wayland
+
+This section details the manual installation and configuration of Wayland
+compositors and related services and utilities.
+
+## Installation
+
+Unlike [Xorg](./xorg/index.md), Wayland isn't implemented as a server that
+applications like a window manager and compositor talk to. Instead, the Wayland
+compositor implements all functions related to displaying windows on the screen.
```
st3r4g: I'd reword this to make it clearer:
```suggestion
Unlike [Xorg](./xorg/index.md), Wayland implementations combine the display
server, the window manager and the compositor in a single application.
```

src/config/wayland.md
```diff
@@ -0,0 +1,106 @@
...
+
+```
+if test -z "${XDG_RUNTIME_DIR}"; then
+    export XDG_RUNTIME_DIR=/tmp/${UID}-runtime-dir
+    if ! test -d "${XDG_RUNTIME_DIR}"; then
+        mkdir -p "${XDG_RUNTIME_DIR}"
+        chmod 0700 "${XDG_RUNTIME_DIR}"
+    fi
+fi
+```
```
st3r4g: Although it appears [here](https://wayland.freedesktop.org/building.html), @Duncaen pointed out that this snippet has some flaws, but I don't remember the details now.

src/config/wayland.md
```diff
@@ -0,0 +1,106 @@
...
+compositors, and is installed as a dependency for most of them. Its package is
+`xorg-server-xwayland`. For Weston, the correct package is `weston-xwayland`.
+
+## Configuration
+
+The Wayland API uses the `XDG_RUNTIME_DIR` environment variable to determine the
+directory the Wayland socket. If this environment isn't already set in your
+setup, you can add this snippet to your shell profile or to a script that
+starts-up your Wayland compositor of choice. This shouldn't be necessary with
+any of the desktop environments.
```
st3r4g: You could mention that `elogind` sets this up for you.

src/config/wayland.md
```diff
@@ -0,0 +1,106 @@
...
+        mkdir -p "${XDG_RUNTIME_DIR}"
+        chmod 0700 "${XDG_RUNTIME_DIR}"
+    fi
+fi
+```
+
+It is also possible that some applications use the `XDG_SESSION_TYPE`
+environment variable in some way, which requires that you set it to `wayland`.
+
+## Video drivers
```
st3r4g: This part is important, I wouldn't put it at the end. Having `sway` crash because you forgot the drivers is a common mistake, and not immediately recognized by everybody. Ideally, the drivers would be pulled in as a dependency as soon as you install a GL application (Wayland compositor), but this requires `mesa` to be a dependency of `libglvnd` which is not trivial at the moment.

src/config/wayland.md
ericonr: That does sound better. Thanks!

src/config/wayland.md
ericonr: As you mentioned below, elogind also sets it up, so it's definitely a solution. However, I feel like including the script could make life easier for some people (as it did for me, as a matter of fact). If it has issues, though, it's probably a good idea to remove it, seeing as we want the Void documentation to be properly curated.

ericonr: I'm also thinking of including the `XDG_SESSION_TYPE` variable, and the fact that you're supposed to set it to `wayland`. Do you know any applications that use its value? My own scripts do, and I feel like it's the "correct" thing to do, but I have no idea if it's necessary.

src/config/wayland.md
ericonr: > this requires mesa to be a dependency of libglvnd which is not trivial at the moment.

Is there any help I can give in that front? Would it be a packaging issue?

I mostly followed the style of the Xorg page for the position of everything. I will see if I can find a better location for Drivers.

ericonr: I moved the Video Drivers to below DEs and WMs, so I could refer to them without issue.

src/config/wayland.md
Duncaen: `$XDG_RUNTIME_DIR` is usually on `/run` because only privileged users can create directories there opposed to `/tmp` where this could haven been maliciously created before your shell script runs on login.

ericonr: Considering this is a possible security hazard, do you think it should be documented as such, or fully removed?

flexibeast: i would prefer the documentation include a (more) secure way of setting up `XDG_RUNTIME_DIR`, since (a) it's necessary for Wayland, and (b) if people aren't provided with a more secure way, they'll instead use whatever 'solution' seems to 'work' - whether that's something they've put together themselves, or something they've just copied and pasted from somewhere else.

xtraeme: The right solution is to use elogind.

ericonr: All right, I will leave elogind as the only option. Do you think it's best to explain why elogind is necessary (it has to create the directory and set `XDG_RUNTIME_DIR`), or should I just mention that elogind is necessary for wayland and leave it at that?

ericonr: Fixed.

bobertlo:
I really like the look of this. I have been waiting for the dust to settle because I am less savvy with Wayland. I will try to do a review and get this merged ASAP. Thanks for contributing!

PR #170, opened 2020-03-28
improve size estimate for /boot partition

dicktyr:


bobertlo:
Nice, can confirm. Maybe it would be appropriate to say 'at least 20-60MB then? Also could you please format this? 

See:
https://docs.voidlinux.org/contributing/void-docs/style-guide.html#formatting

And yes I can confirm this on an x86_64 install:
```
# du -h /boot
2.3M	/boot/grub/fonts
2.7M	/boot/grub/i386-pc
5.0M	/boot/grub
125M	/boot
# vkpurge rm all
Removing kernel 5.4.27_1...
Removing /boot/config-5.4.27_1...
...
# du -h /boot
2.3M	/boot/grub/fonts
2.7M	/boot/grub/i386-pc
5.0M	/boot/grub
62M	/boot
```

dicktyr:
> say 'at least 20-60MB then?

that line refers to just an image file

> Also could you please format this? 

sorry I used GitHub's web interface

I'm not familiar with git
and I prefer to avoid local installation

does that mean web-GUI PRs are not useful?

bobertlo:
Oh no problem then. Thanks for the info I'll merge it manually.

dicktyr:
> Oh no problem then. Thanks for the info I'll merge it manually.

I was confused thinking vmdfmt was doing something clever
but I guess it was just an issue of line length/wrapping
and indeed a web PR is fine (if it matches vmdfmt)

bobertlo:
Oh yeah. I think maybe we should just say that each kernel version takes a significant amount of space, and then specify that 'for example a default `x86_64` kernel version takes almost 60MB in /boot at the time of writing'

I don't want to have more specific numbers than necessary to try to update :)

dicktyr:
> x86_64

good point

> I don't want to have more specific numbers than necessary to try to update :)

of course
but "at the time of writing" doesn't seem useful
and I think it does deserve occasional updating
because it's important for those who want to isolate /boot
and I once made the mistake of trusting numbers from Arch
when partitioning an SSD (at a time when capacity was precious)

so how's this then?

> On most modern systems, a separate `/boot` partition is no longer necessary to boot properly. If you choose to use one, note that Void does not remove old kernels after updates by default and each kernel version will take at least 20-60MB, so plan accordingly (e.g. `/boot` with one `x86_64` Linux 5-series kernel and grub occupies about 60MB).

bobertlo:
I'm just trying to be realistic about the amount this will get updated and 'at the time of writing' is never untrue :)

Do you have a basis for 20-60MB (specifically the low end) or is this based on current info? I don't want to give a lot of specifics because there are a lot of different platforms too. Giving just the 60MB example for `x86_64` will cover a majority of users and is based on data at hand. If more first hand info is available I am willing to add that.

Not trying to argue, just explain that my reasoning was deeper than being lazy!

dicktyr:
"20MB" is from the current document

how's this?

> On most modern systems, a separate `/boot` partition is no longer necessary to boot properly. If you choose to use one, note that Void does not remove old kernels after updates by default and also that the kernel tends to increase in size with each new version, so plan accordingly (e.g. `/boot` with one `x86_64` Linux 5.x kernel and grub occupies about 60MB).

bobertlo:
Great!

bobertlo:
Feel free to update this or I will just copy and paste and update your commit after work.

dicktyr:
updated

bobertlo:
Great. Thank you for contributing!

Issue #49, opened 2019-02-03
config/xorg: Session and seat management documentation


nilium:
**Source:** [config/xorg/index.md](https://github.com/void-linux/void-docs/blob/master/src/config/xorg/index.md)

This needs to be written by someone with knowledge of the subject, and it's not clear what needs to be covered in the first place. Attempts so far have filled out bits of the documentation (such as #38) but topics like using dbus-launch, ConsoleKit2, elogind, and other programs / services probably merits enough depth so that it's clear to users what the right thing to do is.

It'd also help to collect examples of previous questions (from reddit, the mailing list, IRC, etc.) and their solutions (when available) to figure out what troubleshooting problems come up.

bobertlo:
Upon learning for the first time about the interconnected nature of dbus/bluetoothd/ConsoleKit/PulseAudio/PolicyKit/etc, I agree much more these should be documented... grudgingly.

I am starting to wonder though, if there should be a separate section documenting some sort of compatibility layer with Desktop Linux(™). I'm not being facetious. I have a couple machines I could test the 'flavor images' on, but I have no good idea how we can support these things from the base image gracefully.

I'm going to now go nuke all of this software I've tried to test off my EDC ThinkPad, bbl.

ericonr:
We should at least remove ConsoleKit2 from there, as it seems to have been mostly superseded by elogind, and isn't update anymore. I opened an issue to remove it completely from packages here https://github.com/void-linux/void-packages/issues/20577, if it is I can try and rewrite a bit this one.

@bobertlo What do you think about splitting up session management into its own subject at the same level as Xorg and Wayland? It pertains to both of them (and to a tty session too, as a matter of fact), not only to Xorg.

bobertlo:
@ericonr Definite yes on ConsoleKit2. It's actually removed from the templates repo, just still on the package server.

For moving the page, let's think about that. Maybe we could put that section and xorg under graphics? Or it might fit better in a 'Desktop Environments' section later? I think it would be clearer as a sub section vs just 'Session Management' alone. That's kind of silly reasoning but IMO true :)

ericonr:
@bobertlo yeah, I noticed the removal. Seems there's a fix on the way to remove the actual package.

[...]

PR #185, opened 2020-04-10
External applications

anjandev:
Im unsure if this is something to put in the void-docs repo. I was getting a cryptic error when I used `gem install jekyll` after installing ruby from void repos. This was fixed after alot of googling and figuring out I needed to install the `ruby-devel` package.

### DIFF
--- a/src/config/external-applications.md
+++ b/src/config/external-applications.md
@@ -52,8 +52,8 @@ Flatpak documentation](https://flatpak.org/setup/Void%20Linux/).
 ### Troubleshooting
 
 Some apps may not function properly (e.g. not being able to access the host
-system's files). This may be fixed by installing one or more of the
-`xdg-desktop-portal`, `xdg-desktop-portal-gtk`, `xdg-user-dirs`,
+system's files). Some of these issues can be fixed by installing one or more of
+the `xdg-desktop-portal`, `xdg-desktop-portal-gtk`, `xdg-user-dirs`,
 `xdg-user-dirs-gtk` or `xdg-utils` packages.
 
 Some Flatpaks require D-Bus. If you are using a window manager (e.g. i3wm), you
--- a/src/config/external-applications.md
+++ b/src/config/external-applications.md
@@ -1,0 +1,68 @@
+# External Applications
+
+## Programming Languages
+
+The Void repositories have a number of Python and Lua packages. If possible,
+install packages from the Void repositories or consider packaging the library or
+application you need. Packaging your application allows for easier system
+maintenance and can benefit other Void Linux users, so consider making a pull
+request for it. The contribution instructions can be found
+[here](https://github.com/void-linux/void-packages/blob/master/CONTRIBUTING.md).
+
+To keep packages smaller, Void has separate `devel` packages for header files
+and development tools. If you install a library or application via a language's
+package manager (e.g. `pip`, `gem`), or compile one from source, you may need to
+install the programming language's `-devel` package. This is specially relevant
+for `musl` libc users, due to pre-built binaries usually targeting `glibc`
+instead.
+
+| Language | Package Manager                | Void Package    |
+|----------|--------------------------------|-----------------|
+| Python3  | pip, anaconda, virtualenv, etc | `python3-devel` |
+| Python2  | pip, anaconda, virtualenv, etc | `python2-devel` |
+| Ruby     | gem                            | `ruby-devel`    |
+| lua      | luarocks                       | `lua-devel`     |
+
+## Restricted Packages
+
+Some packages have legal restrictions on their distribution (e.g. Discord), may
+be too large, or have another condition that makes it difficult for Void to
+distribute. These packages have build templates, but the packages themselves are
+not built or distributed. As such, they must be built locally. For more
+information see the page on [restricted
+packages](../repositories/restricted.md).
+
+## Non-x86_64 Arch
+
+The Void build system runs on x86_64 servers, both for compiling and cross
+compiling packages. However, some packages (e.g. `libreoffice`) do not support
+cross-compilation. These packages have to be built locally on a computer running
+the same architecture and libc as the system on which the package is to be used.
+For building packages, see the [building instructions](../packages/building.md).
+
+## Flatpak
+
+Flatpak is another method for installing external proprietary applications on
+Linux. For information on using Flatpak with Void Linux, see the [official
+Flatpak documentation](https://flatpak.org/setup/Void%20Linux/).
+
+> Flatpak's sandboxing will not necessarily protect you from any security and/or
+> privacy-violating features of proprietary software.
+
+### Troubleshooting
+
+Some apps may not function properly (e.g. not being able to access the host
+system's files). This may be fixed by installing one or more of the
+`xdg-desktop-portal`, `xdg-desktop-portal-gtk`, `xdg-user-dirs`,
+`xdg-user-dirs-gtk` or `xdg-utils` packages.
+
+Some Flatpaks require D-Bus. If you are using a window manager (e.g. i3wm), you
+can launch your window manager with
+[dbus-run-session(1)](https://man.voidlinux.org/dbus-run-session.1). Some
+Flatpaks can also require [Pulseaudio](./media/pulseaudio.md).
+
+## Octave Packages
+
+Some Octave packages require external dependencies to compile and run. For
+example, to build the control package, you must install the `openblas-devel`,
+`libgomp-devel`, `libgfortran-devel`, `gcc-fortran`, and `gcc` packages.
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -45,6 +45,7 @@
       - [sndio](./config/media/sndio.md)
    - [Bluetooth](./config/bluetooth.md)
    - [TeX Live](./config/texlive.md)
+   - [External Applications](./config/external-applications.md)
    - [Printing](./config/print/index.md)
       - [Installing Printing Drivers](./config/print/drivers.md)
       - [Configuring a New Printer](./config/print/config.md)


ericonr:
Could you make a more general entry? For example, building some python modules can also require the presence of python3-devel. I think Lua modules require lua-devel too.

You could make an entry for something like "building and installing external modules for programming languages", and explain that the devel package might be necessary.  More importantly, for people who use musl, the pre-built binaries that are sometimes available definitely won't work for them, so they have an even higher change of encountering this issue and needing to install something-devel packages.

anjandev:
@ericonr I would be happy to make this more general however I have no experience installing from pip - I package all my pip modules via xbps-src. Furthermore, I have no experience with lua.

I can change the title to something more generic and have this as a PR to add this page with ruby as the first entry and maybe people who have experience with other languages can add onto what I have.

anjandev:
I added python details but I need help with lua modules.

ericonr:
So, @anjandev, first of all, sorry for the delay in getting here!

Have you taken a look at #188? The idea is to put all this information together into a single section. Do you have any experience with those other subjects? If not, I can help you with writing them, or I can adopt the PR (not sure how that would work).

Regarding the specific languages:

* `{ruby,python,python3}-devel` packages are dependencies only for other devel packages. That means it makes sense to mention them in this.
* `lua-devel` does as well, but I'd have to look into if it can be really necessary for external packages (probably)

That said, I feel like making a subheader for each language can be overkill. We should, perhaps in a table (https://www.tablesgenerator.com/markdown_tables):

* mention each language
* mention an example of a package manager for it (gems, pip, etc) (only an example, because python for one has an infinity of them)
* specify the devel package

Or we could really only list the languages and tell the user to grab the respective devel package if an external package fails to build.

@bobertlo do you have any strong opinions on this?

src/config/programming-languages.md
```diff
@@ -0,0 +1,14 @@
+# Building and Installing External Modules for Programming Languages
+
+To keep your system lean, Void splits packages and their devel packages. If you
+install program via pip, gem, etc. you may need to install the programming
+language's devel package to ensure your program builds.
+
+## Python
+
+To ensure python modules install correctly via pip, install the `python-devel`
```
ericonr, line 9: Also, I hope most Void users are using Python3 for now. The package for that one is `python3` c:

anjandev: Fixed

anjandev:
@ericonr I have some experience with flatpaks, building restricted packages with xbps, etc. I can try my hand at it. I have given the maintainers the ability to edit my PR so you guys can certainly help wherever you feel it's necessary. I will try to merge your other requested changes when I have the time.

ericonr:
@anjandev should be quite brief. I'd say that for now we should write only the first 3 items from the list that's in the issue.

Most of the paragraphs will be very simple.

* "restricted packages exist, bla bla" -> link to restricted packages page, which already exists
* "flatpak exists, does this, has applications like spotify and steam, install `flatpak` package"
* "languages etc" (the one we discussed here) 
 
Probably the most complicated one will be the explanation about the types of restricted software that exist.

anjandev:
@ericonr changes have been made

src/config/external-applications.md
```diff
@@ -0,0 +1,51 @@
...
+
+> The Void repositories have a number of python and lua packages. If possible,
+> try to install packages from the Void repos using xbps-install or use xbps-src
+> to package your python or lua application. Packaging your application with
+> xbps-src will keep your system clean. If you send your package upstream it
+> will allow other Void users to benifit and allow them to help you maintain
+> your package. To learn how to package for xbps-src see the
+> [manual](https://github.com/void-linux/void-packages/blob/master/Manual.md)
+> and [contributing
+> instructions](https://github.com/void-linux/void-packages/blob/master/CONTRIBUTING.md).
```
ericonr, line 13: To start off, use Python, Lua and Linux when referring to them, unless you're talking about the package itself.

I'm having a complicated thought process regarding this one. People who develop in those languages sometimes don't really have the option of depending on distro packages, due to versioning issues and a bunch of other stuff. Perhaps something closer to:

```suggestion
> The Void repositories have a number of Python and Lua packages. If possible,
> install packages from the Void repositories or consider packaging the library or application you need.
> Packaging your application allows for easier system maintenance and can benefit
> other Void Linux users, so consider making a pull request for it. The contribution instructions can be found
>  [here](https://github.com/void-linux/void-packages/blob/master/CONTRIBUTING.md).
```

src/config/external-applications.md
```diff
@@ -0,0 +1,51 @@
...
+
+To keep your system lean, Void splits packages and their devel packages. If you
+install a program via the languages package manager (pip, gem, etc.) or compile
+via source you may need to install the programming language's devel package to
+ensure your program builds.
```
ericonr, line 18:
```suggestion
To keep packages smaller, Void has separate `devel` packages for header files and development tools. 
If you install a library or application via a language's package manager (i.e. `pip`, `gem`) or compile
one from source, you may need to install the programming language's `-devel` package. This is specially
relevant for `musl` libc users, due to pre-built binaries usually targeting `glibc` instead.
```

src/config/external-applications.md
```diff
@@ -0,0 +1,51 @@
...
+
+| Language | Package Manager                | Void Package  |
+|----------|--------------------------------|---------------|
+| Python   | pip, anaconda, virtualenv, etc | python3-devel |
+| Ruby     | gem                            | ruby-devel    |
+| lua      | luarocks                       | lua-devel     |
```
ericonr, line 24: Use code blocks: `python3-devel`, `ruby-devel` and `lua-devel`. I'm not completely against making a section for Python 2 and one for Python 3, either.

src/config/external-applications.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Restricted Packages
+
+Some packages have legal restrictions on their distribution (ie. discord), may
+be too large, or have another condition that makes it difficult for the void
+team to distribute. These packages are maintained but not distributed. As such,
+they must be built locally. For more information see the page on [restricted
+packages](../repositories/restricted.md).
```
ericonr, line 32:
```suggestion
Some packages have legal restrictions on their distribution (i.e. Discord), may
be too large or have another condition that makes it difficult for the Void
team to distribute. These packages are maintained but not distributed. As such,
they must be built locally. For more information see the page on [restricted
packages](../repositories/restricted.md).
```

src/config/external-applications.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Non-x86_64 Arch
+
+The entire Void build system is using the x86_64 architecture. xbps-src allows
+us to cross-compile for other architectures and their glibc or musl variant.
+However, some packages (ie. libreoffice) do not support cross-compilation. For
+these appplications, you must build the application locally on the machine you
+want to use the package. For compiling and installing xbps-src applications see
+the [building instructions](../packages/building.md).
```
ericonr, line 41:
```suggestion
The Void build system runs on x86_64 computers, both for compiling and cross compiling packages.
However, some packages (i.e. `libreoffice`) do not support cross-compilation.
These packages have to be built locally on a computer running the same version of Void Linux as the one running
where you want to use the package. For building packages,  see the [building instructions](../packages/building.md).
```

src/config/external-applications.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Flatpak
+
+Flatpak is another method for installing external proprietary applications on
+linux. For information on using flatpak with Void Linux, see the [official
+flatpak documentation](https://flatpak.org/setup/Void%20Linux/).
```
ericonr, line 47: You need to capitalize Linux and Flatpak in this paragraph. And I think "external" in this case doesn't need to be mentioned. If it is proprietary, it is external already.

Perhaps add "proprietary applications, such as Spotify and Steam."

@bobertlo do you think we have to mention that Steam also has a package? With zero guarantees, as it stands.

src/config/external-applications.md
```diff
@@ -0,0 +1,51 @@
...
+
+> Flatpak's sandboxing will not protect you from the malicious features
+> proprietary applications may have. Whenever possible, try to use free software
+> with Void Linux.
```
ericonr, line 51: I really like this paragraph, I just don't know if it fits the tone we want to have on the docs.

@bobertlo what do you think?

flexibeast: i'm not sure the last sentence is necessary; i think it would be sufficient to just say something like "Flatpak's sandboxing will not necessarily protect you from any security- and/or privacy-violating features of proprietary software".

anjandev: I think proprietary applications usually have more malicious features than just merely security issues and privacy violations.

ie: vendor lock-in, censorship, deception, using one software to enforce the use of another one of your corporation's offerings.

It's more accurate to mention malicious features or anti-features.

flexibeast: i don't disagree with your points in general - although i would disagree with any implication that FOSS never has anti-features - but the Handbook is technical documentation, and i'm concerned about starting down a steep slippery slope. i'd prefer that we didn't set a precedent for the Handbook to become a place where people argue over licensing issues. So i think that what we need to focus on is letting people know that 'sandboxing' is not a magical protective incantation.

(i regularly encounter people who think that merely using OpenBSD means their system is secure, regardless of what software they install and use.)

anjandev: That is a good point. If anyone has a recommendation to change the wording such that the tone is better while ensuring users know that sandboxing is not magic, I will accept it.

src/config/external-applications.md
anjandev: I added both your suggestions

anjandev:
@ericonr I kind of brainfarted and messed up using github's webgui and git locally. It should be fixed now please check if you can. I have merged most of your changes except for the last recommendation which I will change as soon as someone recommends good wording.

src/config/external-applications.md
ericonr: @anjandev, I really like flexibeast's wording.

> Flatpak's sandboxing will not necessarily protect you from any security- and/or privacy-violating features of proprietary software.

It's short and sweet.

ericonr:
I had a quick look, it looks very good. Once we decide on that last paragraph I can make a final pass.

src/config/external-applications.md
anjandev: Changed to @flexibeast's wording

anjandev:
I was having some trouble with flatpaks today so I copied some of the info from the void wiki to the page.

https://wiki.voidlinux.org/Flatpak

src/config/external-applications.md
```diff
@@ -0,0 +1,85 @@
...
+
+> Flatpak's sandboxing will not necessarily protect you from any security and/or
+> privacy-violating features of proprietary software.
+
+However, some apps may not function properly (e.g. not being able to access the
+host system's files). To fix this, open a console and enter:
+
+```
+# xbps-install -S xdg-desktop-portal xdg-desktop-portal-gtk xdg-user-dirs xdg-user-dirs-gtk xdg-utils
+```
```
ericonr, line 57: You should list the packages that need to be installed, instead of giving the xbps-install command.

src/config/external-applications.md
```diff
@@ -0,0 +1,85 @@
...
+# flatpak search org.gtk.Gtk3theme
+```
+
+### Troubleshooting
+
+Some Flatpaks need dbus. Enable dbus with the command
+
+```
+# ln -s /etc/sv/dbus /var/service/
+```
```
ericonr, line 82: Same for this one. Just say that the `dbus` service needs to be enabled.

ericonr: Also, the name is "D-Bus"if you're not taking about the package.

src/config/external-applications.md
```diff
@@ -0,0 +1,85 @@
...
+
+```
+# flatpak install -y flathub org.gtk.Gtk3theme.Plata-Noir
+```
+
+You can search for additional Gtk themes with the command
+
+```
+# flatpak search org.gtk.Gtk3theme
+```
```
ericonr, line 74: Is there some official place where this documentation can be found? It might be better to link to that instead. Otherwise I think it's okay to leave it.

anjandev: I could not find this info anywhere else. Leaving it in

src/config/external-applications.md
anjandev: Changed. Thanks

src/config/external-applications.md
anjandev: Changed. Thanks

anjandev:
@ericonr I have never gotten flatpak to really work correctly with all apps on any distro so I don't really know. 

Regardless, I think we should port the instructions for polkit from here: 

https://wiki.voidlinux.org/Post_Installation

src/config/external-applications.md
```diff
@@ -0,0 +1,79 @@
...
+Linux. For information on using Flatpak with Void Linux, see the [official
+Flatpak documentation](https://flatpak.org/setup/Void%20Linux/).
+
+> Flatpak's sandboxing will not necessarily protect you from any security and/or
+> privacy-violating features of proprietary software.
+
+However, some apps may not function properly (e.g. not being able to access the
+host system's files). To fix this, install the following packages:
+xdg-desktop-portal xdg-desktop-portal-gtk xdg-user-dirs xdg-user-dirs-gtk
+xdg-utils
```
ericonr, line 55: Remember to use backticks for formatting package names :)

```
`xdg-utils`
```

ericonr:
@anjandev I will take a look at my current setup with flatpak. Regarding polkit, I mentioned it here https://github.com/void-linux/void-docs/issues/49 .

If you feel up to it ;)

src/config/external-applications.md
anjandev: fixed in latest patch

src/config/external-applications.md
```diff
@@ -0,0 +1,79 @@
...
+Linux. For information on using Flatpak with Void Linux, see the [official
+Flatpak documentation](https://flatpak.org/setup/Void%20Linux/).
+
+> Flatpak's sandboxing will not necessarily protect you from any security and/or
+> privacy-violating features of proprietary software.
+
+However, some apps may not function properly (e.g. not being able to access the
+host system's files). To fix this, install the following packages:
+`xdg-desktop-portal` `xdg-desktop-portal-gtk` `xdg-user-dirs`
+`xdg-user-dirs-gtk` `xdg-utils`
```
ericonr, line 55: Perhaps some commas between them too? Then an "and" before the last one too. More like a list.

ericonr:
About polkit. flatpak enables it by itself.

src/config/external-applications.md
```diff
@@ -0,0 +1,79 @@
...
+
+```
+# flatpak install -y flathub org.gtk.Gtk3theme.Plata-Noir
+```
+
+You can search for additional Gtk themes with the command
+
+```
+# flatpak search org.gtk.Gtk3theme
+```
```
ericonr, line 72: These commands don't need root, right? So the most correct way to display them would be using `$` instead of `#`.

ericonr:
@anjandev I think that's my last comment, the rest looks good to me! Thanks a lot :D

src/config/external-applications.md
```diff
@@ -0,0 +1,87 @@
...
+$ flatpak search org.gtk.Gtk3theme
+```
+
+### Octave Packages
+
+Some octave packages require external dependencies to compile and run.
+
+For example to install the control package, you must install the:
+`openblas-devel`, `libgomp-devel`, `libgfortran-devel`, `gcc-fortran`, and `gcc`
+packages.
```
ericonr, line 80: This shouldn't be under Flatpak, should it? I think the header should be `##` only. 

Suggestions:

* "Some Octave packages"
* everything in a single paragraph
* "For example, to build" -> make it clear it's building something instead of just downloading it. I would also do "the `openblas-devel` ...", without a colon.

anjandev: changes added

src/config/external-applications.md
```diff
@@ -0,0 +1,85 @@
...
+
+```
+$ flatpak search org.gtk.Gtk3theme
+```
+
+## Octave Packages
+
+Some Octave packages require external dependencies to compile and run. For
+example, to build the control package, you must install the `openblas-devel`
+`libgomp-devel` `libgfortran-devel` `gcc-fortran` `gcc` packages.
```
ericonr, line 78: You removed the colon, but please keep the commas and "and" :p

anjandev:
> You removed the colon, but please keep the commas and "and" :p

I pushed a new commit with this change. Thanks

src/config/external-applications.md
```diff
@@ -0,0 +1,85 @@
...
+Some Octave packages require external dependencies to compile and run. For
+example, to build the control package, you must install the `openblas-devel`,
+`libgomp-devel`, `libgfortran-devel`, `gcc-fortran`, and `gcc` packages.
+
+### Troubleshooting
+
+Some Flatpaks need D-Bus. Enable the dbus service with runit.
+
+Some Flatpaks require Pulseaudio to be set up properly, see
+[Pulseaudio](./media/pulseaudio.md).
```
ericonr, line 85: Suggest `dbus` instead of dbus. Since you're repeating "Some Flatpaks", I'd get these in the same paragraphs, like

"... runit. They can also require Pulseaudio..."

ericonr, line 85: And this header should be moved to above the octave one, since it should be under Flatpak ;)

src/config/external-applications.md
```diff
@@ -0,0 +1,83 @@
+# External Applications
+
+## Programming Languages
+
+> The Void repositories have a number of Python and Lua packages. If possible,
+> install packages from the Void repositories or consider packaging the library
+> or application you need. Packaging your application allows for easier system
+> maintenance and can benefit other Void Linux users, so consider making a pull
+> request for it. The contribution instructions can be found
+> [here](https://github.com/void-linux/void-packages/blob/master/CONTRIBUTING.md).
```
flexibeast, line 10: Further to [this recent comment of mine](https://github.com/void-linux/void-docs/pull/217#discussion_r419204420), i think this text shouldn't be quoted, particularly if the intent of the quoting is to achieve layout effects - i'd prefer to keep our markup as semantic as possible.

src/config/external-applications.md
```diff
@@ -0,0 +1,83 @@
...
+
+To keep packages smaller, Void has separate `devel` packages for header files
+and development tools. If you install a library or application via a language's
+package manager (i.e. `pip`, `gem`) or compile one from source, you may need to
```
flexibeast, line 14: "e.g." not "i.e."

Comma before "or".

src/config/external-applications.md
```diff
@@ -0,0 +1,83 @@
...
+| Language | Package Manager                | Void Package    |
+|----------|--------------------------------|-----------------|
+| Python3  | pip, anaconda, virtualenv, etc | `python3-devel` |
+| Python2  | pip, anaconda, virtualenv, etc | `python2-devel` |
+| Ruby     | gem                            | `ruby-devel`    |
+| lua      | luarocks                       | `lua-devel`     |
+
+## Restricted Packages
+
+Some packages have legal restrictions on their distribution (i.e. Discord), may
```
flexibeast, line 28: "e.g." not "i.e."

src/config/external-applications.md
```diff
@@ -0,0 +1,83 @@
...
+be too large or have another condition that makes it difficult for the Void team
```
flexibeast, line 29: Comma before "or".

Suggest "Void" rather than "the Void team".

src/config/external-applications.md
```diff
@@ -0,0 +1,83 @@
...
+to distribute. These packages are maintained but not distributed. As such, they
```
flexibeast, line 30: Hmm, i think i'd prefer this to say something more like:

"These packages have build templates, but the packages themselves are not built or distributed."

src/config/external-applications.md
```diff
@@ -0,0 +1,83 @@
...
+must be built locally. For more information see the page on [restricted
+packages](../repositories/restricted.md).
+
+## Non-x86_64 Arch
+
+The Void build system runs on x86_64 computers, both for compiling and cross
```
flexibeast, line 36: s/system/server/? Otherwise it might be read as implying that users must use x86_64 to build packages.

src/config/external-applications.md
```diff
@@ -0,0 +1,83 @@
...
+compiling packages. However, some packages (i.e. `libreoffice`) do not support
+cross-compilation. These packages have to be built locally on a computer running
+the same version of Void Linux as the one running where you want to use the
```
flexibeast, line 39: Presuming that by "version" we mean "architecture" - and, i assume, libc - i'd suggest something along the lines of:

"These packages have to be built locally on a Void system with the same architecture and libc as the system on which the package is to be used."

src/config/external-applications.md
```diff
@@ -0,0 +1,83 @@
...
+
+Flatpak is another method for installing external proprietary applications on
+Linux. For information on using Flatpak with Void Linux, see the [official
+Flatpak documentation](https://flatpak.org/setup/Void%20Linux/).
+
+> Flatpak's sandboxing will not necessarily protect you from any security and/or
+> privacy-violating features of proprietary software.
+
+However, some apps may not function properly (e.g. not being able to access the
+host system's files). To fix this, install the following packages:
```
flexibeast, line 53: i imagine not all Flatpak issues can be solved simply by installing the `xdg-*` packages, so i suggest something like:

"In some instances, this may be fixed by installing one or more of the \`xdg-desktop-portal\`, \`xdg-desktop-portal-gtk\`, \`xdg-user-dirs\`, \`xdg-user-dirs-gtk\` or \`xdg-utils\` packages."

src/config/external-applications.md
```diff
@@ -0,0 +1,83 @@
...
+`xdg-desktop-portal`, `xdg-desktop-portal-gtk`, `xdg-user-dirs`,
+`xdg-user-dirs-gtk`, and `xdg-utils`.
+
+### Installing Themes
+
+You can install themes with Flatpak and Flatpak will automatically match your
+computer's Gtk theme to one installed with Flatpak. For example, if you use the
```
flexibeast, line 60: "GTK"

Does this mean that as soon as one installs a theme Flatpak, one's system will automatically use that theme? Does one have to do a `flatpak install` whenever one wants to change Flatpak theme, even if one wants to change back to a Flatpak theme installed previously?

src/config/external-applications.md
```diff
@@ -0,0 +1,83 @@
...
+theme Plata-Noir in your regular desktop environment, you can install the
+Flatpak counterpart with the command
+
+```
+$ flatpak install -y flathub org.gtk.Gtk3theme.Plata-Noir
+```
+
+You can search for additional Gtk themes with the command
```
flexibeast, line 68: "GTK"

src/config/external-applications.md
```diff
@@ -0,0 +1,83 @@
...
+
+```
+$ flatpak search org.gtk.Gtk3theme
+```
+
+### Troubleshooting
+
+Some Flatpaks require D-Bus, enable the `dbus` service with runit. They can also
```
flexibeast, line 76: Do they require a system bus, or a session bus, or both? The `dbus` service provides the system bus, but if a Flatpak requires only a session bus, enabling the `dbus` service will not be necessary, and a session bus will need to be made available instead. (For example, i haven't enabled the `dbus-service`, but start i3 via `dbus-run-session`.)

src/config/external-applications.md
```diff
@@ -0,0 +1,83 @@
...
+require Pulseaudio, see [Pulseaudio](./media/pulseaudio.md).
```
flexibeast, line 77: "require \[PulseAudio\](./media/pulseaudio.md)."

flexibeast:
@ericonr: Done. :-)

src/config/external-applications.md
anjandev: I am not sure about this. I will change it to the starting `dbus-run-session`. Maybe starting the systemwide dbus is unneeded.

src/config/external-applications.md
anjandev: I think Im going to take this section out. I have been using flatpaks for the last couple days and I cannot find an application that is not following my system's theme.

If this is an issue, people can refer to the official documentation.

src/config/external-applications.md
flexibeast: Well, it depends on the graphical session one is using. `dbus-run-session` won't be needed in the GNOME context, for example, because it will have created a session bus via other means. So we can't simply specify use of `dbus-run-session` here.

src/config/external-applications.md
```diff
@@ -0,0 +1,66 @@
...
+Flatpak is another method for installing external proprietary applications on
+Linux. For information on using Flatpak with Void Linux, see the [official
+Flatpak documentation](https://flatpak.org/setup/Void%20Linux/).
+
+> Flatpak's sandboxing will not necessarily protect you from any security and/or
+> privacy-violating features of proprietary software.
+
+In some instances, this may be fixed by installing one or more of the
+`xdg-desktop-portal`, `xdg-desktop-portal-gtk`, `xdg-user-dirs`,
+`xdg-user-dirs-gtk` or `xdg-utils` packages.
```
ericonr, line 54: It's not clear what "this" is. You might have deleted the context :B 

I think switching "this" for "issues" should be enough.

ericonr: As a matter of fact, whynot include this paragraph in troubleshooting?

src/config/external-applications.md
```diff
@@ -0,0 +1,66 @@
...
+
+### Troubleshooting
+
+Some Flatpaks require D-Bus. If you are using a window manager (e.g. i3wm) add
+`dbus-run-session` to your window manager's launch. They can also require
+[Pulseaudio](./media/pulseaudio.md).
```
ericonr, line 60: I don't really like this phrasing... Something like "you can launch your window manager with dbus-run-session" feels much cleaner. If you could link its man page, that'd be even better. https://man.voidlinux.org/dbus-run-session.1

ericonr:
Tight! I think we can merge this one :D 

Thanks a lot!

@flexibeast can you do the honours? I need to sleep and don't trust myself fully rn :p

flexibeast:
@anjandev, @ericonr: Thank you both for your work on this! i just made one small change before merging, so as to not imply that all Flatpak issues can be resolved by installing `xdg-*` packages. :-)

PR #189, opened 2020-04-13
graphical-session: create section and update subsections

ericonr:
* Remove ConsoleKit2 references

* Make it about graphical sessions as a whole, session management is
needed for both Xorg and Wayland

I need some help for this one, though.

In the session management stuff, should D-Bus be listed? I feel like somewhere we should properly explain it, but idk if it would be correct to include in the session management part.

Should the session management stuff even be included inside graphical session? It's pretty much required for proper graphical stuff, but it also has uses outside that.

At least under Intel, Xorg tends to prefer the modesetting driver, and that is the only driver that works for Nvidia PRIME as well. Does anyone know of a good way to include it in the Xorg page? I don't know exactly how to go bout explaining the whole thing, but I think it's important to mention that sometimes the xf86 video drivers aren't really necessary.

Related to #49


src/SUMMARY.md
```diff
@@ -40,12 +40,13 @@
...
       - [NetworkManager](./config/network/networkmanager.md)
       - [ConnMan](./config/network/connman.md)
-   - [Graphics](./config/graphics/index.md)
-      - [Intel GPU](./config/graphics/intel.md)
-      - [NVIDIA Optimus](./config/graphics/optimus.md)
-   - [Xorg](./config/xorg/index.md)
-      - [Session Management](./config/xorg/session-management.md)
-   - [Wayland](./config/wayland.md)
+   - [Graphical Session](./config/graphical-session/index.md)
+      - [Graphics](./config/graphical-session/graphics/index.md)
```
bobertlo, line 44: This could probably be 'Graphics Drivers' at this point

src/config/graphical-session/session-management.md
```diff
@@ -0,0 +1,18 @@
...
+
+## elogind
+
+[elogind(8)](https://man.voidlinux.org/elogind.8) is a standalone version of
+`systemd-logind`, a service to manage user logins. This service provides
+necessary features for most desktop environments and Wayland compositors. To
+make use of its features, install the `elogind` package:
+
+```
+# xbps-install -S elogind
```
bobertlo, line 15: As mentioned in #190 we are, in fact, avoiding sharing this command every time. :)

ericonr: In my defense, it's legacy documentation :p

src/SUMMARY.md
ericonr: Agreed

src/config/graphical-session/session-management.md
ericonr: I removed the install command, but didn't like the package installation page. Do you think that's necessary?

ericonr:
@bobertlo What do you think about this fix? The Wayland page mentions the session management page, so if someone needs it they will be directed to it.

ericonr:
Removed `xbps-install` stuff.

bobertlo:
Thanks!

PR #192, opened 2020-04-14
 wpa_supplicant: document new DRIVER option 

matthias-t:
https://github.com/void-linux/void-packages/pull/20972


matthias-t:
If you pass only `-D <driver>` to `OPTS`, then no interface or configuration file is specified, as `OPTS` overrides all other options. This confused me when setting up `wpa_supplicant`. But it's true that one can set `-i` and `-c` there as well.

bobertlo:
Oh, missed that. Fair enough!

bobertlo:
d98fabccb4446b924680eda82c855ec4bba8afe2

bobertlo:
Thanks. Please be sure to squash commits in the future :)

matthias-t:
Okay, sure! I'll try to get the formatting right, too

PR #195, opened 2020-04-17
config/wayland: clarify how to set up wayland without elogind

travankor:
Describes how to run wayland without elogind.

Information is pulled from https://wayland.freedesktop.org/building.html. Note this doesn't follow the freedesktop spec verbatim, but it allows you to run a Wayland session without elogind or dbus.

### DIFF
--- a/src/config/graphical-session/wayland.md
+++ b/src/config/graphical-session/wayland.md
@@ -85,9 +85,14 @@ compositors, and is installed as a dependency for most of them. Its package is
 ## Configuration
 
 The Wayland API uses the `XDG_RUNTIME_DIR` environment variable to determine the
-directory for the Wayland socket. In order to securely set this variable, you
-need the `elogind` service installed and enabled as your [session
-manager](./session-management.md).
+directory for the Wayland socket.
+
+Install `elogind` as your [session manager](./session-management.md) to
+automatically setup `XDG_RUNTIME_DIR`.
+
+Alternatively, manually set the environment variable through the shell. Make
+sure to create a dedicated user directory and set its permissions to `700`. A
+good default location is `/run/user/$(id -u)`.
 
 It is also possible that some applications use the `XDG_SESSION_TYPE`
 environment variable in some way, which requires that you set it to `wayland`.


CameronNemo:
Isn't `pam_rundir` functional?

travankor:
I think you can use it (or the shell snippet in the link), but is there a security/other issue that creating/destroying the directory addresses? XDG also specifies that you mount this on tmpfs, but for wayland this seems unnecessary.

src/config/graphical-session/wayland.md
```diff
@@ -85,9 +85,12 @@ compositors, and is installed as a dependency for most of them. Its package is
 ## Configuration
 
 The Wayland API uses the `XDG_RUNTIME_DIR` environment variable to determine the
-directory for the Wayland socket. In order to securely set this variable, you
-need the `elogind` service installed and enabled as your [session
-manager](./session-management.md).
+directory for the Wayland socket. You can either install `elogind` as your
+[session manager](./session-management.md), or you can manually set the
+environment variable.
```
ericonr, line 90: Throwing elogind here feels weird, making it clear that you can "install elogind as your session manager, which takes care of it, or..." seems better.

src/config/graphical-session/wayland.md
```diff
@@ -85,9 +85,12 @@ compositors, and is installed as a dependency for most of them. Its package is
...
+
+If you set the value manually, make sure to create a directory and set
+permissions of `700`. A good default location is `/run/user/$(id -u)`.
```
ericonr, line 93: "set its permissions to `700`"

src/config/graphical-session/wayland.md
```diff
@@ -85,9 +85,14 @@ compositors, and is installed as a dependency for most of them. Its package is
 ## Configuration
 
 The Wayland API uses the `XDG_RUNTIME_DIR` environment variable to determine the
-directory for the Wayland socket. In order to securely set this variable, you
-need the `elogind` service installed and enabled as your [session
-manager](./session-management.md).
+directory for the Wayland socket.
+
+Install `elogind` as your [session manager](./session-management.md) to
+automagically setup `XDG_RUNTIME_DIR`.
```
ericonr, line 91: No matter how much I like that word, "automagically" is a bit too informal, unfortunately. Could use "automatically", instead.

travankor: technically this is part of UNIX/computer culture :p
http://www.catb.org/jargon/html/A/automagically.html

nevertheless, I changed it

ericonr:
So, it seems ready to merge. However, given that elogind should become a dependency for the relevant compositors, we might have to change this, when that PR's been merged. For now, I will just wait for CI.

PR #197, opened 2020-04-18
Fix inconsistencies in style

### LINKED ISSUE #190: Is a consistent style desired? Should documentation be more terse?
So, usually package installation is presented as the whole `xbps-install` command, but something that the Arch wiki actually does right IMHO is just list the packages that people have to install and hyperlink the word "install" to the guide about installing packages. That feels more concise, at least to me.

This also goes for enabling services. There are instructions for enabling dbus, for example, all around the docs. Wouldn't it be better to just mention "make sure that the dbus service is enabled" and link to the services page as well? That's what the NetworkManager page does, while the IWD and Bluetooth page both have the `ln -s /etc/sv/dbus /var/service` command.

This is not a really big issue, but it would be interesting if we could define some sort of guidelines around writing docs and keep a consistent style. Whatever decision is made, I can port the current pages to that (or not, if it's deemed unnecessary).

ericonr:
Fixes #190 

I might have removed too much stuff, please let me know if that's the case. If I have removed too little I'd like to know as well.

There have also been some additions, to allow for even more removal of duplication.

### DIFF
--- a/src/config/bluetooth.md
+++ b/src/config/bluetooth.md
@@ -16,27 +16,10 @@ ID TYPE     DEVICE      SOFT      HARD
 
 ## Installation
 
-Install the `bluez` package.
-
-```
-# xbps-install -S bluez
-```
-
-Enable the `bluetoothd` and `dbus` services.
-
-```
-# ln -s /etc/sv/dbus /var/service/
-# ln -s /etc/sv/bluetoothd /var/service/
-```
-
-Add your user to the `bluetooth` group and restart the `dbus` service, or simply
-reboot the system. Note that restarting the `dbus` service may kill processes
-making use of it.
-
-```
-# usermod -a -G bluetooth $USER
-# sv restart dbus
-```
+Install the `bluez` package and enable the `bluetoothd` and `dbus` services.
+Then, add your user to the `bluetooth` group and restart the `dbus` service, or
+simply reboot the system. Note that restarting the `dbus` service may kill
+processes making use of it.
 
 > Note: To use an audio device such as a wireless speaker or headset, ALSA users
 > need to install the `bluez-alsa` package, while
@@ -44,42 +27,11 @@ making use of it.
 
 ## Usage
 
-Manage Bluetooth connections and controllers using `bluetoothctl`. It uses a
-command line interface; to find out what commands are available, enter `help`.
-To exit, enter `exit` or `quit`.
-
-```
-$ bluetoothctl
-
-[bluetooth]# help
-Menu main:
-Available commands:
--------------------
-advertise             Advertise Options Submenu
-scan                  Scan Options Submenu
-gatt                  Generic Attribute Submenu
-list                  List available controllers
-show [ctrl]           Controller information
-select <ctrl>         Select default controller
-devices               List available devices
-paired-devices        List paired devices
-...
-```
+Manage Bluetooth connections and controllers using `bluetoothctl`, which
+provides a command line interface and also accepts commands on standard input.
 
 Consult the [Arch Wiki](https://wiki.archlinux.org/index.php/Bluetooth#Pairing)
-page for an example on how to pair a device.
-
-`bluetoothctl` also accepts taking commands from stdin.
-
-```
-$ echo "devices" | bluetoothctl
-Agent registered
-[bluetooth]# devices
-Device XX:XX:XX:XX:XX:XX Mouse
-Device XX:XX:XX:XX:XX:XX Keyboard
-Device XX:XX:XX:XX:XX:XX Speaker
-Device XX:XX:XX:XX:XX:XX Headset
-```
+for an example of how to pair a device.
 
 ## Configuration
 
--- a/src/config/kernel.md
+++ b/src/config/kernel.md
@@ -2,17 +2,12 @@
 
 ## Kernel series
 
-Void Linux provides many kernel series in the default repository,
+Void Linux provides many kernel series in the default repository. These are
+named `linuxX.X`, for example `linux4.19`. You can query for all available
+kernel series by runnning:
 
 ```
 $ xbps-query --regex -Rs '^linux[0-9.]+-[0-9._]+'
-[-] linux3.16-3.16.63_1          The Linux kernel and modules (3.16 series)
-[-] linux3.18-3.18.124_1         The Linux kernel and modules (3.18 series)
-[-] linux4.14-4.14.98_1          The Linux kernel and modules (4.14 series)
-[*] linux4.19-4.19.25_1          The Linux kernel and modules (4.19 series)
-[-] linux4.20-4.20.12_1          The Linux kernel and modules (4.20 series)
-[-] linux4.4-4.4.176_1           The Linux kernel and modules (4.4 series)
-[-] linux4.9-4.9.160_1           The Linux kernel and modules (4.9 series)
 ```
 
 The `linux` meta package which is installed by default depends on one of the
@@ -22,44 +17,25 @@ modules.
 ## Removing old kernels
 
 When updating the kernel, old versions are left behind in case it is necessary
-to roll back to an older version. Over time, many old kernel version accumulate
-and make updating dmks modules take a long time. Thus, it may be advisable to
-clean old kernels from time to time.
+to roll back to an older version. Over time, old kernel versions can accumulate,
+consuming disk space and increasing the time taken by dkms module updates. Thus,
+it may be advisable to clean old kernels from time to time.
 
 Removing old kernels is done using the
-[vkpurge(8)](https://man.voidlinux.org/vkpurge.8) utility. `vkuprge` comes
-pre-installed on every Void Linux system. This utility runs the necessary hooks
-when removing old kernels.
-
-To list kernels that can be purged:
-
-```
-$ vkpurge list
-3.8.2_1
-```
-
-To remove a specific kernel version like `3.8.2_1`:
-
-```
-# vkpurge rm 3.8.2_1
-```
-
-To remove `all` kernels except the latest kernel of each series and the kernel
-that is currently booted:
-
-```
-# vkpurge rm all
-```
+[vkpurge(8)](https://man.voidlinux.org/vkpurge.8) utility. `vkpurge` comes
+pre-installed on every Void Linux system. This utility runs the necessary
+[hooks](#kernel-hooks) when removing old kernels.
 
 ## Kernel modules
 
-Kernel modules are typically drivers for devices or filesystems. Normally the
-kernel automatically loads required modules, but sometimes it may be necessary
-to explicitly load modules at boot.
+Kernel modules are typically drivers for devices or filesystems.
 
-### Loading kernel modules at boot
+### Loading kernel modules during boot
 
-To load kernel modules at boot time, a `.conf` file like
+Normally the kernel automatically loads required modules, but sometimes it may
+be necessary to explicitly specify modules to be loaded during boot.
+
+To load kernel modules during boot, a `.conf` file like
 `/etc/modules-load.d/virtio.conf` needs to be created with the contents:
 
 ```
@@ -69,34 +45,36 @@ virtio-net
 
 ### Blacklisting kernel modules
 
-There are two different methods to blacklist kernel modules, for the initramfs
-and for the booted system. Some modules are loaded by the initramfs very early
-in the boot process, those have to be blacklisted in the initramfs.
+Blacklisting kernel modules is a method for preventing modules from being loaded
+by the kernel. There are two different methods for blacklisting kernel modules,
+one for modules loaded by the initramfs and one for modules loaded after the
+initramfs process is done. Modules loaded by the initramfs have to be
+blacklisted in the initramfs configuration.
 
-To blacklist modules, create a `.conf` file like `/etc/modprobe.d/radeon.conf`
-with the contents:
+To blacklist modules loaded after the initramfs process, create a `.conf` file,
+like `/etc/modprobe.d/radeon.conf`, with the contents:
 
 ```
 blacklist radeon
 ```
 
-#### dracut
+#### Blacklisting modules in the initramfs
 
-To blacklist modules from being included in a dracut initramfs a `.conf` file
-needs to be created like `/etc/dracut.conf.d/radeon.conf` with the contents:
+After making the necessary changes to the configuration files, the initramfs
+needs to be [regenerated](#kernel-hooks) for the changes to take effect on the
+next boot.
+
+##### dracut
+
+Dracut can be configured to not include kernel modules through a configuration
+file. To blacklist modules from being included in a dracut initramfs, create a
+`.conf` file, like `/etc/dracut.conf.d/radeon.conf`, with the contents:
 
 ```
 omit_drivers+=" radeon "
 ```
 
-Now initramfs needs to be regenerated to make the changes take effect on the
-next reboot:
-
-```
-# dracut --force
-```
-
-#### mkinitcpio
+##### mkinitcpio
 
 To blacklist modules from being included in a mkinitcpio initramfs a `.conf`
 file needs to be created like `/etc/modprobe.d/radeon.conf` with the contents:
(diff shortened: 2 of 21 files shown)


src/config/kernel.md
```diff
@@ -71,16 +46,22 @@ virtio-net
 
 There are two different methods to blacklist kernel modules, for the initramfs
 and for the booted system. Some modules are loaded by the initramfs very early
-in the boot process, those have to be blacklisted in the initramfs.
+in the boot process; those have to be blacklisted in the initramfs.
 
-To blacklist modules, create a `.conf` file like `/etc/modprobe.d/radeon.conf`
-with the contents:
+To blacklist modules for the booted system, create a `.conf` file like
```
bobertlo, line 51: This seems a little unclear?

ericonr: Probably, yeah. I will think about how to explain it better.

src/config/locales.md
```diff
@@ -9,26 +9,16 @@ locale -a
...
 To enable the locale `xxxx`, un-comment or add the relevant line in
-`/etc/default/libc-locale.conf` and run
-
-```
-# xbps-reconfigure -f glibc-locales
-```
+`/etc/default/libc-locale.conf` and
+[reconfigure](../xbps/packages/management.md) the `glibc-locales` package.
 
 ## Setting the system locale
```
bobertlo, line 15: This isn't so much related to your changes but looking at this, I think it could be reorganized to 'set' the locale first, and then mention that it must then be enabled.

src/config/kernel.md
```diff
@@ -22,34 +17,14 @@ modules.
...
 Removing old kernels is done using the
 [vkpurge(8)](https://man.voidlinux.org/vkpurge.8) utility. `vkuprge` comes
-pre-installed on every Void Linux system. This utility runs the necessary hooks
-when removing old kernels.
-
-To list kernels that can be purged:
-
-```
-$ vkpurge list
-3.8.2_1
```
bobertlo, line 38: I'm okay with the removal of this content, but this is one place I would like a second opinion.

src/config/kernel.md
ericonr: Just updated most of the text about kernel modules, if you want to take a look.

src/config/kernel.md
ericonr: About this:

Pros:

* I feel like people should be able to open man pages to see what the comments do, instead of relying on just the docs
* If anything changes in vkpurge, it isn't necessary to change the documentation in several places

Cons:

* A doc that just gives you the command is much more welcoming

bobertlo: Yeah I have no strong feelings one way or the other. I think consistency is nice and will help in general. Having this stay is friendly though.

bobertlo:
I'd appreciate a general yes or no on the spirit and scope of these changes from anyone involved with the project. I personally feel they can more or less all be merged. I'll do another more thorough review before merging this if there is support.

src/config/bluetooth.md
```diff
@@ -16,70 +16,22 @@ ID TYPE     DEVICE      SOFT      HARD
...
-gatt                  Generic Attribute Submenu
-list                  List available controllers
-show [ctrl]           Controller information
-select <ctrl>         Select default controller
-devices               List available devices
-paired-devices        List paired devices
-...
-```
+Manage Bluetooth connections and controllers using `bluetoothctl`, which uses a
+command line interface and also accepts commands from stdin.
```
flexibeast, line 31: s/uses/provides/? And should 'stdin' be capitalised? (This appears to be the only place it's used in the documentation at the moment.)

src/config/bluetooth.md
```diff
@@ -16,70 +16,22 @@ ID TYPE     DEVICE      SOFT      HARD
...
-```
-$ echo "devices" | bluetoothctl
-Agent registered
-[bluetooth]# devices
-Device XX:XX:XX:XX:XX:XX Mouse
-Device XX:XX:XX:XX:XX:XX Keyboard
-Device XX:XX:XX:XX:XX:XX Speaker
-Device XX:XX:XX:XX:XX:XX Headset
-```
+page for an example of how to pair a device.
```
flexibeast, line 34: i'd remove "page" here.

src/config/kernel.md
```diff
@@ -22,44 +17,25 @@ modules.
 ## Removing old kernels
 
 When updating the kernel, old versions are left behind in case it is necessary
-to roll back to an older version. Over time, many old kernel version accumulate
-and make updating dmks modules take a long time. Thus, it may be advisable to
-clean old kernels from time to time.
+to roll back to an older version. Over time, many old kernel versions can
+accumulate, taking up a lot of space and making dkms modules updates take a long
+time. Thus, it may be advisable to clean old kernels from time to time.
```
flexibeast, line 22: Suggest: "versions can accumulate, consuming disk space and increasing the time needed to do dkms module updates. Thus, it may be advisable to remove ..."

src/config/kernel.md
flexibeast: i feel the pros listed outweigh the cons, but wouldn't object to the content staying in. Either way, "`vkuprge`" -> "`vkpurge`". :-)

src/config/kernel.md
```diff
@@ -69,34 +45,36 @@ virtio-net
 
 ### Blacklisting kernel modules
 
-There are two different methods to blacklist kernel modules, for the initramfs
-and for the booted system. Some modules are loaded by the initramfs very early
-in the boot process, those have to be blacklisted in the initramfs.
+Blacklisting kernel modules is a method for preventing these modules from being
```
flexibeast, line 48: Suggest removing "these".

src/config/kernel.md
```diff
@@ -69,34 +45,36 @@ virtio-net
...
+loaded by the kernel. There are two different methods for blacklisting kernel
+modules, one for modules loaded by the initramfs and one for modules loaded
+after the initramfs process is done. Modules loaded by the initramfs have to be
+blacklisted in the initramfs configuration.
 
-To blacklist modules, create a `.conf` file like `/etc/modprobe.d/radeon.conf`
-with the contents:
+To blacklist modules that are loaded after the initramfs process, create a
```
flexibeast, line 54: Suggest removing "that are".

src/config/kernel.md
```diff
@@ -105,71 +83,54 @@ file needs to be created like `/etc/modprobe.d/radeon.conf` with the contents:
...
+`gummiboot` and `lilo`.
 
-## Dynamic Kernel Module Support (dkms)
+### Install hooks
 
-There are kernel modules that are not part of the Linux source tree that are
-build at install time using dkms and [kernel hooks](#kernel-hooks).
+The `{pre,post}-install` hooks are executed by
+[xbps-reconfigure(1)](https://man.voidlinux.org/xbps-reconfigure.1) when it's
+used to configure a Linux kernel, such as building its initramfs. This happens
```
flexibeast, line 98: suggest "when configuring a".

src/config/kernel.md
```diff
@@ -105,71 +83,54 @@ file needs to be created like `/etc/modprobe.d/radeon.conf` with the contents:
...
+# xbps-reconfigure --force linuxX.X
 ```
 
+If run manually, they serve to apply initramfs configuration changes to the next
+boot.
+
+### Remove hooks
+
+The `{pre,post}-remove` hooks are executed by
+[vkpurge(8)](https://man.voidlinux.org/vkpurge.8) when it's used to remove old
```
flexibeast, line 112: Suggest "when removing old".

src/config/kernel.md
```diff
@@ -105,71 +83,54 @@ file needs to be created like `/etc/modprobe.d/radeon.conf` with the contents:
...
+available modules can be listed by searching for `dkms` in the package
+repositories.
+
 ## cmdline
 
 ### GRUB
 
-Kernel command line arguments can be added through the grub bootloader by
+Kernel command line arguments can be added through the GRUB bootloader by
 editing `/etc/default/grub` and changing the `GRUB_CMDLINE_LINUX_DEFAULT`
```
flexibeast, line 127: Suggest "`/etc/default/grub`, changing" given the rest of the sentence.

src/config/kernel.md
```diff
@@ -105,71 +83,54 @@ file needs to be created like `/etc/modprobe.d/radeon.conf` with the contents:
...
-variable and then regenerating the grub configuration.
-
-```
-# vi /etc/default/grub
-# grub-mkconfig -o /boot/grub/grub.cfg
-```
+variable and then reconfiguring the `grub` package you have installed.
```
flexibeast, line 128: Suggest removing "you have installed".

src/config/man.md
```diff
@@ -10,11 +10,10 @@ pages.
 $ man 1 chroot
 ```
 
-The [mandoc](http://mandoc.bsd.lv/) toolset contains
-[apropos(1)](https://man.voidlinux.org/apropos.1) to search for manual pages,
-[apropos(1)](https://man.voidlinux.org/apropos.1) uses a database that can be
-updated and generated with the
-[makewhatis(8)](https://man.voidlinux.org/makewhatis.8) command.
+The `mandoc` toolset contains [apropos(1)](https://man.voidlinux.org/apropos.1)
+to search for manual pages. `apropos` uses a database that can be updated and
```
flexibeast, line 14: Suggest ".1), which can be used to search".

src/config/post-install.md
```diff
@@ -6,15 +6,8 @@ machine.
...
-```
+After installing and connecting the network, new systems should be
+[updated](../xbps/updating.md).
 
 This will ensure that your system has applicable security patches and software
 upgrades that happened after the installation media was created.
-
-> Note: XBPS must use a separate transaction to update itself. If your first
-> update includes the package `xbps`, you will need to run an additional update
-> for the rest of the system.
```
flexibeast, line 20: Hmm, not sure about removing this. i'd prefer to keep it (unless this information is mentioned elsewhere), so that people don't end up thinking that the independent `xbps` package update was the only update necessary.

src/config/print/index.md
```diff
@@ -3,20 +3,5 @@
...
-```
-# ln -s /etc/sv/cupsd /var/service/
-```
-
-Wait until the service is marked available:
-
-```
-# sv status cupsd
-```
+As pre-requisites, install the `cups` package and enable the `cupsd` service.
```
flexibeast, line 6: Suggest "prerequisites", on the basis of [the Wiktionary entry](https://en.wiktionary.org/wiki/prerequisite).

src/config/security/hashboot.md
```diff
@@ -1,21 +1,18 @@
...
 
 ## Installation
 
-```
-# xbps-install -S hashboot
-```
-
-To verify BIOS, `flashrom` needs to be installed, too.
+Install the `hashboot` package. To verify the BIOS, `flashrom` needs to be
+installed, too.
```
flexibeast, line 15: Suggest "installed as well".

src/xbps/repositories/official/installing.md
```diff
@@ -1,12 +1,8 @@
 # Installing Repositories
 
 Void provides official [repositories](./index.md) that may be installed via a
-package. For example, to install the `nonfree` repository, install the package
```
flexibeast, line 4: Suggest "that may be enabled via the installation of the relevant package. For example, to enable the".

src/config/kernel.md
ericonr: fiexd! :p

src/config/bluetooth.md
ericonr: Fixed both. I capitalised to STDIN, do you think we should make it `STDIN`?

src/config/bluetooth.md
ericonr: removed

src/xbps/repositories/official/installing.md
ericonr: Good point. Fixed.

src/config/print/index.md
ericonr: You can probably base that on the entire internet, I'm not sure where that spelling came from.

Fixed!

src/config/security/hashboot.md
ericonr: Definitely. Fixed.

src/config/post-install.md
ericonr: I removed this one because it's exactly the same one that is in the upgrading page.

src/config/man.md
ericonr: Much better. Fixed.

src/config/kernel.md
ericonr: I changed it to "and increasing the time taken by dmks module updates"

src/config/kernel.md
ericonr: Yup

src/config/kernel.md
ericonr: Yup

src/config/kernel.md
ericonr: Yup

src/config/kernel.md
ericonr: I don't use GRUB, so I didn't know where the kernel hooks came from. Seems it's from the `grub` package, not the specific ones, like `grub-x86_64-efi`.

Fixed!

src/config/post-install.md
flexibeast: Ah okay, fair enough.

src/config/bluetooth.md
flexibeast: Well, i've been looking into how it's written elsewhere, and it looks like lower-case might actually be the way to go: cf. e.g. [this](https://www.gnu.org/software/libc/manual/html_node/Standard-Streams.html) and [this](https://www.gnu.org/software/libc/manual/html_node/Standard-Streams.html). So my Perliness might be showing. :-) i'm slightly inclined towards `stdin` rather than 'stdin', but would be interested to know what others think.

ericonr: I think you pasted the same link twice.

I searched for it a bit, as well. I found STDIN, **STDIN**, stdin and `stdin`. The file descriptors in C are a good reference, and if it comes from code I'd argue for `stdin`.

At least where I could find, the Arch Wiki uses stdin.

flexibeast: Gah, sorry! The first link was supposed to be https://pubs.opengroup.org/onlinepubs/009695399/functions/stdin.html.

ericonr: No problem! Hmm... They seem to just use *stdin*, which I wouldn't say is the worst choice available to us.

q66: I'd write "also accepts commands on standard input"

ericonr: Yup, changed it! Thanks :)

bobertlo:
Noone has objected so I think we can go ahead and merge this. Would you mind rebasing over the locale changes? If not I can when I have time. I'll do another careful review since this is pretty big but otherwise I think we are good to merge.

bobertlo:
And I'd like to merge this before anything else changes!

ericonr:
@bobertlo rebased ;)

ericonr:
@bobertlo fixed formatting as well :p, I had forgotten to run `format.sh`.

src/xbps/repositories/official/installing.md
```diff
@@ -1,12 +1,8 @@
...
 
-Void provides official [repositories](./index.md) that may be installed via a
-package. For example, to install the `nonfree` repository, install the package
-`void-repo-nonfree`:
-
-```
-# xbps-install -S void-repo-nonfree
-```
+Void provides official [repositories](./index.md) that may be enabled via the
+installationof the relevant package. For example, to enable the `nonfree`
```
bobertlo, line 4: installation of

src/xbps/packages/management.md
```diff
@@ -4,11 +4,15 @@ Most general package management is done with the following commands:
 
 - [xbps-query(1)](https://man.voidlinux.org/xbps-query.1) displays information
    about packages installed locally, or, with the `-R` flag, packages contained
-   in repositories.
+   in repositories. It can also be used to search for packages.
```
bobertlo, line 7: I don't feel strongly but maybe we could say that `xbps-query` searches for and displays information...?

ericonr: Feels more concise. I think it would be better.

ericonr: I made another small change in this paragraph, to make the part about the `-R` flag clearer.

PR #198, opened 2020-04-18
rc-files: remove reference to TZ

ericonr:
The preferred way of setting localtime is by symlinking it to
/etc/localtime.

Related to https://github.com/void-linux/void-runit/pull/40

### DIFF
--- a/src/config/rc-files.md
+++ b/src/config/rc-files.md
@@ -21,15 +21,6 @@ KEYMAP=fr
 For further details, refer to
 [loadkeys(1)](https://man.voidlinux.org/loadkeys.1).
 
-### TIMEZONE
-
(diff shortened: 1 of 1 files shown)


bobertlo:
This seems okay to me. Maybe it would be worth mentioning but saying the symlink is preferred? I don't feel strongly either way on that.

Does anyone else have opinions on this?

ericonr:
Duncaen recommended removing it on IRC.

It follows the style of the recommendation for hostname, which isn't in the docs page but it's a note in `/etc/rc.conf`.

bobertlo:
Ok I'm merging this. I already had mixed feelings about documenting variables in a configuration file that is already documented internally anyways.

PR #203, opened 2020-04-20
feat(config/services/logging): mention svlogtail

CameronNemo:
This is the only detail from the INSTALL.msg that was missing.


ericonr:
Thanks!

PR #204, opened 2020-04-20
feat(config/security/apparmor): add section

CameronNemo:
It is important to outline how to enable apparmor,
because there are a few steps and one is unique to Void.

### DIFF
--- a/src/config/security/apparmor.md
+++ b/src/config/security/apparmor.md
@@ -1,0 +1,18 @@
+# AppArmor
+
+AppArmor is a mandatory access control mechanism (like SELinux). It can
+constrain programs based on pre-defined or generated policy definitions.
+
+Void ships with some default profiles for several services, such as `dhcpcd` and
+`wpa_supplicant`. Container runtimes such as LXC and podman integrate with
+AppArmor for better security for container payloads.
+
+To use AppArmor on a system, one must:
+
+1. Install the `apparmor` package.
+2. Set the `APPARMOR` variable in `/etc/default/apparmor` to `enforce` or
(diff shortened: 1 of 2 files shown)


ericonr:
What do you think about linking the kernel command line part to the relevant `kernel.md` section?

Is the additional void step the `/etc/default/apparmor` part?

bobertlo:
This does seem appropriate for the docs. Looks good to me, but I don't have any experience with this.

CameronNemo:
>What do you think about linking the kernel command line part to the relevant kernel.md section?

Done.

>Is the additional void step the /etc/default/apparmor part?

yep. I wish we never introduced that (apparmor already has a mechanism to disable it entirely as well as run specific profiles in complain mode), but it was before my time.

ericonr:
I don't understand much about AppArmor. Do you think that could be removed without thoroughly screwing  with users' setups?

src/config/security/apparmor.md
```diff
@@ -0,0 +1,18 @@
+# AppArmor
+
+AppArmor is a manadatory access control mechanism (like SELinux). It can
+constrain programs based on pre-defined or generated policy definitions.
```
ericonr, line 4: manadatory -> mandatory

CameronNemo: done

CameronNemo:
>Do you think that could be removed without thoroughly screwing with users' setups?

The complain setting has no equivalent in the upstream scripts. So it may be a little challenging.

ericonr:
Got it. I will merge it, then. Thanks!

PR #206, opened 2020-04-20
ZFS & USB Advanced Install Guides

aneonlight:
Changes:

- Created a ZFS on Root Install guide
- Created a Live USB to Full Install guide
- Updated SUMMARY.md to reflect the new pages
- Added guides to the list guides/index.md

Things to look at:

- Grammar and cohesion
- Anything UEFI (I've tested both guides, but don't have a UEFI system)
- Formatting

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,5 +1,8 @@
-For information about how to contribute to the Void documentation, please read the following sections of the Void Handbook:
+For information about how to contribute to the Void documentation, please read
+the following sections of the Void Handbook:
 
 - [About this Handbook](https://docs.voidlinux.org/about-handbook/index.html)
-- [Style Guide](https://docs.voidlinux.org/contributing/void-docs/style-guide.html)
-- [Submitting Changes](https://docs.voidlinux.org/contributing/void-docs/submitting.html)
+- [Style
+   Guide](https://docs.voidlinux.org/contributing/void-docs/style-guide.html)
+- [Submitting
+   Changes](https://docs.voidlinux.org/contributing/void-docs/submitting.html)
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,8 +1,5 @@
-For information about how to contribute to the Void documentation, please read
-the following sections of the Void Handbook:
+For information about how to contribute to the Void documentation, please read the following sections of the Void Handbook:
 
 - [About this Handbook](https://docs.voidlinux.org/about-handbook/index.html)
-- [Style
-   Guide](https://docs.voidlinux.org/contributing/void-docs/style-guide.html)
-- [Submitting
-   Changes](https://docs.voidlinux.org/contributing/void-docs/submitting.html)
+- [Style Guide](https://docs.voidlinux.org/contribuing/void-docs/style-guide.html)
+- [Submitting Changes](https://docs.voidlinux.org/contributing/void-docs/submitting.html)
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,5 +1,5 @@
 For information about how to contribute to the Void documentation, please read the following sections of the Void Handbook:
 
 - [About this Handbook](https://docs.voidlinux.org/about-handbook/index.html)
-- [Style Guide](https://docs.voidlinux.org/contribuing/void-docs/style-guide.html)
+- [Style Guide](https://docs.voidlinux.org/contributing/void-docs/style-guide.html)
 - [Submitting Changes](https://docs.voidlinux.org/contributing/void-docs/submitting.html)
--- a/src/config/locales.md
+++ b/src/config/locales.md
@@ -8,6 +8,7 @@ $ locale -a
 
 ## Enabling locales
 
+<<<<<<< HEAD
 To enable a certain locale, un-comment or add the relevant lines in
 `/etc/default/libc-locales` and [reconfigure](../xbps/index.md) the
 `glibc-locales` package.
--- a/src/config/locales.md
+++ b/src/config/locales.md
@@ -8,7 +8,6 @@ $ locale -a
 
 ## Enabling locales
 
-<<<<<<< HEAD
 To enable a certain locale, un-comment or add the relevant lines in
 `/etc/default/libc-locales` and [reconfigure](../xbps/index.md) the
 `glibc-locales` package.
--- a/src/installation/guides/fullusb.md
+++ b/src/installation/guides/fullusb.md
@@ -1,8 +1,8 @@
-# Live Install To Full Install
+# Live USB To Full Install
 
-This installation guide is to convert a Live USB Void Install into a Full
-Pervasive Void Install. It's recommended to use a machive with at least 4GB of
-RAM, and a USB with at least 4GB of storage.
+This installation guide converts a Live USB of Void Linux into a Full Pervasive
+Void Install. It's recommended to use a machine with at least 4GB of RAM, and a
+USB with at least 4GB of storage.
 
 > **Note:**
 > 
@@ -54,6 +54,11 @@ partitioning tool:
 
 `# cfdisk -z /dev/sdc`
 
+> **Important:**
+> 
+> *Before creating the partition tables it's essential to remove the existing
+> GRUB install, this can be done using one of the following commands.*
+> 
 > *For BIOS/MBR use:*
 > 
 > `# dd if=/dev/zero of=/dev/sdc seek=1 count=2047`
@@ -131,8 +136,6 @@ machine:
 
 `(chroot)# echo voidhost >/etc/hostname`
 
-> **Optional:**
-
 Install GRUB onto the USB stick specifying the path to the USB:
 
 ```
@@ -164,6 +167,12 @@ done
 linux5.4: configured successfully.
 ```
 
+## Configuration
+
+For a general system configuration please consult the [Date and
+Time](./../../config/date-time.md), [Rc Files](./../../config/rc-files.md), and
+[Locales](./../../config/locales.md) pages of the handbook.
+
 ## Cleanup
 
 Exit chroot:
@@ -182,20 +191,3 @@ Restart the system using [reboot(8)](https://man.voidlinux.org/reboot) and log
 in:
 
 `# reboot`
-
-## Configuration
-
-Configure the defaults in the new system's `/etc/rc.conf`. Below are some values
-most system's will want to set.
-
-`````````
-# see loadkeys(8)
-KEYMAP="us"
-# available timezones at /usr/share/zoneinfo
-TIMEZONE="America/Chicago"
-# set to UTC or localtime
-HARDWARECLOCK="UTC"
-`````````
-
-Finally configure the system's locales consulting the
-[Locales](./../../config/locales.md) page of the handbook.
--- a/src/installation/guides/fullusb.md
+++ b/src/installation/guides/fullusb.md
@@ -1,17 +1,19 @@
 # Live USB To Full Install
 
 This installation guide shows how to convert a Live USB of Void Linux into a
-Full Pervasive Install. It's recommended to use a machine with at least 4GB of
-RAM, and a USB with at least 4GB of storage to perform the installation.
+Full, Pervasive Install. It is recommended to use a machine that has at least
+4GB of RAM, and a USB drive with more than 4GB of storage to perform the
+installation.
 
 ## Setup
 
-To begin [download](./../downloading.md) a Void Live ISO and [write it to a
-USB](./../live-images/prep.md).
+To begin, [download](./../downloading.md) a Void Live ISO and [write it to a USB
+drive](./../live-images/prep.md).
 
-To modify the same device that the Operating System was booted from the root
-filesystem must first be placed elsewhere, in this case the RAM. This can be
-done from the Live USB by choosing the RAM option when prompted at boot:
+In order to modify the device that the operating system was booted from, the
+root filesystem must first be placed elsewhere. In this case, it's placed into
+the machine's RAM. This can be done from the Live USB by choosing the RAM option
+when prompted at boot:
 
 ```
 Void Linux
@@ -19,17 +21,16 @@ Void Linux
 Boot First HD Found by BIOS
 ```
 
-And un-mounting the remaining connection between the USB and the system:
+And then un-mounting the remaining connection between the USB and the system:
 
 `# umount /run/initramfs/live`
 
-Then locate the USB using [fdisk(8)](https://man.voidlinux.org/fdisk.8) or
-[lsblk(8)](https://man.voidlinux.org/lsblk.8):
+Then locate the USB using [fdisk(8)](https://man.voidlinux.org/fdisk.8), with
+the `-l` argument, or [lsblk(8)](https://man.voidlinux.org/lsblk.8):
 
-`# fdisk -l`
-
-**Important:** Before creating the partition tables it's essential to remove the
-existing GRUB install, this can be done using one of the following commands.
+**Important:** Before creating the partition tables it is essential to remove
+the existing GRUB install, which can be done using one of the following
+commands.
 
 For BIOS/MBR use:
 
@@ -42,4 +43,4 @@ For BIOS/GPT use:
 ## Installation
 
 To continue the installation, follow the [chroot install guide](./), which
-continues to detail the rest of the installation process.
+details the rest of the installation process.
--- a/src/installation/guides/fullusb.md
+++ b/src/installation/guides/fullusb.md
@@ -2,14 +2,14 @@
 
 This installation guide shows how to convert a Live USB of Void Linux into a
 Full Pervasive Install. It's recommended to use a machine with at least 4GB of
-RAM, and a USB with at least 4GB of storage.
+RAM, and a USB with at least 4GB of storage to perform the installation.
 
 ## Setup
 
-If it has not been done so already, [download](./../live-images/downloading.md)
-a Void Live iso and [write it to a USB](./../live-images/prep.md).
+To begin [download](./../live-images/downloading.md) a Void Live ISO and [write it to a USB](./../live-images/prep.md).
 
-To begin, boot into the RAM option when prompted in GRUB.
+To modify the same device that the Operating System was booted from the root filesystem must first be placed elsewhere, in this case the RAM.
+This can be done from the Live USB by choosing the RAM option when prompted at boot:
 
 ```
 Void Linux
@@ -17,16 +17,12 @@ Void Linux
 Boot First HD Found by BIOS
 ```
 
-Then un-mount the USB from the system, which allows the machine to continue
-running even without the media (as root has been placed entirely in RAM), which
-in turn allows for modifications to be made to the USB.
+And un-mounting the remaining connection between the USB and the system:
 
-```
-# umount /run/initramfs/live
-```
+`# umount /run/initramfs/live`
 
-Locate the USB using [fdisk(8)](https://man.voidlinux.org/fdisk) or
-[lsblk(8)](https://man .voidlinux.org/lsblk):
+Then locate the USB using [fdisk(8)](https://man.voidlinux.org/fdisk) or
+[lsblk(8)](https://man.voidlinux.org/lsblk):
 
 `# fdisk -l`
 
@@ -45,5 +41,5 @@ Locate the USB using [fdisk(8)](https://man.voidlinux.org/fdisk) or
 
 ## Installation
 
-To continue, follow the [chroot installation guide](), which details the rest of
+To continue the installation, follow the [chroot install guide](./), which continues to detail the rest of
 the installation process.
--- a/src/installation/guides/fullusb.md
+++ b/src/installation/guides/fullusb.md
@@ -1,14 +1,8 @@
 # Live USB To Full Install
 
-This installation guide converts a Live USB of Void Linux into a Full Pervasive
-Void Install. It's recommended to use a machine with at least 4GB of RAM, and a
-USB with at least 4GB of storage.
-
-> **Note:**
-> 
-> All commands in this guide are utilizing the
-> [BASH(1)](https://man.voidlinux.org/bash) shell. Other shells may be used, but
-> results may vary.
+This installation guide shows how to convert a Live USB of Void Linux into a
+Full Pervasive Install. It's recommended to use a machine with at least 4GB of
+RAM, and a USB with at least 4GB of storage.
 
 ## Setup
 
@@ -23,36 +17,18 @@ Void Linux
 Boot First HD Found by BIOS
 ```
 
-Then un-mount the USB from the system, leaving the entire root filesystem in
-RAM, which allows it to continue running even when the media is disconnected:
+Then un-mount the USB from the system, which allows the machine to continue
+running even without the media (as root has been placed entirely in RAM), which
+in turn allows for modifications to be made to the USB.
 
 ```
 # umount /run/initramfs/live
 ```
 
-Locate the USB using [fdisk(8)](https://man.voidlinux.org/fdisk):
+Locate the USB using [fdisk(8)](https://man.voidlinux.org/fdisk) or
+[lsblk(8)](https://man .voidlinux.org/lsblk):
 
-```
-# fdisk -l
-Disk /dev/sdc: 28.66 GiB, 30752636928 bytes, 60063744 sectors
-Disk model: Ultra           
-Units: sectors of 1 * 512 = 512 bytes
-Sector size (logical/physical): 512 bytes / 512 bytes
-I/O size (minimum/optimal): 512 bytes / 512 bytes
-Disklabel type: gpt
-Disk identifier: 20E776A8-6DB3-6449-B740-0E080C28A890
-
-Device     Start      End  Sectors  Size Type
-/dev/sdc2   2048 60063710 60057567 28.7G Linux filesystem
-```
-
-## Partitioning
-
-Prepare the USB for installation by creating one of the following partition
-schemes using [cfdisk(8)](https://man.voidlinux.org/cfdisk) or another
-partitioning tool:
-
-`# cfdisk -z /dev/sdc`
+`# fdisk -l`
 
 > **Important:**
 > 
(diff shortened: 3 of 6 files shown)


Vaelatern:
oh my, a zfs on root install guide.

src/installation/guides/zfs.md
```diff
@@ -0,0 +1,330 @@
...
+Installation finished. No error reported.
+```
+
+Create and modify `/etc/dracut.conf.d/zol.conf` to configure Dracut to include
+at least the following values:
+
+```
+hostonly="yes"
+nofsck="yes"
+add_dracutmodules+=" zfs "
```
Vaelatern, line 226: Can modern grub find `/boot` on zfs? Last I checked you needed an older pool version for this to work, have they fixed that?

src/installation/guides/zfs.md
```diff
@@ -0,0 +1,330 @@
...
+# see loadkeys(8)
+KEYMAP="us"
+# available timezones at /usr/share/zoneinfo
+TIMEZONE="America/Chicago"
+# set to UTC or localtime
+HARDWARECLOCK="UTC"
+````````
+
+Finally configure the system's locales consulting the
+[Locales](./../../config/locales.md) page of the handbook.
```
Vaelatern, line 330: Configuration should probably happen before cleanup, and should include setting the timezone with an `ln` instead of the `TIMEZONE` variable. I'd rather not duplicate rc.conf's comments here at all, there is a separate doc page for that.

src/installation/guides/zfs.md
```diff
@@ -0,0 +1,330 @@
+# ZFS On Root
+
+This installation guide is assuming that an already existing Full Void Linux
+Install is available. Not just a Live USB.
+
+If an existing Void Install doesn't exist please consult the [Live USB To Full
+Install](./fullusb.md) guide or the [General
+Instalation](./../live-images/index.md) guide. Then continue along with this
+guide.
```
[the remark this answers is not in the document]
aneonlight: I saw that, but I wasn't sure what the general thought was on recommending void builds not from the website

src/installation/guides/zfs.md
```diff
@@ -0,0 +1,330 @@
...
+Partition    Size       Type
+1             ?G     Solaris Root(bf00)
+```
+
+For a BIOS/GPT system using GRUB:
+
+```
+Partition    Size       Type
+1             2M     BIOS boot    
+2             ?G     Solaris Root
```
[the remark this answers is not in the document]
aneonlight: I'll play around with it and get back on that

flexibeast:
Thanks for your contribution!

The Handbook does have a [style guide](https://docs.voidlinux.org/contributing/void-docs/style-guide.html), but yes, the style is currently uneven due to many of the pages having been ported from the wiki. Work to address this is ongoing (cf. e.g. https://github.com/void-linux/void-docs/pull/197).

i've been using ":" rather than "." before command listings in my recent work, but @bobertlo might have a suggestion as to which convention we should go with.

src/installation/guides/zfs.md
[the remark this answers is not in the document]
Vaelatern: In this case the author is a long-time maintainer on Void Linux and the author of the xtools. You need a full install, or a custom live iso (like hrmpf is).

src/installation/guides/zfs.md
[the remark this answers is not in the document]
aneonlight: It is, GRUB does require a bios partition on GPT and even if no partition table is given ZFS creates one for you.

src/config/locales.md
```diff
@@ -9,7 +9,7 @@ locale -a
 ## Enabling locales
 
 To enable the locale `xxxx`, un-comment or add the relevant line in
-`/etc/default/libc-locale.conf` and run
+`/etc/default/libc-locale` and run
```
bobertlo, line 12: this is embarrasing. but isn't it `libc-locales`?

Also, this should really be in a separate PR.

aneonlight:
Thanks! I went ahead and pulled out all the non-specific sections. 

The descriptions on the table probably needs rewording, but in general I'll try and look over the clarity of the guides as a whole.

ericonr:
I think this should be part of a greater rework around the advanced guides. @bobertlo (I think) had mentioned wanting a "pick your adventure" style guide, where there is a basic guide and several optional steps around that.

We should probably look into merging the chroot guide -> restructuring the FDE guide around that one -> structuring these guides around the same framework. If you could stick around to help with that, I'd love it, but I think this PR, specifically, will have to wait for a bit. If you want to leave it a bit more ready, though, take a look at the CI errors ;)

src/installation/guides/zfs.md
```diff
@@ -0,0 +1,271 @@
...
+
+ZFS pools should use id's to maintain universal functionality across computers
+and to prevent the misrecognition of drives.
+
+Id's can be located in the `/dev/disk/by-id` folder:
+
+> **Note:**
+> 
+> `wwn-` entries and `ata-` entries are referencing equivalent drvies and either
+> can be used.
```
ericonr, line 61: We have made some changes to the style guide, including using quotes only for actual quotes. Notes should be part of the normal text, without any bolding. If you can manage to fit them inside the text flow, even better!

src/installation/guides/zfs.md
```diff
@@ -0,0 +1,271 @@
...
+Prepare the drive for installation by creating one of the following partition
+schemes using [cfdisk(8)](https://man.voidlinux.org/cfdisk) or another
+partitioning tool:
+
+`# cfdisk -z /dev/sda`
+
+> **Warning:**
+> 
+> The disk being partitioned will be formatted and any existing data on the disk
+> will be destroyed.
```
ericonr, line 36: For warnings/important notices, prefer to use:

```
**Warning**: The disk...
```

So, no quotes and all in the same line.

src/installation/guides/zfs.md
ericonr: Also, typo in "drives".

src/installation/guides/zfs.md
```diff
@@ -0,0 +1,271 @@
...
+```
+$ ls -l /dev/disk/by-id
+total 0
+lrwxrwxrwx 1 root root  9 Apr 16 17:34 ata-Hitachi_HDS721010CLA332_JP6911HD2AHS9F -> ../../sda
+lrwxrwxrwx 1 root root 10 Apr 16 17:34 ata-Hitachi_HDS721010CLA332_JP6911HD2AHS9F-part1 -> ../../sda1
+lrwxrwxrwx 1 root root 10 Apr 16 17:34 ata-Hitachi_HDS721010CLA332_JP6911HD2AHS9F-part2 -> ../../sda2
+lrwxrwxrwx 1 root root  9 Apr 16 17:34 wwn-0x5000cca373e0f5d9 -> ../../sda
+lrwxrwxrwx 1 root root 10 Apr 16 17:34 wwn-0x5000cca373e0f5d9-part1 -> ../../sda1
+lrwxrwxrwx 1 root root 10 Apr 16 17:34 wwn-0x5000cca373e0f5d9-part2 -> ../../sda2
+```
```
ericonr, line 72: If you could shorten the output of the command to the smallest possible that still shows what you mean, that would be awesome!

src/installation/guides/zfs.md
```diff
@@ -0,0 +1,271 @@
...
+
+`# zpool create -f -m none zroot dev`
+
+| Command | Action                                            |
+|---------|---------------------------------------------------|
+| -f      | Force the creation of the pool                    |
+| -m none | Set the mountpoint to none                        |
+| zroot   | The name of the pool                              |
+| dev     | The device to be used in the creation of the pool |
+|         | (ie. ata-XXXXX-partX or wwn-XXXX-partX)           |
```
ericonr, line 90: I'm not entirely sure we should use this table, but if it were to stay I would rename the "Command" column to "Argument" and use code blocks for all of them.

src/installation/guides/zfs.md
```diff
@@ -0,0 +1,271 @@
...
+
+To ensure the pool's creation was successful use:
+
+```
+$ zpool list
+NAME    SIZE  ALLOC   FREE  CKPOINT  EXPANDSZ   FRAG    CAP  DEDUP    HEALTH  ALTROOT
+zroot   928G   432K   928G        -         -     0%     0%  1.00x    ONLINE  -
+```
```
ericonr, line 98: In this case, perhaps do something like "To ensure the pool's creation was successful, check the output of `zpool list`."

Do you think it needs an example of the command output?

src/installation/guides/zfs.md
```diff
@@ -0,0 +1,271 @@
...
+
+> **Optional:**
+> 
+> Add compression to the pool using lz4 or another alogorithm, see
+> [zfs(8)](https://man.voidlinux.org/zfs#Native_Properties).
+> 
+> `# zfs set compression lz4 zroot`
```
ericonr, line 105: Compression can be set during pool creation. I think it makes more sense to mention it there, right?

ericonr: I really like the idea of linking to specific sections in man pages, though! Do you think you could use `zfs.8` in the link, though? That way there isn't any ambiguity.

src/installation/guides/zfs.md
```diff
@@ -0,0 +1,271 @@
...
+
+Following the chroot guide, install `grub` and `zfs` from the new system which
+will build the requirements necessary to boot.
+
+`(chroot)# xbps-install -S -R $REPO grub zfs`
+
+## GRUB Installation
+
+Confirm the ZFS modules are loaded in the new system using
+[lsmod(8)](https://man.voidlinux.org/lsmod):
```
ericonr, line 136: Same thing for this command, `lsmod.8` in the link.

src/installation/guides/zfs.md
```diff
@@ -0,0 +1,271 @@
...
+resume
+rootfs-block
+terminfo
+udev-rules
+usrmount
+base
+fs-lib
+shutdown
+========================================================================
+```
```
ericonr, line 245: "Check that `zfs` is in the output of `lsinitrd -m`."

CONTRIBUTING.md
```diff
@@ -1,5 +1,8 @@
...
+For information about how to contribute to the Void documentation, please read
+the following sections of the Void Handbook:
 
 - [About this Handbook](https://docs.voidlinux.org/about-handbook/index.html)
-- [Style Guide](https://docs.voidlinux.org/contributing/void-docs/style-guide.html)
-- [Submitting Changes](https://docs.voidlinux.org/contributing/void-docs/submitting.html)
+- [Style
+   Guide](https://docs.voidlinux.org/contributing/void-docs/style-guide.html)
+- [Submitting
+   Changes](https://docs.voidlinux.org/contributing/void-docs/submitting.html)
```
ericonr, line 8: `vmdfmt` doesn't need to be run for this file, I don't think this needs to be in this PR, right?

src/config/locales.md
```diff
@@ -8,9 +8,12 @@ $ locale -a
...
 
-To enable a certain locale, un-comment or add the relevant lines in
-`/etc/default/libc-locales` and [reconfigure](../xbps/packages/management.md)
-the `glibc-locales` package.
+To enable the locale `xxxx`, un-comment or add the relevant line in
+`/etc/default/libc-locale.conf` and run
+
+```
+# xbps-reconfigure -f glibc-locales
+```
```
ericonr, line 16: I think you might need to rebase your repo to get the latest version of this.

src/installation/guides/zfs.md
aneonlight: It could go either way. For me I like the additional examples as some people following the guide may be less competent with ZFS. However, being an advanced installation guide, it could also be said that the only people following it would already be familiar with ZFS

src/installation/guides/zfs.md
aneonlight: Thanks for catching that

src/installation/guides/zfs.md
aneonlight: You're right. See if where I placed it now makes more sense.

aneonlight:
Hopefully all the discrepancies between this and the upstream branch have been fixed. It'll probably continue failing the CL build for the time being until the Chroot Guide gets pushed through and I can plug in the links.

src/installation/guides/zfs.md
ericonr: Well, I know I did my first (and only) ZFS install blind, so keeping it like this for now is ok c:

src/installation/guides/fullusb.md
```diff
@@ -0,0 +1,45 @@
+# Live USB To Full Install
+
+This installation guide shows how to convert a Live USB of Void Linux into a
+Full Pervasive Install. It's recommended to use a machine with at least 4GB of
+RAM, and a USB with at least 4GB of storage to perform the installation.
```
ericonr, line 5: I prefer not to use contractions (even if sometimes one can go through accidentaly :p). So "It is".

ericonr: "...4GB of RAM and a USB stick with..."

src/installation/guides/fullusb.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Setup
+
+To begin [download](./../downloading.md) a Void Live ISO and [write it to a
+USB](./../live-images/prep.md).
```
ericonr, line 10: "To begin, ..."

ericonr: "USB stick"

src/installation/guides/fullusb.md
```diff
@@ -0,0 +1,45 @@
...
+
+To modify the same device that the Operating System was booted from the root
+filesystem must first be placed elsewhere, in this case the RAM. This can be
+done from the Live USB by choosing the RAM option when prompted at boot:
```
ericonr, line 14: "In order to modify the device that the operating system was booted from, the root filesystem must first be placed elsewhere. In our case, it will be placed in RAM. This can be done ...."

src/installation/guides/fullusb.md
```diff
@@ -0,0 +1,45 @@
...
+
+```
+Void Linux
+**Void Linux(RAM)**
+Boot First HD Found by BIOS
+```
+
+And un-mounting the remaining connection between the USB and the system:
```
ericonr, line 22: "and then un-mounting..."

Since you are continuing the phrase.

src/installation/guides/fullusb.md
```diff
@@ -0,0 +1,45 @@
...
+
+`# umount /run/initramfs/live`
+
+Then locate the USB using [fdisk(8)](https://man.voidlinux.org/fdisk.8) or
+[lsblk(8)](https://man.voidlinux.org/lsblk.8):
```
ericonr, line 27: "using fdisk with the `-l` flag or lsblk" -> that way you avoid having an entire line for the command.

src/installation/guides/fullusb.md
```diff
@@ -0,0 +1,45 @@
...
+
+`# fdisk -l`
+
+**Important:** Before creating the partition tables it's essential to remove the
+existing GRUB install, this can be done using one of the following commands.
```
ericonr, line 32: "it is essential"

ericonr: ", which can be done ..."

src/installation/guides/zfs.md
```diff
@@ -0,0 +1,227 @@
+# ZFS On Root
+
+This installation guide assumes that an already existing Full Void Linux install
+is available. Not just a Live USB.
+
+If an existing Void install does not exist please consult one of the [other
+installation guides](./index.md) or use a custom iso such as
+[hrmpf](https://github.com/leahneukirchen/hrmpf) then continue along with this
+guide.
```
ericonr, line 9: I think these paragraphs can be cut a bit and merged. What do you think?

src/installation/guides/zfs.md
```diff
@@ -0,0 +1,227 @@
...
+
+Create a pool specifying the previously created Solaris Root Partition from the
+disk:
+
+Note: Many users with Advanced Format hard drives will want to also add the `-o
+ahisft=12` argument. This sets the drive's logical sectors to 4k in size.
+Additionally many may want to utilize compression with their pools using either
+lz4 or another alogorithm, see
+[zfs(8)](https://man.voidlinux.org/zfs.8#Native_Properties). This can be done
+after the creation of the pool using: `# zfs set compression lz4 zroot`
```
ericonr, line 74: I'd mention that the user should look into the man page for all the relevant filesystem options, and that most of them can also be set afterwards. No need for an example command like `zfs set compression...`.

I also think this paragraph shouldn't be a Note at all.

Perhaps we should do something like this guide https://github.com/zdykstra/zfsbootmenu/wiki#zfs-pool-creation to give a brief example of flags that they may want to set, all of which should be looked into. The only ones that need to be explained are the ones that you already included, I think.

src/installation/guides/zfs.md
```diff
@@ -0,0 +1,227 @@
...
+
+```
+$ ls -l /dev/disk/by-id
+total 0
+lrwxrwxrwx 1 root root 10 Apr 16 17:34 ata-Hitachi_HDS721010CLA332_JP6911HD2AHS9F-part1 -> ../../sda1
+lrwxrwxrwx 1 root root 10 Apr 16 17:34 wwn-0x5000cca373e0f5d9-part1 -> ../../sda1
+```
+
+Create a pool specifying the previously created Solaris Root Partition from the
+disk:
```
ericonr, line 67: This should perhaps go after the paragraph that's currently a note, so it's directly above the command that it's introducing. Perhaps you can put the Notes after the commands (and the table)?

src/installation/guides/zfs.md
```diff
@@ -0,0 +1,227 @@
...
+Building DKMS module: zfs-0.8.3... done.
+Executing post-install kernel hook: 20-dracut ...
+Executing post-install kernel hook: 50-grub ...
+Generating grub configuration file ...
+Found linux image: /boot/vmlinuz-5.4.32_1
+Found initrd image: /boot/initramfs-5.4.32_1.img
+Found void on /dev/sda2
+done
+linux5.4: configured successfully.
+```
```
ericonr, line 191: I think some of this might fit as a link to `kernel.md#initramfs`.

src/installation/guides/zfs.md
```diff
@@ -0,0 +1,227 @@
...
+# umount /mnt/void/dev/pts
+# umount /mnt/void/{dev,proc,sys}
+# zpool export zroot
+```
+
+Restart the system using [reboot(8)](https://man.voidlinux.org/reboot.8) and log
+in:
+
+`# reboot`
```
ericonr, line 227: In this case the code block for the command can be left out. You can just specify which command they need and that's enough.

src/installation/guides/zfs.md
ericonr: I left some comments on the new version

src/installation/guides/zfs.md
```diff
@@ -0,0 +1,225 @@
...
+```
+Partition    Size       Type
+1             2M     BIOS boot    
+2             ?G     Solaris Root
+```
+
+ZFS pools should use id's to maintain universal functionality across computers
+and to prevent the misrecognition of drives.
+
+Id's can be located in the `/dev/disk/by-id` folder:
```
zdykstra, line 52: All instances of `Id's` should instead be `IDs`, an apostrophe implies possession - we simply want the plural form here. `folder` should instead be `directory`.

src/installation/guides/zfs.md
zdykstra: > Can modern grub find `/boot` on zfs? Last I checked you needed an older pool version for this to work, have they fixed that?

Any modern feature flags on your zpool will cause it to be no longer readable by GRUB. Unless you use other tools, `/boot` on ZFS with GRUB isn't advisable.

src/installation/guides/zfs.md
```diff
@@ -0,0 +1,225 @@
...
+Installation finished. No error reported.
+```
+
+Create and modify `/etc/dracut.conf.d/zol.conf` to configure Dracut to include
+at least the following values:
+
+```
+hostonly="yes"
+nofsck="yes"
+add_dracutmodules+=" zfs "
```
zdykstra, line 169: Because the pool was created with disks out of `/dev/disk/by-id`, dracut also needs to know to look there. 
The following should also be added to `zol.conf`:

`persistent_policy="by-id"`

aneonlight: Thanks for catching this

ericonr:
The chroot guide has been merged! We can now follow it as the basic structure that this guide fits into.

src/installation/guides/zfs.md
```diff
@@ -0,0 +1,225 @@
...
+alternate mountpoint:
+
+```
+# zpool export zroot
+# zpool import -R /mnt/void zroot
+```
+
+**Important:** Please consult the Base Installation & Configuration sections,
+ignoring the configuration of a fstab, of the [chroot guide](./). These provide
+an in depth guide of a general configuration that is required before continuing.
```
ericonr, line 112: "of the configuration"

src/installation/guides/zfs.md
```diff
@@ -0,0 +1,225 @@
...
+
+After consulting the guide, install `grub` and `zfs` from the new system which
+will build the requirements necessary to boot:
+
+`(chroot)# xbps-install -S -R $REPO grub zfs`
```
ericonr, line 117: I'd prefer if you only mentioned the packages themselves, without the command. If you feel users need an example for how to install packages, make something like `[install](../../xbps/index.md)` (this should be a link to where package installation is explained).

src/installation/guides/zfs.md
```diff
@@ -0,0 +1,225 @@
...
+
+`(chroot)# zpool set cachefile=/etc/zfs/zpool.cache zroot`
+
+Then notate the bootable system for GRUB's autoconfig:
+
+`(chroot)# zpool set bootfs=zroot/ROOT/void zroot`
+
+Finally ensure GRUB recognizes the ZFS module before being installed.
+
+Note: When using id's the ZPOOL_VDEV_NAME_PATH variable must be set to 1.
```
ericonr, line 147: id's -> IDs

src/installation/guides/zfs.md
ericonr: I feel like this is a big footgun, Probably best to advise a separate partition for `/boot` or `/boot/efi` (or zfsbootmenu together with that).

src/installation/guides/zfs.md
```diff
@@ -0,0 +1,225 @@
...
+
+Ensure that the Dracut configuration took and the zfs module is listed in
+[lsinitrd(1)](https://man.voidlinux.org/lsinitrd):
+
+Note: It may be necessary to specify the path to the img if the kernel version
+has changed, see [lsinitrd(1)](https://man.voidlinux.org/lsinitrd).
+
+`(chroot)# lsinitrd -m`
+
+As well check that the ZFS cache is recognized:
```
ericonr, line 203: "Check that the ZFS cache is recognized as well:"

src/installation/guides/zfs.md
```diff
@@ -0,0 +1,225 @@
...
+-rw-r--r--   1 root     root         1376 Apr 17 01:30 etc/zfs/zpool.cache
+```
+
+## Cleanup
+
+Exit chroot:
+
+`(chroot)$ exit`
+
+And un-mount the chroot environment:
```
ericonr, line 216: "and"

src/installation/guides/zfs.md
emacsomancer: @zdykstra can you mention which features interfere with GRUB? I have `/boot` on zfs and all features are listed as 'active' or 'enabled'.

zdykstra: It's surprisingly hard to find a definitive list of "these features will break GRUB". Instead, I'm able to find "these features are compatible with GRUB".  The most concise description I can find is the official OpenZFS guide to installing Ubuntu to ZFS - https://openzfs.github.io/openzfs-docs/Getting%20Started/Ubuntu/Ubuntu%2020.04%20Root%20on%20ZFS.html#step-2-disk-formatting .

As I understand it, if you have `feature@encryption` enabled on the pool, GRUB will refuse to read the pool, even if you don't actually have any encrypted filesystems. I haven't used GRUB to boot ZFS in quite a while, so I can't conclusively say beyond this though. I've moved on from GRUB and I use https://github.com/zdykstra/zfsbootmenu now to boot.

emacsomancer: @zdykstra On my local install I have:

    dozer  feature@encryption             enabled                        local

With not only `/` but also `/boot` on ZFS, with both sitting on a LUKS-encrypted partition, and I can boot with GRUB without issue, so simply having the feature enabled doesn't seem to block GRUB.

Now, I haven't tried to see what happens if I actually try to use native zfs encryption. I certainly imagine a natively-encrypted ZFS `/boot` GRUB would not be happy about.

(I think your zfsbootmenu looks like a better option in general, though I haven't had time to try to test using a working setup for non-UEFI boot.)

zdykstra: You for sure can't use native encryption for `/boot` with GRUB, hah. More importantly, If there's an encrypted filesystem anywhere on the pool that contains `/boot`, GRUB will error out. Project Trident has a legacy/GRUB mode in their installer, and they were forced to skip creating an encrypted `/home/$user` filesystem on the same pool that `/boot` is on, even though GRUB has no need to touch `/home/$user`. Once `feature@encryption` is marked as `active`, things get hinky.

I think if anything is to be changed here, it should be to follow in the official OpenZFS footsteps and provide an explicit list of pool features to be enabled, or to split `/boot` into a new dedicated pool with only those known-good features enabled. 

Not to derail, but `zfsbootmenu` works well under a legacy/BIOS setup by utilizing `syslinux` as the first stage bootloader.

emacsomancer: So perhaps it's not about 'enabled' features, but 'active' features. (Since I'm on the latest zfs and have all features enabled.)

I think providing an explicit list of features to be enabled would be greatly preferable. There's advantages to having `/boot` as part of the same pool (e.g. you don't need to worry about `/boot` running out of space).

(Is [this](https://github.com/zdykstra/zfsbootmenu/tree/syslinux-docs#syslinuxextlinux) still the best resource for setting up `zfsbootmenu` for a non-EFI system?)

zdykstra: I have 26 features enabled on my current pool, there are only 12 features enabled on the boot pool in the OpenZFS documentation. Comparing those two, we get this:

```
diff -u boot-features-sorted all-features-sorted 
--- boot-features-sorted	2020-06-25 11:02:49.915090112 -0500
+++ all-features-sorted	2020-06-25 11:02:41.771950496 -0500
@@ -1,12 +1,26 @@
+feature@allocation_classes
 feature@async_destroy
 feature@bookmarks
+feature@bookmark_v2
+feature@device_removal
+feature@edonr
 feature@embedded_data
 feature@empty_bpobj
 feature@enabled_txg
+feature@encryption
 feature@extensible_dataset
 feature@filesystem_limits
 feature@hole_birth
 feature@large_blocks
+feature@large_dnode
 feature@lz4_compress
+feature@multi_vdev_crash_dump
+feature@obsolete_counts
+feature@project_quota
+feature@resilver_defer
+feature@sha512
+feature@skein
 feature@spacemap_histogram
+feature@spacemap_v2
+feature@userobj_accounting
 feature@zpool_checkpoint
```

I don't have a nuanced enough understanding of what each of those features enables, but they're specifically skipped when creating `bpool` in the Ubuntu install guide. I think at the minimum, if the OpenZFS development themselves follows the practice of limiting which feature flags are turned on, we should as well.

ericonr: Do you think it's feasible to maintain a list like that?

zdykstra: > Do you think it's feasible to maintain a list like that?

I don't see any real issue with it. It doesn't change when OpenZFS adds features, it changes with GRUB adds support for new features. Since they haven't added any new feature support since 2015, this seems to be relatively set in stone.

ericonr: The follow-up then is: is it worth recommending GRUB + ZFS when it has so many limitations?

ericonr:
Hi! If you'd like, you can split the USB guide into another PR, that way we can move with it faster, since ZFS is a more complicated subject/

ericonr:
Please see #397, which should also be mentioned here.

subnut:
Why this PR hasn't been merged? 😕 

I would very much like an official root-on-ZFS install guide, as I am thinking of doing so soon.

zdykstra:
Mostly because the Void-first solution is https://github.com/zbm-dev/zfsbootmenu/wiki . The guide as described here is a legacy way of using ZFS on Void

subnut:
@zdykstra Thanks a lot for that link. That guide is really good.

Shouldn't we mention that guide (you linked above) at least  _somewhere_ in the Void Docs? (preferably inside [Partitioning Notes](https://docs.voidlinux.org/installation/live-images/partitions.html))

paper42:
> Shouldn't we mention that guide (you linked above) at least  _somewhere_ in the Void Docs? (preferably inside [Partitioning Notes](https://docs.voidlinux.org/installation/live-images/partitions.html))

#652

ahesford:
The major component of this PR, installing to a ZFS root, has been superseded by 5c08c2050817dadc2461d9daf4444fec448db4a1. The "Live USB to Full Install" page really amounts to "select the RAM option from the boot menu so that you can overwrite the USB device"; that doesn't seem to warrant a dedicated page in the docs.

If I've overlooked anything helpful in the now-active ZFS guide, please PR improvements as appropriate.

PR #211, opened 2020-04-23
alsa: Mention `apulse` package.

flexibeast:


Vaelatern:
A little more elaboration would be nice. What does apulse do?

flexibeast:
Thanks, updated.

Issue #174, opened 2020-04-02
GNOME setup from base install needs documentation


bobertlo:
Following the wiki, installing gnome and enabling the gdm service does not work. I have been able to figure it out, but not document the process. You definitely need to also install xorg and xinit. Also I think elogind is required?

The main issue is that if you don't have everything right and start gdm it locks up. I have been adding gnome-session to my xinitrc and making sure it works, then gdm seems to be able to work? I also would `touch /etc/sv/gdm/down` and `sv once gdm` to test before getting things to work for sure.

I need to do more VM testing before trying to write up anything, but I think this would be useful information. We don't want people "breaking" their systems.

flexibeast:
> Also I think elogind is required?

That's my understanding. From [the elogind README](https://github.com/elogind/elogind/blob/master/README.md):

> [T]o run GNOME these days means that you need to have support for the login1 D-Bus interface, which is currently only provided by systemd. That is the origin of this project: to take the excellent logind functionality from systemd and provide it as a standalone package.

bobertlo:
Yeah, I'll have to start up some VMs to debug this. Because it seems like just starting dbus and then starting elogind is not enough and you need to reboot? This is pretty important information for anyone who wishes to run GNOME from a manual or base install. :)

flexibeast:
Agreed. :-)

i've just created some VMs to help work on these sorts of issues. In this particular case, we're trying to work out all the steps necessary to get GNOME working?

bobertlo:
Great!

Right. Specifically from a fresh install. I'm thinking a reboot after the set of necessary services is sufficient, but haven't had time to check.

flexibeast:
Okay, working on creating a GNOME section based on the wiki page.

flexibeast:
Closing via https://github.com/void-linux/void-docs/commit/b2db298502350ed628d2fa7d6dedca9cf75365d2.

PR #212, opened 2020-04-23
[WIP] gnome: Add section.

flexibeast:
Addresses #174.

This section reflects what i learned after a few hours experimenting with different scenarios in VMs:

* Merely enabling the `dbus` service is insufficient; after doing so, GNOME fails to start unless the system is first rebooted.
* The `gdm` package is required even for a `startx`-based session - again, without it, GNOME fails to start. Installing `gdm` pulls in a *lot* of dependencies, and i'm not sure which of them are strictly needed.
* If `adwaita-icon-theme` is not installed, no icons are displayed in the GNOME session, rendering it unusable.

Comments welcome!

### DIFF
--- a/src/config/graphical-session/gnome.md
+++ b/src/config/graphical-session/gnome.md
@@ -1,0 +1,30 @@
+# GNOME
+
+## Pre-installation
+
+Install the `dbus` package, ensure the `dbus` service is enabled, and reboot for
+the changes to take effect.
+
+## Installation
+
+Install the `gnome` package for a GNOME environment which includes GNOME
+applications.
+
+A minimal GNOME environment can be created by installing the `gnome-session`,
+`gdm` and `adwaita-icon-theme` packages. (Note, however, that not all GNOME
+features may be present or functional.) `gdm` defaults to providing a Wayland
+session, via the `mutter` window manager. For an Xorg session, install the
+`xorg` package, then select 'GNOME on Xorg' at the GDM login screen, or start
+`gnome-session` via `~/.xinitrc`.
+
+If you require [ZeroConf](http://www.zeroconf.org/) support, install the `avahi`
+package and enable the `avahi-daemon` service.
+
+## Display Manager
+
+To use a graphical display manager for logging in to GNOME, enable the `gdm`
+service. Test the `gdm` service before enabling it:
+
+```
+# sv once gdm
+```
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -48,6 +48,7 @@
       - [Xorg](./config/graphical-session/xorg.md)
       - [Wayland](./config/graphical-session/wayland.md)
       - [Session Management](./config/graphical-session/session-management.md)
+      - [GNOME](./config/graphical-session/gnome.md)
    - [Multimedia](./config/media/index.md)
       - [ALSA](./config/media/alsa.md)
       - [PulseAudio](./config/media/pulseaudio.md)


HadetTheUndying:
Really recommend not even including a Minimal Install section. Gnome does not function well without all of its associated packages. You also might want to add a bit about `chrome-gnome-shell` 

Also might want to include some mentions about optional services to enable. 
From the old Wiki;
```# ln -s /etc/sv/NetworkManager /var/service # Make sure other Networkmanagers such as dhcpcd or wicd are disabled!
# ln -s /etc/sv/avahi-daemon /var/service   # zeroconf support
# ln -s /etc/sv/bluetoothd /var/service     # Bluetooth Support
# ln -s /etc/sv/cupsd /var/service          # Printer support
# ln -s /etc/sv/gdm /var/service            # Make sure other display managers such as LXDM are disabled
```

I'm unsure why you had to reboot. I use gnome on both of my workstations and this has never been the case for me.

flexibeast:
@HadetTheUndying: Thanks for your feedback!

> Really recommend not even including a Minimal Install section. Gnome does not function well without all of its associated packages.

Fair enough. @bobertlo, what do you think?

> You also might want to add a bit about `chrome-gnome-shell`

This will only be needed if the "Minimal Install" part is dropped, as the current template for `gnome` pulls in `chrome-gnome-shell` as a dependency.

> Also might want to include some mentions about optional services to enable.
From the old Wiki

i saw that part on the old wiki page, but decided to drop it. i'm happy to add mention of those services back in if @bobertlo wants that.

> I'm unsure why you had to reboot. I use gnome on both of my workstations and this has never been the case for me.

Interesting. Was that after setting up GNOME from a base install, which is what i was doing in my VMs? How long ago did you do this? (i'm wondering if something has changed since you last installed GNOME.) @bobertlo also [found that a reboot seemed to be needed](https://github.com/void-linux/void-docs/issues/174#issuecomment-609461389).

Vaelatern:
As for a minimal section, a disclaimer that gnome does not work well in a minimal install is adequate. Some people don't have the bandwidth to make things work beautifully, but still need an interface they recognize.

bobertlo:
Yeah, I have had to reboot every time I've setup GNOME lately. I have been using a base chroot install since I run full disk encryption.

I don't think there is any harm in mentioning extra services briefly. Some of those (like bluetooth and cups) though I am unsure about.

I tend to agree the minimal install section as written is maybe not the best way to offer it. Also the `gnome` does not pull all of the necessary `xorg` stuff with it. so there should maybe be a link to that page. I think I would rather just see this all presented together. The minimal header seems to be misleading and present maybe an undesirable goal to the reader.

I would err on not including the bug report at the end. If it is found to still be an issue (which I haven't noticed in my wayland sessions) it could be added back in. This is specifically is one of the types of things I really want to be careful about importing from the wiki. Those pages are quite old and a lot of the software has greatly improved, either in upstream support or our packaging.

It may also be worth mentioning that using `gdm` defaults to a wayland session, while adding `gnome-session` to your `.xinitrc` uses xorg. Also it is definitely worth noting that one should try `# sv once gdm` before enabling the service because it is hard to disable. I made this mistake on a brand new installation with full disk encryption :|

flexibeast:
@bobertlo: Thanks for the feedback!

> I don't think there is any harm in mentioning extra services briefly. Some of those (like bluetooth and cups) though I am unsure about.

\*nod\* My thought is that, since both those things have their own Handbook sections, there's no need to mention them.

> I tend to agree the minimal install section as written is maybe not the best way to offer it. Also the gnome does not pull all of the necessary xorg stuff with it. 

Yeah, `gnome` not pulling in `xorg`, `elogind`, `gdm` or `adwaita-icon-theme` is why i mentioned those things under "Pre-installation". Is/are there any reason(s) why the `gnome` package doesn't have those packages as dependencies, since they're basically not optional? (Obviously `adwaita-icon-theme` is technically optional *if* one is using an existing GNOME config - but if not, GNOME without the icons provided by `adwaita-icon-theme` is .... suboptimal.)

So are you saying that, instead of mentioning `xorg`, there should just be a link to the "Xorg" section of the Handbook? Should there be a link to the "Wayland" section also?

> I think I would rather just see this all presented together. The minimal header seems to be misleading and present maybe an undesirable goal to the reader.

Fair enough - i'll remove that part.

> I would err on not including the bug report at the end. If it is found to still be an issue (which I haven't noticed in my wayland sessions) it could be added back in. This is specifically is one of the types of things I really want to be careful about importing from the wiki. Those pages are quite old and a lot of the software has greatly improved, either in upstream support or our packaging.

\*nod\*

> It may also be worth mentioning that using gdm defaults to a wayland session, while adding gnome-session to your .xinitrc uses xorg. Also it is definitely worth noting that one should try # sv once gdm before enabling the service because it is hard to disable.

Sure, will update the PR accordingly.

flexibeast:
i've just updated the PR with people's suggestions thus far.

@bobertlo, i've taken the liberty of linking to the "Xorg" and "Wayland" sections in the "Pre-installation" part - hope that's what you were after. :-)

HadetTheUndying:
I think that looks pretty good. Covers everything.

src/config/graphical-session/gnome.md
```diff
@@ -0,0 +1,34 @@
...
+service is enabled, and reboot for the changes to take effect.
+
+## Installation
+
+Install the `gnome` package for a GNOME environment which includes GNOME
+applications.
+
+A minimal GNOME environment can be created by installing the `gnome-session` and
+`gdm` packages. Note, however, that such an minimal environment may not work
+well.
```
bobertlo, line 18: I feel like this could be worded better as 'not all features may be present' or 'functional' or something?

bobertlo:
One small point, but otherwise looks good to me.

src/config/graphical-session/gnome.md
flexibeast: That's the sort of phrasing i was looking for! How about: "Note, however, that not all GNOME features may be present or functional"?

src/config/graphical-session/gnome.md
```diff
@@ -0,0 +1,34 @@
+# GNOME
+
+## Pre-installation
+
+GNOME requires a display server such as [Xorg](./xorg.md) or a
+[Wayland](./wayland.md) compositor.
```
ericonr, line 6: Wouldn't it be more correct to say that "it can provide its own Wayland compositor (`mutter`)" ?

src/config/graphical-session/gnome.md
```diff
@@ -0,0 +1,34 @@
...
+
+Install the `elogind` and `adwaita-icon-theme` packages, ensure the `dbus`
+service is enabled, and reboot for the changes to take effect.
```
ericonr, line 9: So, if people are installing this from the `gnome` package, which is what you're recommending below, they will get elogind and the icon theme as dependencies. At least that's what `xbps-query -Rx gnome --fulldeptree` tells me. We can recommend they reboot after completing the installation, instead.

src/config/graphical-session/gnome.md
ericonr: We could mention `elogind` and `adwaita-icon-theme` here, instead. `gdm` seems to pull in elogind already, so we'd only need to mention adwaita. Do we need GDM for a basic GNOME session?

So the "minimal GNOME environment" would require `gnome-session`, `gdm` (or `elogind`) and `adwaita-icon-theme`.

src/config/graphical-session/gnome.md
```diff
@@ -0,0 +1,34 @@
...
+applications.
+
+A minimal GNOME environment can be created by installing the `gnome-session` and
+`gdm` packages. Note, however, that such an minimal environment may not work
+well.
+
+Starting `gnome-session` via `~/.xinitrc` provides an Xorg session.
+
+If you require ZeroConf support, install the `avahi` package and enable the
+`avahi-daemon` service.
```
ericonr, line 23: Should we link http://www.zeroconf.org/ ? Or is that a case of "you know it if you need it"?

src/config/graphical-session/gnome.md
ericonr: We should probably require gdm, it's what pulls in the GNOME window manager.

src/config/graphical-session/gnome.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Display manager
```
ericonr, line 25: Also perhaps capitalize Manager? I think it's how most titles are written.

ericonr:
Can you start a Wayland session from the terminal? Both sessions file use the same `gnome-session` executable to launch GNOME. It makes sense that putting gnome-session in .xinitrc would help with getting gnome-session to use Xorg, so if you start without that and simply run `gnome-session` from a terminal it should start a Wayland session?

Perhaps determine how this works so we can make it clear how to start the wayland session from a terminal.

flexibeast:
@ericonr: Thanks for all your feedback and help - much appreciated!

i've now tried installing installing `dbus`, enabling the `dbus` service, rebooting the VM, installing `gnome-session`, `gdm` and `adwaita-icon-theme`, then starting `gnome-session` directly from the terminal. It fails, with repeated errors like:

    Cannot open display:
    Unable to init server: Could not connect: Connection refused

not to mention errors like:

    ServiceUnknown: The name org.freedesktop.systemd1 was not provided by any .service files

ericonr:
@flexibeast if you restart that machine, does it work? If it doesn't, then we know that you can either start `gnome-session` from `.xinitrc` or use `gdm`.

https://stackoverflow.com/questions/31213773/how-to-start-gnome-wayland-session-from-command-line#31213849

flexibeast:
@ericonr: i shut down that VM last night after posting my comment, and just restarted it now - having not done anything else with it - to check if that made any difference.

`gnome-session` alone still results in the same error; `gnome-session --session gnome-wayland` results in:

    WARNING: Falling back to non-systemd startup procedure due to error: GDBus.Error:org.freedesktop.DBbus.Error.ServiceUnknown: The name org.freedesktop.systemd1 was not provided by any .service files
    WARNING: Could not get session id for session Check that logind is properly installed and pam_systemd is getting used at login.
    Unable to init server: Could not connect: Connection refused

i confirmed the `dbus` service is running, and that the `elogind` package is installed. i then tried enabling the `elogind` service, and without a reboot, doing `gnome-session`, which resulted in `dconf-CRITICAL` errors:

    unable to create directory '/run/user/1000/dconf': Permission denied. dconf will not work properly

culiminating in:

    WARNING: IceLockAuthFile failed: No such file or directory

Trying `gnome-session --session gnome-wayland` had the same effect. Subsequently restarting the VM made no difference.

ericonr:
@flexibeast GNOME seems like a weird beast. In that case, I think the section covers most of what is needed to set it up.

Do you think we should create a separate section for desktop environments, like #223 ? I think GNOME can be standalone for now, and if we need to add one for whatever reason we can create a new section.

flexibeast:
@ericonr: Thanks! i'd like to also get an ok from @bobertlo before merging this.

Re. #223, i agree - let's not create a distinct section for DEs until there's something more than just GNOME. i'm hoping to put together a section for KDE, so we could create a DE section once that's complete.

bobertlo:
I didnt see this mention but came here to say LGTM

flexibeast:
@bobertlo: Thanks, merged.

PR #223, opened 2020-05-01
section: Add graphics environment section

linuxpotion:


Vaelatern:
that's not how commit messages work.

ericonr:
I think we could wait to add this one if it proves necessary. Most Xorg setup is described in `xorg.md`, and Wayland stuff is mostly distro agnotic, if you use the tidbits given out in `wayland.md`. We *are* adding a GNOME section right now, but I don't think we need a new section to contain it.

@flexibeast any thoughts, since you're writing the GNOME one?

flexibeast:
@ericonr: i guess once the "KDE" section gets added, we could maybe have a "Desktop Environments" section? Which would make even more sense should we wish to eventually add pages for e.g. XFCE, MATE, LXDE, etc. (Do we?) But i don't think we should go beyond DEs to including information about all the various WMs; that certainly seems unnecessary to me.

ericonr:
@flexibeast I think it depends on the the difficulty of setting up those DEs. If they are too complicated, we should create a section. If they aren't, I would leave KDE and GNOME as separate ones, because they are the biggest DEs and it would discourage adding pages for other DEs unless they are actually necessary.

flexibeast:
@ericonr: \*nod\* Good points.

So my thinking is:

* For now, we don't create a distinct "Desktop Environments" section, but leave "GNOME" and "KDE" where they are.
* If it later becomes apparent that sections are needed for other DEs, we create them, and possibly create a DE section at that point.
* We close this PR.

WDYT?

ericonr:
@flexibeast agreed on all points! I will close it, then.

PR #113, opened 2019-08-25
new page: config/network/connman

lugubris:

PR #225, opened 2020-05-01
management: Add reference to xbps.d(5) man page.

flexibeast:


ericonr:
What do you think including a small item about each xbps-application? We have the list for "most general package management", then we add the more "specific" tools. Also mention xbps.d on that one.

That way I think we could close something like #113, perhaps.

flexibeast:
@ericonr: i like that idea, especially since XBPS is a unique package manager that isn't (yet?) used and documented elsewhere; the Handbook seems like a good place to give a high-level overview of its components.

How would you feel about this PR being merged for now, and a separate issue/PR being opened for your suggestion?

Not sure i understand how your suggestion would help close #113 though .... Was that the actual issue number you meant?

ericonr:
I think merging it is fine! And yes, I meant #103 :p

flexibeast:
Thanks, merged. New issue opened as #235.

Ah! Yes, i see what you mean re. #103. :-)

PR #229, opened 2020-05-02
book.toml: add linkcheck plugin

ericonr:
This addition wouldn't make anyone unable to build locally, because it only uses the plugin if it finds it in the path. For CI, it makes sense to error out.

I need help with setting up CI for it, though. The current one only runs `vmdfmt`, we'd need to run `mdbook build` as well.

### DIFF
--- a/src/contributing/void-docs/style-guide.md
+++ b/src/contributing/void-docs/style-guide.md
@@ -99,6 +99,23 @@ They should not be formatted like this:
 [https://www.example.com/](https://www.example.com/)
 ```
 
+### Checking links
+
+If you're including new links (either internal or external) in the docs or
+changing the current file structure, you should install `mdbook-linkcheck`,
+which can be obtained from the Void repos or by using `cargo`. You can then
+build the mdBook locally, which will run a linkcheck as well, or run it in
+standalone mode:
+
+```
+$ mdbook-linkcheck -s
+```
+
+This way, linkcheck will verify all the references, and warn you if there are
+any issues. If any link you're using is correct but generating errors for some
+reason, you can add its domain to the exclude list in `book.toml`, under the
+`[mdbook.linkcheck]` key.
+
 ## Case
 
 Filenames and directories should use [kebab
--- a/book.toml
+++ b/book.toml
@@ -6,3 +6,7 @@ title = "Void Linux Handbook"
 
 [output.html]
 theme = "src/theme"
+
+[output.linkcheck]
+follow-web-links = true
+exclude = [ "kernel.org", "\\.onion", "localhost", "hushan.tech" ]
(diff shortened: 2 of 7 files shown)


ericonr:
@the-maldridge is there an easy way to test how this would behave in our current infrastructure? In my system it doesn't find the plugin and simply warns the user, exiting without an error. I don't know if the current system processess the standard output. I'd like to merge this for local testing when messing with PRs, but need  to be sure it doesn't affect deployment.

the-maldridge:
https://github.com/void-ansible-roles/mdbook/blob/master/templates/mdbook_build.sh.j2

ericonr:
Shouldn't be an issue, then. Thanks!

ericonr:
@flexibeast I added some recommendations for checking links in style guide as well. If you think they are okay, I can merge this after I get the mdbook-linkcheck package accepted into void-packages.

src/contributing/void-docs/style-guide.md
```diff
@@ -99,6 +99,22 @@ They should not be formatted like this:
...
+run it in standalone mode:
+
+```
+$ mdbook-linkcheck -s
+```
+
+This way, linkcheck will verify all the references, and warn you if there are
+any issues. If any link you're using is correct but generating errors for some
+reason, you can add it to the exclude list in `book.toml`.
+
```
flexibeast, line 117: Any reason(s) we shouldn't make this non-optional, like `vmdfmt`, rather than just 'recommended'? If not, how about something like:

"If adding links or changing the link structure, use \`mdbook-linkcheck\`. \`mdbook-linkcheck\` verifies all links, both internal and external. Install the \`mdbook-linkcheck\` package, then do:

\```
$ mdbook-linkcheck -s
\```

If there are links that are correct, but generating errors for some reason, the [domains? links? The exclude list in this PR has domains] can be added to the exclude list in \`book.toml\`."

ericonr: It's non optional in that you can not run it locally, then CI should catch any issues and you will have to solve them manually. What do you think?

Regarding the links, yes, I should specify domain. Web jargon is not my thing :p

flexibeast: Well, i'm thinking about trying to reducing unnecessary work. We tell people to run `vmdfmt` on their contribution before submitting so that everyone involved doesn't have to go through the dance of the submitter being told that the CI check has failed and that they need to fix the PR; i'd like to take the same approach with `mdbook-linkcheck`.

ericonr: Fair enough. It's in void-packages already so it shouldn't be too much of an issue. I will reword it :)

PR #233, opened 2020-05-03
downloading: remove header about using GPG

### LINKED ISSUE #232: asc file refered to in downloads doesn't exist
https://docs.voidlinux.org/installation/live-images/downloading.html#verify-image-authenticity links to the Void GPG key, which errors out with 404 right now (https://alpha.de.repo.voidlinux.org/live/current/void_images.asc). We should probably reword this part.

@the-maldridge I'm not sure how to explain this one. Should I just remove the mention about it being possible to download the asc file from that link?

ericonr:
The current verification is done through signify, and the GPG key is expired.

Fixes #232

### DIFF
--- a/src/installation/downloading.md
+++ b/src/installation/downloading.md
@@ -1,10 +1,12 @@
-# Downloading Images
+# Downloading Installation Media
 
-The most recent live images can be downloaded from
-<https://alpha.de.repo.voidlinux.org/live/current/>. Previous releases can be
-found under <https://alpha.de.repo.voidlinux.org/live/>, organized by date.
+The most recent live images and rootfs tarballs can be downloaded from
+<https://alpha.de.repo.voidlinux.org/live/current/>. They can also be downloaded
+from [other mirrors](../xbps/repositories/mirrors/index.md). Previous releases
+can be found under <https://alpha.de.repo.voidlinux.org/live/>, organized by
+date.
 
-## Verify images
+## Verifying images
 
 Each image release's directory contains two files used to verify the image(s)
 you download. First, there is a `sha256.txt` file containing image checksums to
@@ -18,7 +20,7 @@ want to download both files:
 $ wget http://alpha.de.repo.voidlinux.org/live/current/sha256.{txt,sig}
 ```
 
-### Verify image integrity
+### Verifying image integrity
 
 You can verify the integrity of a downloaded file using
 [sha256sum(1)](https://man.voidlinux.org/sha256sum.1) with the `sha256.txt` file
@@ -32,16 +34,17 @@ void-live-x86_64-musl-20170220.iso: OK
 
 This verifies that the image is not corrupt.
 
-### Verify digital signature
+### Verifying digital signature
 
 Prior to using any image you're strongly encouraged to validate the signatures
 on the image to ensure they haven't been tampered with.
 
 Current images are signed using a signify key that is specific to the release.
-If you're on Void already, you can obtain the keys from the void-release-keys
+If you're on Void already, you can obtain the keys from the `void-release-keys`
 package, which will be downloaded using your existing XBPS trust relationship
-with your mirror. You will also need a copy of signify; on Void this is provided
-by the `outils` package.
+with your mirror. You will also need a copy of
+[signify(1)](https://man.voidlinux.org/signify.1); on Void this is provided by
+the `outils` package.
 
 If you are not currently using Void Linux you will need to obtain a copy of
 signify by other means, and the appropriate signing key from our git repository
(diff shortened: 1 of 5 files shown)


Vaelatern:
Should really be replaced with a much shorter section on verifying with signify?

ericonr:
That's what I was thinking too. Do you think the current one in the downloads page is enough?

Vaelatern:
I'd rather docs be as authoritative as possible. I like your proposal.

travankor:
I think you can use `minisign` wih `signify`, too (might need to double check this).

ericonr:
@Vaelatern can you take a look at the current state? I removed the GPG information because it only pertains to old images. Should we keep it? I will make a PR for the website too.

@travankor I will take a look at minisign. Do you think we'd need an example with it or just mention that it can be used as well?

Vaelatern:
These changes look reasonable.

ericonr:
@travankor minisign seems to require removing the first line from the release key, at least from what I tried. It also required renaming the sig file to `sha256.txt.minisig`. GIven that it has a bunch of additional steps, do you think we should still recommend it?

travankor:
In that case, probably not. Although the main advantages of minisign is that it is more widespread and available than signify.

Maybe mention if you cannot get signify to work then minisign is an option?

ericonr:
@travankor accessibility is important still, I think, so I added the necessary steps for minisign as well. Could you take a look at them, please?

ericonr:
@pullmoll I'm removing the instructions for GPG signing, but it seems you still use them in your repositories. Are you considering switching to signify?

I'm not sure how I could keep GPG instructions while giving context for their usage, if you would still like to keep them.

src/installation/downloading.md
```diff
@@ -0,0 +1,88 @@
...
+   [here](https://repology.org/project/signify-openbsd/versions) for your
+   distribution.
+- Install `signify-osx` with homebrew in macOS.
+
+If you can't obtain `signify` for some reason (e.g. you are on Windows and can't
+use WSL or MinGW), you can use
+[minisign(1)](https://man.voidlinux.org/minisign.1) to verify the file.
+
+If you are not currently using Void Linux, it will also be necessary to obtain
+the appropriate signing key from our git repository
```
flexibeast, line 60: "Git"

src/installation/downloading.md
```diff
@@ -0,0 +1,88 @@
...
+[here](https://github.com/void-linux/void-packages/tree/master/srcpkgs/void-release-keys/files/).
+
+Once you've obtained the key, you can verify your image with the `sha256.sig`
+file. An example is shown here verifying the GCP musl filesystem from the
+20191109 release:
```
flexibeast, line 65: "The following example demonstrates verifying ..."

src/installation/downloading.md
```diff
@@ -0,0 +1,88 @@
...
+
+```
+$ signify -C -p /etc/signify/void-release-20191109.pub -x sha256.sig void-GCP-musl-PLATFORMFS-20191109.tar.xz
+Signature Verified
+void-GCP-musl-PLATFORMFS-20191109.tar.xz: OK
+```
+
+If the verification process does not spit out the expected "OK" status then do
```
flexibeast, line 73: s/spit out/produce/
"status, do"

src/installation/downloading.md
```diff
@@ -0,0 +1,88 @@
...
+not use it! Please alert the Void Linux team of where you got the image and how
+you verified it and we will follow up.
```
flexibeast, line 75: "it, and"
Perhaps s/follow up/investigate/? But i don't mind either way.

src/installation/downloading.md
```diff
@@ -0,0 +1,88 @@
...
+
+For verification with `minisign`, it is necessary to rename the `sha56.sig` file
+to `sha256.txt.minisig` and remove the first line from the `.pub` release key.
+An example is shown here for verifying the `sha256.txt` file from the 20191109
```
flexibeast, line 79: "The following example demonstrates verifying ..."

src/installation/downloading.md
```diff
@@ -0,0 +1,88 @@
...
+release:
+
+```
+$ minisign -Vm sha256.txt -f -p void-release-20191109.pub
+void-release-20191109.pub: Success
+```
+
+The same warning as above applies. If the verification process isn't successful,
+do not use the file and warn the Void Linux team.
```
flexibeast, line 88: "file - warn"

src/installation/downloading.md
ericonr: Definitely! I think I didn't revise this part too much, should have done it a bit more.

src/installation/downloading.md
ericonr: I used "follow up on it". What do you think?

flexibeast: Sure. that's fine. :-)

PR #234, opened 2020-05-03
contributing: add small index and link to Join Us page

ericonr:
@flexibeast @bobertlo can you take a quick look at this one? I feel like we needed to link from this page to the Join Us one.

### DIFF
--- a/src/contributing/index.md
+++ b/src/contributing/index.md
@@ -3,7 +3,11 @@
 There's more to running a distribution than just writing code. This section
 explains how to be an active part of Void.
 
+Please also visit the Void Web site for further information about [how to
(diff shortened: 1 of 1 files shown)


flexibeast:
@ericonr: i like this overall. :-) Perhaps we could add a header titled "Section contents" before the links? What styling should we use for lists in the Handbook - dashes, or dot-points?

i'd prefer if the link text matched the names of the sections to which they're linking.

i'd also prefer the final sentence to be moved up to be part of the first paragraph, and perhaps say something like:

"Please also visit the Void Web site for further information about \[how to participate\](https://voidlinux.org/contribute/), including our communication channels, and how to contribute to the Void package repository."

ericonr:
Updated!

PR #237, opened 2020-05-05
kde: Add section.

flexibeast:
This PR is based on the 'KDE' page of the old wiki.

The `kde5` package pulls in the `elogind`, `NetworkManager`, `sddm`, `udisks2` and `upower` packages, so there's no need for the user to install them manually.

The `dbus` service does not need to be enabled for a Plasma session to be started via `startx`. However, one does need to enable *both* the `dbus` and `NetworkManager` services (rather than e.g. `NetworkManager` alone) in order to use the "Networks" widget.

Since `elogind` is available, there's no need to include a 'troubleshooting' question with the answer "Install `elogind`". i presume the presence of `elogind` also means there should be no need for users to be added to the `storage` group?

i've also not included the "No shutdown, reboot buttons ..." entry, as i didn't observe this behaviour in testing.

For some reason, trying to do `sv once sddm` didn't work, but resulted in the message

    fail: sddm: unable to change to service directory: file does not exist
    
Does anyone have any suggestions as to what the problem might be here? i'd certainly like the page to suggest that users test out SDDM that way before enabling it.

The addition of this section would mean we could create a 'Desktop Environments' section in the Handbook, but i'd prefer to not do anything on that front until this PR is merged. One step at a time. :-)

### DIFF
--- a/src/config/graphical-session/kde.md
+++ b/src/config/graphical-session/kde.md
@@ -1,0 +1,19 @@
+# KDE
+
+## Installation
+
+Install the `kde5` package, and optionally, the `kde5-baseapps` package.
+
+To use the "Networks" widget, enable the `dbus` and `NetworkManager` services.
+
+Installing the `kde5` package also installs the `sddm` package, which provides
+the `sddm` service for the Simple Desktop Display Manager; [test the
+service](../services/managing.html#testing-services) before enabling it. If you
+are not intending to run SDDM via a remote X server, you will need to install
+either the `xorg-minimal` package or the `xorg` package. By default, SDDM will
+start an Xorg-based Plasma session, but you can request a Wayland-based Plasma
+session instead.
+
+If you wish to start an Xorg-based session from the console, use
+[startx](./xorg.html#startx) to run `startplasma-x11`. For a Wayland-based
+session, run `startplasma-wayland` directly.
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -49,6 +49,7 @@
       - [Xorg](./config/graphical-session/xorg.md)
       - [Wayland](./config/graphical-session/wayland.md)
       - [GNOME](./config/graphical-session/gnome.md)
+      - [KDE](./config/graphical-session/kde.md)
    - [Multimedia](./config/media/index.md)
       - [ALSA](./config/media/alsa.md)
       - [PulseAudio](./config/media/pulseaudio.md)


ericonr:
As a matter of curiosity, do you think these two guides could fit inside a single page? It could make searching for them slightly harder, but that's the only disadvantage I can think of.

Regarding the issue you had, I've seen mentions of SDDM here https://github.com/void-linux/void-packages/issues/660#issuecomment-405060639 , but according to the checklist it should have been solved already. I can look further into it later.

flexibeast:
@ericonr:

> As a matter of curiosity, do you think these two guides could fit inside a single page? It could make searching for them slightly harder, but that's the only disadvantage I can think of.

They certainly could, but one issue is that if we end up starting to add content about other DEs as well (MATE, LXDE), the page could start to get a bit large, even without taking into consideration the possibility that those DEs might require more details than GNOME and KDE.

> I can look further into it later.

Thanks, that would be much appreciated. :-)

ericonr:
For the record, enabling the service worked for me. `sv once` is really not working, and acpid, for example, had the same issue you mentioned.

flexibeast:
@ericonr: Thanks for looking into this. So, change both `gnome.md` and this PR to specify `runsv /etc/sv/{gdm,sddm}`?

ericonr:
That's what I'd do. With `# runsv`, of course.

flexibeast:
Thanks, PR updated.

src/config/graphical-session/kde.md
```diff
@@ -0,0 +1,21 @@
...
+
+To use the "Networks" widget, enable the `dbus` and `NetworkManager` services.
+
+To start Plasma via `startx`, install the `xorg` package and call
+`startplasma-x11` from `~/.xinitrc`.
+
+## Display Manager
+
+To use a graphical display manager for logging in to KDE, enable the `sddm`
+service. Test the `sddm` service before enabling it:
```
ericonr, line 15: I'm never sure about this one. "log in to" or "log into"?

flexibeast: My thought is that the infinitive is "to log in", not to "to log into", which [Wiktionary](https://en.wiktionary.org/wiki/log_in) seems to agree with ....

ericonr: https://english.stackexchange.com/questions/5302/log-in-to-or-log-into-or-login-to agrees with you :D, so we can go with that. Let's wait for the CI then I think we can merge it?

flexibeast:
@ericonr: As per [this comment](https://github.com/void-linux/void-docs/pull/249#issuecomment-625564118), before we merge, i'd like to get @Duncaen's input on what we should have in the "Display Manager" section.

ericonr:
Hmm damn! I had thought of the solution of adding a down file, but it seemed too complex, in a way. Since it isn't a single command, I'd suggest we add the proper section in the services section and link to it. That way, if things change, we have a single place where we need to change them.

flexibeast:
The outcome of https://github.com/void-linux/void-packages/issues/21766 will need to be reflected in this PR.

flexibeast:
Also blocked by https://github.com/void-linux/void-packages/issues/21862.

Wojnr:
Plasma-workspace is also required to be detected in sddm (other display managers too i guess) and is not included in kde5 deps.

Wojnr:
plasma-workspace is also required to enable plasma session in sddm (other display managers too I guess) and is not included in kde5 deps. I think without it I got black screen but I'm not sure. Will check everything next week like I said on reddit.

the-maldridge:
I have resolved the blocked issue, please see my closing remark.

flexibeast:
Now that https://github.com/void-linux/void-packages/issues/21862 has been resolved, i think the remaining blockers are:

* the `dbus-run-session` issue discussed in https://github.com/void-linux/void-packages/issues/21766, which i think, from a docs perspective, we can probably consider 'resolved' once #263 is merged. We can always revise this section later based on the outcome of https://github.com/void-linux/void-packages/issues/21766.
* some comments by user "Mateusz Wojnarowski". For some reason the below comments are no longer accessible; i've taken the URLs and contents from GH notification emails. 

  * https://github.com/void-linux/void-docs/pull/237#issuecomment-628961425:
  > plasma-workspace is also required to be detected in sddm (other display managers too i guess) and is not included in kde5 deps. 
  * https://github.com/void-linux/void-docs/pull/237#issuecomment-628962178
  > plasma-workspace is also required to enable plasma session in sddm (other display managers too I guess) and is not included in kde5 deps. I think without it I got black screen but I'm not sure. Will check everything next week like I said on reddit.

Wojnr:
I removed them because it was no longer an issue when I reinstalled void. In both cases I installed kde5 package which has plasma-desktop in deps, which has plasma-workspace. I have no clue why it didn't install first time. Basically I followed tutorials and installed xorg, kde5 and sddm and plasma worked instantly. As I said, no idea why it didn't happen first time, I did exactly the same things.

flexibeast:
@Wojnr: Ah okay, thanks for clarifying.

flexibeast:
@ericonr: Now you've merged #263, and further to [this comment of mine](https://github.com/void-linux/void-docs/pull/237#issuecomment-629925605), i'd like to merge this.

src/config/graphical-session/kde.md
```diff
@@ -0,0 +1,17 @@
...
+Install the `kde5` package, and optionally, the `kde5-baseapps` package.
+
+To use the "Networks" widget, enable the `dbus` and `NetworkManager` services.
+
+Installing the `kde5` package also installs the `sddm` package, which provides
+the `sddm` service for the Simple Desktop Display Manager; [test the
+service](../services/managing.html#testing-services) before enabling it. If you
+are not intending to run SDDM via a remote X server, you will need to install
+either the `xorg-minimal` package or the `xorg` package. SDDM defaults to
+providing an Xorg session, but a Wayland session is available.
```
ericonr, line 14: it's not clear if SDDM is itself using Wayland/Xorg (like GDM does) or if it can load a Plasma Wayland or Xorg session. I'd use "Plasma" instead of "SDDM", because you can load those sessions from any DM (or no DM at all).

src/config/graphical-session/kde.md
```diff
@@ -0,0 +1,17 @@
...
+
+If you wish to start an Xorg-based KDE session from the console, use
+[startx](./xorg.html#startx) to run `startplasma-x11`.
```
ericonr, line 17: If you want to be thorough, you can mention that there's a `startplasma-wayland` program for the Wayland session.

ericonr: It wouldn't be launched through `startx`, of course.

src/config/graphical-session/kde.md
flexibeast: @ericonr: Ah, yes, point taken. SDDM itself needs an X server (cf. https://github.com/void-linux/void-packages/issues/21862). Basically what i'm trying to get at is: "SDDM will start an Xorg-based Plasma session by default, but you can also start a Wayland-based Plasma session." Does that make more sense? If so, i'll try to find better wording along those lines.

src/config/graphical-session/kde.md
flexibeast: @ericonr: Good suggestion; i'll add that in.

src/config/graphical-session/kde.md
ericonr: I think that works.

flexibeast: Thanks, PR updated.

PR #240, opened 2020-05-05
services/configuring: overhaul section

travankor:


Vaelatern:
Correct me if I'm wrong @ericonr but I think we don't need an example of editing config, as it's clear what is meant (and even if we did, it would be wrong to suggest tee).

travankor:
Fair enough I see what you mean, but the "editing services" section has a lot of examples (that might be unecessary, like `vi service`), while the "configuring services" section is sparse.

Maybe, it would be a good idea to remove some of these examples to have a more consisitent tone?

I think all I wanted to say was that you usually need to pass the variables as `OPTS`, unless the service does something different (which they can, see the agetty)

travankor:
Followed @Vaelatern's advice and removed the duplicated examples.

src/config/services/configuring.md
```diff
@@ -5,28 +5,13 @@ options can be set via a `conf` file in the service directory. This allows
 service customization without modifying the service directory provided by the
 relevant package.
 
+Check the service file for how to pass configuration parameters, but services
+usually expect `OPTS=--value1 --value2` as values.
```
ericonr, line 9: I think you need to put the value of OPTS inside quotes here, don't you?

src/config/services/configuring.md
```diff
@@ -5,28 +5,13 @@ options can be set via a `conf` file in the service directory. This allows
...
-with the new one:
-
-```
-# sv down service_name
-# rm /var/service/service_name
-# ln -s /etc/sv/service_name_edited /var/service/
-```
+To edit a service, first duplicate its service directory under a different name.
+Edit the new service file as needed. After editing, the old service should then
+be stopped, disabled and replaced with the new service.
```
ericonr, line 17: I think it'd be better if you explained that this one is necessary because the `run` file will be replaced by the package manager, so editing it is not enough.

And the phrases were separated because they had commands between them, so perhaps they can flow better? Something like

```
To edit a service, first duplicate its service directory under a different name. Then, edit the new service file as needed. Finally, the old service should then be stopped, disabled and replaced with the new service.
```

src/config/services/configuring.md
ericonr: "but usually they are expected as `OPTS`..."

src/config/services/configuring.md
```diff
@@ -5,28 +5,14 @@ options can be set via a `conf` file in the service directory. This allows
...
-After editing, the old service should then be stopped, disabled and replaced
-with the new one:
-
-```
-# sv down service_name
-# rm /var/service/service_name
-# ln -s /etc/sv/service_name_edited /var/service/
-```
+To edit a service, first copy its service directory to a different directory
+name, otherwise `xbps` will overwrite the service directory. Then, edit the new
```
ericonr, line 16: Suggest

```
otherwise [xbps-install(1)] (man page link) might overwrite your changes.
```

travankor: good catch

travankor: btw the man pages on voidlinux.org are outdated. The websitel says September 20,2016, while the local version is March 3, 2020.

ericonr: They are outdated since xbps switched to zstd. I changed the little bits necessary in the tool that extracted manuals so it can work with zstd just yesterday, and we're looking to put it into the infrastructure. Software like iwd doesn't even have manpages there right now.

src/config/services/configuring.md
```diff
@@ -5,28 +5,15 @@ options can be set via a `conf` file in the service directory. This allows
...
-```
-# sv down service_name
-# rm /var/service/service_name
-# ln -s /etc/sv/service_name_edited /var/service/
-```
+To edit a service, first copy its service directory to a different directory
+name, otherwise [xbps-install(1)](https://man.voidlinux.org/xbps-install.1) will
+overwrite the service directory. Then, edit the new service file as needed.
+Finally, the old service should be stopped, and the new service should be
+started.
```
ericonr, line 19: "the new one" ? Sorry, final nitpick, otherwise it's ready to merge.

ericonr:
Great, thanks!

PR #241, opened 2020-05-05
README: use local link instead of full GH one

ericonr:
This makes it possible to inspect different branches of void-docs using
the links. It's also shorter.


flexibeast:
i had no idea one could do this, and this is indeed much better. :-) Thanks!

PR #248, opened 2020-05-07
style-guide: Multiple changes.

flexibeast:
* Note need to make first man page reference a link.
* Correct URL for Void man page server.
* Improve phrasing.

@ericonr: Any objections?

### DIFF
--- a/src/contributing/void-docs/style-guide.md
+++ b/src/contributing/void-docs/style-guide.md
@@ -77,12 +77,28 @@ vi /etc/fstab
 
 ## Links
 
+### Internal links
+
+Links to other sections of the Handbook must be relative. For example:
+
+```
+[example](./example.html#heading)
+
+```
+
+and not:
+
+```
+[example](example.html#heading)
+```
+
 ### Man Page Links
 
-Including links to man page is encouraged. These links should point to their man
-page on `https://docs.voidlinux.org`, have their title section number in
-parenthesis, and contain no formatting in their bodies. For example:
-[man(1)](https://man.voidlinux.org/man.1), and not
+The first reference to a command or man page must be a link to the relevant man
+page on `https://man.voidlinux.org/`.
+
+The link text must contain the title section number in parenthesis, and contain
+no formatting. For example: [man(1)](https://man.voidlinux.org/man.1), not
 [`man(1)`](https://man.voidlinux.org/man.1).
 
 ### Auto Links


ericonr:
I'd like to make it clear that it's not only the first mention of a man page, but if it's an application, the first mention of its command too. Like "the `iwctl(1)` command for...", for example. That is, if you agree with that recommendation.

flexibeast:
i do! And in fact, i've just taken that approach with #254. :-) i'll update this PR.

src/contributing/void-docs/style-guide.md
```diff
@@ -79,10 +79,12 @@ vi /etc/fstab
 
 ### Man Page Links
 
-Including links to man page is encouraged. These links should point to their man
-page on `https://docs.voidlinux.org`, have their title section number in
-parenthesis, and contain no formatting in their bodies. For example:
-[man(1)](https://man.voidlinux.org/man.1), and not
+The first reference to a command or man page must be a link to the relevant man
+page on `https://man.voidlinux.org/`. Subsequent references need not be links,
+but making them links is encouraged.
```
ericonr, line 84: Not sure I agree with the subsequent links part. At least, I've been doing the opposite. What do you think? It makes the file more crowded and can make refactoring more complicated than it needs to be.

flexibeast: You mean the "making them links is encouraged" bit? Fair enough. My thinking was that it might be helpful to readers to have a link to follow at the point of mention, rather than having to go back up the page and find one. But i'm not attached to this, so i'm happy to remove this entire sentence if you'd like.

ericonr: I'm never sure of what the most correct course of action would be. Looking at something like https://www.openbsd.org/faq/faq15.html, which could be considered the standard for docs, they link it each time it's mentioned, so leaving it in might be the best option. The docs arent't necessarily made for "offline visualiation", so it makes sense not to prioritize that, if I think about it.

Leaving it in might be for the best, and I can take some time later to undo my changes.

flexibeast:
i've just removed the part about making subsequent references links - i think #254 suggests that this is something that we should consider on a case-by-case basis.

ericonr:
@flexibeast can you add the part about how to format internal links here?

src/contributing/void-docs/style-guide.md
jeffayle: I know this has already been merged, but I'm not sure if there's a better place to bring it up. If we switched to reference style links for man pages it would be trivial to make *all* references links and would reduce clutter overall. eg.

    some text explaining [command(1)]. More talk about [command(1)]... [command(1)]...

    (at the end of the file)
    [command(1)]: man.voidlinux.org/command.1

flexibeast: @jeffayle: i'm certainly open to this. What's the resulting behaviour? If a user was to select the first link, does that take them to the bottom of the page, from which they can then select the man page link? Or would selecting the first link take them to the man page directly?

jeffayle: It would result in the same html, link would go directly to the man page

flexibeast: Sounds good to me. @ericonr, thoughts?

PR #249, opened 2020-05-07
gnome: Use `runsv` instead of `sv once` in example.

flexibeast:

### DIFF
--- a/src/config/graphical-session/gnome.md
+++ b/src/config/graphical-session/gnome.md
@@ -26,5 +26,5 @@ To use a graphical display manager for logging in to GNOME, enable the `gdm`
 service. Test the `gdm` service before enabling it:
 
 ```
-# sv once gdm
+# runsv /etc/sv/gdm
 ```


Duncaen:
Both are kinda bad, starting runsv is not equivalent to `sv once` as it will just loop on failure which I think the person writing the documentation tried to avoid.

Probably best to touch do something like:
```
# touch /etc/sv/gdm/down
# ln -s /etc
```

Vaelatern:
I personally do a `cd` to that directory and `sudo ./run` but I always worry my env is different for doing so.

flexibeast:
@Duncaen:

Hmm, okay. @bobertlo had [suggested](https://github.com/void-linux/void-docs/pull/212/#issuecomment-618811798) the `sv once` form when i was putting together `gnome.md`, as it can be awkward to disabling a display manager service once it's enabled. However, in putting together `kde.md`, i found that `sv once sddm` [resulted in an error](https://github.com/void-linux/void-docs/pull/237#issue-413337832). As a result, @ericonr [suggested](https://github.com/void-linux/void-docs/pull/237#issuecomment-624914604) using `runsv` instead, and since that superficially worked, i put together this PR to make the Handbook consistent.

So from what you're saying, we should change `gnome.md` to use the command set you described. What should we do for `sddm` in #237?

PR #253, opened 2020-05-08
alsa: Link to `apulse` project README for now.

flexibeast:
If and when i-rinat/apulse/pull/112 gets merged, we can then link to `apulse.1` on `man.voidlinux.org` instead.


ericonr:
Hope it's accepted, but releases seem to be slow as well. That link will probably have to remain there for a while.

PR #257, opened 2020-05-09
Add Section Content headers

### LINKED ISSUE #244: ToCs on pages that introduce subsections
A number of Handbook pages are simply introduction pages for subsections. Such pages should, after the introductory text, contain a 'table of contents' listing those subsections. For the purposes of having a consistent style, and to possibly facilitate data transformation (e.g. to address #243), the format of such ToCs should be:

    ## Section Contents

    - [section name](path)
    - [section name](path)

ericonr:
Fixes #244

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -10,7 +10,7 @@
    - [Live Installers](./installation/live-images/index.md)
       - [Prepare Installation Media](./installation/live-images/prep.md)
       - [Partitioning Notes](./installation/live-images/partitions.md)
-      - [Installation Guide](./installation/live-images/guide.md) =======
+      - [Installation Guide](./installation/live-images/guide.md)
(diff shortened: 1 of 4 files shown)


hippi777:
there are links written as "./whatever.md" and "whatever.md", i think these could be unified, but i dont see a straightforward winner...

the 1st is more aligned with paths to different directories, while the 2nd saves some cpu cycles and that delays the melting of the penguins and the ice caps :D so maybe the latter... :D

ericonr:
@hippi777 I prefer `./whatever.md`, tbh, for the reason you mentioned. I hope we don't melt too many penguis because of that, I like penguins :c

flexibeast:
i think i prefer `./whatever.md` myself, but at any rate, it's already used throughout the Handbook. Given that, this preference should mentioned in the style guide.

Issue #89, opened 2019-03-29
Troubleshooting common xbps errors section is only a stub


CameronNemo:
Good topics to cover:

* diagnosing broken shlibs issues
* verifying that the repositories are configured correctly

ericonr:
@CameronNemo what do you mean by checking if repos are configured? I know I can do `xbps-query -L`, but that's as far as I go.

flexibeast:
Asking about the second point on IRC, @nilium suggested:

> Presumably means verifying the xbps config points to the correct repositories for a user's system.

ericonr:
So basically point people to the repositories page so they can check the links? Also show an example of an error when arch repodata can't be found in the repository?

flexibeast:
Yeah, that was my thought. If i change `/usr/share/xbps.d/00-repository-main.conf` to refer to `current/musl` rather than `current`, then do `xbps-install -S`, i get:

    [*] Updating repository `https://alpha.de.repo.voidlinux.org/current/musl/x86_64-repodata' ...
    ERROR: [reposync] failed to fetch file `https://alpha.de.repo.voidlinux.org/current/musl/x86_64-repodata': Not Found

Is that particular error most likely to be caused by having the incorrect URL? If not, i can say that "incorrect URL" *might* be its cause.

ericonr:
I think with "Not Found" in the end it is. When it's something like DNS we get "Transient error...".

This reminds me, should we document the issues that can be caused by having the wrong system time?

flexibeast:
> should we document the issues that can be caused by having the wrong system time?

Definitely; i can include that in my PR. [This thread](https://old.reddit.com/r/voidlinux/comments/hizldf/void_linux_update_issue/) suggests it results in an "Operation not permitted" error, but i just tried adjusting my local clock to be off by 12 hours, and `xbps-install -S` worked just fine, so i can't confirm which error is actually produced.

flexibeast:
Closed by #400.

PR #258, opened 2020-05-09
xbps/common-errors: add section

### LINKED ISSUE #186: Troubleshooting update errors
We have a section on [updating](https://docs.voidlinux.org/xbps/updating.html) Void, but no discussion on how to troubleshoot common error messages when the update fails.

For example, removing orphans with `xbps-remove -o` sometimes resolves the "unresolvable shlibs" error that occurs when attempting `xbps-install -Su`.

The old wiki has some [discussion](https://wiki.voidlinux.org/Frequently_Asked_Questions#A_.27Transaction_aborted_due_to_unresolved_shlibs.27_error_is_preventing_me_from_updating.21) about a related but different problem as well.

ericonr:
Fixes #186 and #89.

### DIFF
--- a/src/xbps/troubleshooting/common-errors.md
+++ b/src/xbps/troubleshooting/common-errors.md
@@ -1,0 +1,54 @@
+# Common errors
+
+## Errors while updating or installing packages
+
+If there are any errors while updating or installing a new package, make sure
+that you are using the latest version of the remote repository index. Running
+[xbps-install(1)](https://man.voidlinux.org/xbps-install.1) with the `-S` option
+will guarantee that.
+
+### shlib errors
+
+If you get an error with the format:
+
+```
+libllvm8-8.0.1_2: broken, unresolvable shlib `libffi.so.6'
+```
+
+it is likely that orphan packages which have already been removed from the Void
+repos are still installed in your system. This can be solved by running
+[xbps-remove(1)](https://man.voidlinux.org/xbps-remove.1) with the `-o` option,
+which removes orphan packages.
+
+If you get an error message saying
+
+```
+Transaction aborted due to unresolved shlibs
+```
+
+the repositories are in the staging state, which can happen due to large builds.
+The solution is to wait for the builds to finish. You can view the builds'
+progress in the [Buildbot's Waterfall
+Display](https://build.voidlinux.org/waterfall).
+
+### repodata errors
+
+In March 2020, the compression format used for the repository data (repodata)
+was changed from gzip to zstd. If XBPS wasn't updated to version `0.54`
+(released June 2019) or newer, it is not possible to update the system with it.
+Unfortunately, there isn't an error message for this case, but it can be
+detected by running `xbps-install` with the `-Sd` flags. The debug message for
+this error is shown below.
+
+```
+[DEBUG] [repo] `//var/db/xbps/https___alpha_de_repo_voidlinux_org_current/x86_64-repodata' failed to open repodata archive Invalid or incomplete multibyte or wide character
+```
+
+In this situation, it is necessary to follow the steps in
+(xbps-static)[./static.md].
+
+## Broken systems
+
+If your system is for some reason broken and can't perform updates or package
+installations, using a [statically linked version of xbps](./static.md) to
+update and install packages can help you avoid reinstalling the whole system.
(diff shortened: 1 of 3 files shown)


src/xbps/troubleshooting/common-errors.md
```diff
@@ -0,0 +1,32 @@
+# Common errors
+
+## Errors while updating or installing packages
+
+If there are any errors while updating or installing a new package, make sure
+that you are using an updated version of the remote repository index. Running
```
flexibeast, line 6: "using the latest version"

src/xbps/troubleshooting/common-errors.md
```diff
@@ -0,0 +1,32 @@
...
+[xbps-install(1)](https://man.voidlinux.org/xbps-install.1) with the `-S` option
+will guarantee that.
+
+### shlib errors
+
+If you get an error with the format
```
flexibeast, line 12: "format: "

src/xbps/troubleshooting/common-errors.md
```diff
@@ -0,0 +1,32 @@
...
+```
+libllvm8-8.0.1_2: broken, unresolvable shlib `libffi.so.6'
+```
+
+It is likely that orphan packages which have already been removed from the Void
+repos are still installed in your system. This can be solved by running
+[xbps-remove(1)](https://man.voidlinux.org/xbps-remove.1) with the `-o` option,
+which removes orphan packages.
+
+If you get an error message saying that
```
flexibeast, line 23: Just "saying: ", i think.

src/xbps/troubleshooting/common-errors.md
```diff
@@ -0,0 +1,32 @@
...
+
+```
+Transaction aborted due to unresolved shlibs
+```
+
+The repositories are in the staging state, which can happen due to large builds.
```
flexibeast, line 29: "the repositories" - no capital, since this is a continuation of the sentence before the example.

flexibeast:
@ericonr: This is looking good, thanks. :-)

Another common issue i think we should mention is https://www.reddit.com/r/voidlinux/comments/gf33tb/cant_update_void_on_fresh_install/, referring people to the `static.md` page.

ericonr:
@flexibeast definitely, having something to link to static should help. What did you think of my explanations?

src/xbps/troubleshooting/common-errors.md
```diff
@@ -0,0 +1,45 @@
...
+```
+Transaction aborted due to unresolved shlibs
+```
+
+the repositories are in the staging state, which can happen due to large builds.
+The solution is to wait for the builds to finish. You can view the builds'
+progress in the [Buildbot's Waterfall
+Display](https://build.voidlinux.org/waterfall).
+
+### repodata errors
```
Chocimier, line 34: Uh, this paragraph gives impression that we break updates every quarter. We don't. Could you change to something like
> In March 2020 compression format of packages was changed from xz to zstd. If xbps wasn't updated to 0.54 (released June 2019) or newer, it's not possible to update system with it. [for searchability: ] Error message in that case is ... It is necessary to follow the steps ..."

, and remove or similarly rewrite and shorten paragraph in static.md?

ericonr: That's definitely not what I meant to imply. Thanks!

ericonr: @Chocimier ok, now I'm not too sure. repodata errors would be due to the repository information compression format having changed, not being able to unpack the packages themsevlves is a different issue. Do you know when the repository data format was changed? I can look for it a bit.

Is there a case XBPS will be able to understand the repo data information but not be able to unpack the packges?

ericonr: The problem with the error message for this issue is that there isn't any. I'm trying it out on a ROOTFS image from 2018, and it doesn't output anything unless you run the commands with the `-d` flag to get some debug output. With debug on, it does give some good output. I can try to include that. Anyway, the issue is with the repodata format more than the package themselves, because the failure happens earlier.

Chocimier: Ah, right, it's repodata compression what causes problems here, not packages. Changes of the two was done same time, so it's _unlikely_ that xbps can handle one but not another.

PR #259, opened 2020-05-09
book.toml: add ntp.org to exclude list

ericonr:
It was causing timeouts in CI, and sometimes locally as well.
Not very reliable.

@flexibeast looks good? I say we wait for CI.


flexibeast:
Yeah, wait for CI.

PR #263, opened 2020-05-10
Session management

ericonr:
Ok, this one is a slight overhaul of the session management stuff. I still don't mention stuff like polkit, but I don't think it's that needed. D-Bus is explained in very rough terms, so any input is welcome for that part.


ericonr:
@travankor I added the dbus dependency for elogind here. Do you think it's well explained?

src/config/session-management.md
```diff
@@ -0,0 +1,32 @@
...
+desktop environments, if launched through an adequate display manager, will
+launch a D-Bus session themselves.
+
+## elogind
+
+[elogind(8)](https://man.voidlinux.org/elogind.8) is a standalone version of
+`systemd-logind`, a service to manage user logins. This service provides
+necessary features for most desktop environments and Wayland compositors. To
+make use of its features, install the `elogind` package and enable the `dbus`
+service. This might require a reboot.
```
travankor, line 24: All you need to do is restart your session AFAIK (ie. go back to `login`). When you authenticate, `elogind` will start.

src/config/session-management.md
```diff
@@ -0,0 +1,32 @@
...
+
+### Troubleshooting
+
+If you're having any issues with `elogind`, [enable](./services/index.md) its
+service, as waiting for a D-Bus activation can lead to issues.
+
+If you have the [acpid(8)](https://man.voidlinux.org/acpid) service enabled,
```
travankor, line 31: Or you can `ignore` the acpid functions in elogind in `/etc/elogind/logind.conf`

src/config/session-management.md
```diff
@@ -0,0 +1,32 @@
...
+a window manager or interactive shell) with
+[dbus-run-session(1)](https://man.voidlinux.org/dbus-run-session.1). Most
+desktop environments, if launched through an adequate display manager, will
+launch a D-Bus session themselves.
+
+## elogind
+
+[elogind(8)](https://man.voidlinux.org/elogind.8) is a standalone version of
+`systemd-logind`, a service to manage user logins. This service provides
+necessary features for most desktop environments and Wayland compositors. To
```
travankor, line 22: it's needed for rootless Xorg (or at least it's one of the solutions)

src/config/session-management.md
```diff
@@ -0,0 +1,32 @@
+# Session and Seat Management
+
+Session and seat management is not necessary for every setup, but it can be used
+to safely create temporary runtime directories, provide access to hardware
+devices, and control system shutdown.
```
travankor, line 5: and for multiseat

src/config/session-management.md
ericonr: I was talking more specifically about dbus itself. That's what @flexibeast experienced, at least, IIRC. `elogind` requires only that you log out and then log back in?

src/config/session-management.md
ericonr: Do you have some idea on how to explain this? I couldn't find an explicit list of what acpid functions were available in `logind.conf(5)`.

src/config/session-management.md
ericonr: Good point!

src/config/session-management.md
ericonr: Also good point, thanks!

src/config/session-management.md
ericonr: I added mostly what you said, if you have some function list we could consider adding it.

src/config/session-management.md
flexibeast: It seems to depend on the context. In my VM testing, i've found that GNOME requires a reboot after enabling `dbus`, but KDE doesn't. i don't recall anything needing to be done after installing `elogind`.

flexibeast: (Oh and i should add: logging out and back in again after enabling `dbus` is not enough for GNOME. i just checked this a couple of hours ago, and it definitely needs a reboot.)

ericonr:
@flexibeast would it be okay with you if I factored out the `dbus-run-session` information from #282 to link to this page instead?

flexibeast:
@ericonr: Oh! Sorry, i forgot about this PR when doing #282. Yes, certainly.

src/config/session-management.md
```diff
@@ -0,0 +1,38 @@
...
+
+Session and seat management is not necessary for every setup, but it can be used
+to safely create temporary runtime directories, provide access to hardware
+devices and multi-seat capabilities, and control system shutdown.
+
+## D-Bus
+
+D-Bus is an IPC (inter-process communication) mechanism used by userspace
+software in Linux. There are two uses for it: as a system bus, and as a user
+bus.
```
flexibeast, line 11: i wonder whether we should say "user bus" or "session bus". The latter is the technical name for it (cf. e.g. the `DBUS_SESSION_BUS_*` env vars), and it might be better for readers if our terminology matches the terminology used elsewhere.

src/config/session-management.md
```diff
@@ -0,0 +1,38 @@
...
+
+- To use it as a system bus, you should [enable](./services/index.md) the `dbus`
+   service. This might require a system reboot to work properly.
+- To use it as a user bus, you can start a given program (usually a window
```
flexibeast, line 15: As above.

src/config/session-management.md
```diff
@@ -0,0 +1,38 @@
...
+   manager or interactive shell) with
+   [dbus-run-session(1)](https://man.voidlinux.org/dbus-run-session.1). Most
```
flexibeast, line 17: Just to check, are there actually any use-cases where `dbus-run-session(1)` *isn't* appropriate, such that users should use `dbus-launch(1)` instead? @abenson?

src/config/session-management.md
travankor: basically everything that can accept "ignore" should all say "ignore". Like the Handle* section, for example.

src/config/session-management.md
ericonr: I used session bus, but also made it clear that it's for a user session. Do you think it's a good explanation?

src/config/session-management.md
```diff
@@ -0,0 +1,38 @@
...
+
+Session and seat management is not necessary for every setup, but it can be used
+to safely create temporary runtime directories, provide access to hardware
+devices and multi-seat capabilities, and control system shutdown.
+
+## D-Bus
+
+D-Bus is an IPC (inter-process communication) mechanism used by userspace
+software in Linux. There are two uses for it: as a system bus and as a session
+bus, the latter being specific to a user session.
```
flexibeast, line 11: "D-BUS can be run as a system bus or as a session bus, the latter ..."

src/config/session-management.md
flexibeast: Thanks, that looks good.

src/config/session-management.md
flexibeast: Er, sorry, i meant @ahesford. But of course @abenson is welcome to chime in as well. :-)

ahesford: `dbus-launch` can be used without the `--exit-with-X11` argument in ways that `dbus-run-session` cannot, like launching a short-lived program but keeping the message bus alive. However, in the context of this document, the choice is really between `dbus-run-session` and `dbus-launch --exit-with-x11`. I don't believe there is any case where the latter would be preferred. For one, the `--exit-with-x11` option obviously requires X11. The `dbus-run-session` command is more general and, in my view, takes a more robust approach to lifetime management. (DBus is kept alive as long as the child, not when some second, single-purpose process loses its connection to an X server.)

src/config/session-management.md
ericonr: What do you think of my phrasing now?

travankor: options related to acpi? acpi is the specification, while acpid is the daemon

ahesford: How about
```
By default, `elogind` listens for and processes ACPI events related to lid-switch activation and presses of power, suspend and hibernate keys. This will conflict with `acpid` if it is installed an enabled. Either disable `acpid` when enabling `elogind` or modify `/etc/elogind/logind.conf` to ignore these events within `elogind`. The `HandlePowerKey`, `HandleSuspendKey`, `HandleHibernateKey` and multiple `HandleLidSwitch` options in [logind.conf(5)](https://man.voidlinux.org/logind.conf.5) should be set to `ignore` to stop `elogind` from responding to ACPI events.
```

ericonr: @travankor "options related to controlling acpi events", possibly.

@ahesford do you think those will remain valid? I don't want to let the docs grow stale.

ahesford: If you want to be agnostic, then
```
Either disable `acpid` when enabling `elogind` or configure `elogind` to `ignore` ACPI events
in `/etc/elogind/logind.conf`. The [logind.conf(5) man page](https://man.voidlinux.org/logind.conf.5)
describes several configuration options, all starting with `Handle`, that should be set to
`ignore` to avoid interfering with `acpid`.
```

ericonr: Awesome! Thank you both :D

src/config/session-management.md
ericonr: In that case, it makes sense to leave this item as is, right?

src/config/session-management.md
ericonr: Mixed it up a bit, I think it looks good now.

src/config/session-management.md
flexibeast: @ericonr: On the basis of @ahesford's comment, yes, i think it can be left as-is.

PR #265, opened 2020-05-10
graphics-drivers/intel: add IOMMU troubleshooting

ericonr:
@jnbr can you take a look at this explanation, please? Just want to make sure it's accurate.

### DIFF
--- a/src/config/graphical-session/graphics-drivers/intel.md
+++ b/src/config/graphical-session/graphics-drivers/intel.md
@@ -1,5 +1,10 @@
 # Intel GPU
 
+Intel GPU support requires the `linux-firmware-intel` package. If you have
+installed the `linux` or `linux-lts` packages, it will be installed as a
+dependency. If you installed a version-specific kernel package (e.g.,
+`linux5.4`), it may be necessary to manually install `linux-firmware-intel`.
+
 ## OpenGL
 
 Install the Mesa DRI package, `mesa-dri`.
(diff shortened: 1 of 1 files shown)


jnbr:
> you should file a bug reporting the problem.

clarify the bug should be reported to upstream, we don't want to deal with them, else lgtm

src/config/graphical-session/graphics-drivers/intel.md
```diff
@@ -24,3 +24,14 @@ environment variable `LIBVA_DRIVER_NAME`:
...
 | `intel-media-driver` | from Broadwell    | `LIBVA_DRIVER_NAME=iHD`  |
+
+## Troubleshooting
+
+The kernels packaged by Void are configured with
+`CONFIG_INTEL_IOMMU_DEFAULT_ON=y`, which can lead to issues with their graphics
+drivers, as reported by the [kernel
+documentation](https://www.kernel.org/doc/html/latest/x86/intel-iommu.html#graphics-problems).
+To fix this, it is necessary to disable IOMMU for the integrated GPU. This can
+be done by adding `intel_iommu=igfx_off` to your [kernel
```
travankor, line 35: I have only noticed this workaround for Intel Generation 5 (Broadwell).

ericonr: Mine (Skylake - don't remember the graphics gen) locks up, but hasn't done so again since I switched on guc firmware. It's hard to determine where the issues come from.

How many devices do you deal with? Do you think your experience is anecdotal or the expected one? The kernel docs didn't suggest any particular gen.

travankor: Well, it is broken _in the hardware_ for Broadwell, meaning you will have to enable this if you have 5th gen. For other versions, it could be firmware or something else, so I'm not sure.

travankor: I didn't have problems with gen 7 and gen 3, but you're right this is all anecdotal on my side.

travankor: https://www.phoronix.com/scan.php?page=news_item&px=Intel-IOMMU-Gfx-Default-Try

ericonr: I left it as a generic option, but which is recommened for Broadwell. What do you think?

travankor: Seems reasonable. Maybe mention that you need `linux-firmware-intel`, somewhere on this page.

ericonr: It's a dependency for the `linux` and `linux-lts` packages. Do you think it needs to be?

travankor: Fair enough

ericonr: Ended up adding it :p Thanks for the suggestion!

ericonr:
@flexibeast can I merge this one?

flexibeast:
@ericonr: LGTM - merged.

PR #269, opened 2020-05-10
config/graphical-session/xorg: Remove catalyst

st3r4g:
`catalyst` was removed from the repo some months ago


ericonr:
Hmm that table will probably be replaced by the nvidia page. This works for now, though. Thanks!

PR #271, opened 2020-05-11
gnome: minor changes

not-chicken:


src/config/graphical-session/gnome.md
```diff
@@ -2,30 +2,52 @@
...
 
-A minimal GNOME environment can be created by installing the `gnome-session`,
-`gdm` and `adwaita-icon-theme` packages. (Note, however, that not all GNOME
-features may be present or functional.) `gdm` defaults to providing a Wayland
-session, via the `mutter` window manager. For an Xorg session, install the
-`xorg` package, then select 'GNOME on Xorg' at the GDM login screen, or start
-`gnome-session` via `~/.xinitrc`. GNOME applications can be installed via the
-`gnome-apps` package.
+A minimal GNOME environment is available via the `gdm` package. (Note, however,
+that not all GNOME features may be present or functional.)
```
ericonr, line 17: I think you should include `gnome-session` and the `adwaita-icon-theme` here as well, right?

src/config/graphical-session/gnome.md
```diff
@@ -2,30 +2,52 @@
...
+
+For an Xorg session, add the following to your `~/.xinitrc` file:
+
+```
+export XDG_SESSION_TYPE=x11
+export GDK_BACKEND=x11
+exec gnome-session
+```
+
+Then call `startx`.
```
ericonr, line 53: "Then, call"

src/config/graphical-session/gnome.md
```diff
@@ -2,30 +2,52 @@
...
+A minimal GNOME environment is available via the `gdm` package. (Note, however,
+that not all GNOME features may be present or functional.)
 
 If you require [ZeroConf](http://www.zeroconf.org/) support, install the `avahi`
 package and enable the `avahi-daemon` service.
 
-## Display Manager
+## Starting GNOME
+
+### From Display Manager
```
ericonr, line 24: What do you think about using "From the Display Manager" and "From the Console"? I'm not sure they are better, but they might be.

src/config/graphical-session/gnome.md
ericonr, line 53: Oh cool, these work?! Just to be sure, did you have to do any other setup for them?

src/config/graphical-session/gnome.md
not-chicken: We are adding them as rundeps to `gdm` in 3.36 update.

src/config/graphical-session/gnome.md
not-chicken: They did work as is while testing in a vm.

src/config/graphical-session/gnome.md
not-chicken: Sounds better with `the` imo

flexibeast:
@not-chicken: Thanks for your work on this!

i just wanted to let you know that, in addition to other recent changes to `gnome.md`, there's probably going to be at least one more change in the next few days, as i factor out some stuff into `xorg.md` - i'll let you know once that's done, so you can rebase.

flexibeast:
Blocked on https://github.com/void-linux/void-packages/pull/21382 and https://github.com/void-linux/void-packages/pull/22936.

ericonr:
I guess this can be updated, now.

not-chicken:
Updated. I've removed the "Starting from console" since that wasn't void specific and was copy pasted by me from archwiki :P

src/config/graphical-session/gnome.md
```diff
@@ -2,27 +2,28 @@
...
 ## Installation
 
-Install the `gnome` package for a GNOME environment which includes GNOME
-applications.
+Install the `gnome` package for a GNOME environment which includes the base
+GNOME desktop and a subset of GNOME applications. Additional applications are
+available via the `gnome-apps` package.
+
+A minimal GNOME environment can be created by installing the `gnome-core`
+package. (Note, however, that not all GNOME features may be present or
```
ericonr, line 18: I think we can remove the parentheses here, what do you think? Still leave the content, just remove the parentheses themselves.

not-chicken: Removed.

src/config/graphical-session/gnome.md
```diff
@@ -2,27 +2,28 @@
...
-an X session can be chosen instead.
-
-If you wish to start an X-based GNOME session from the console, use
-[startx](./xorg.md#startx) to run `gnome-session`.
-
-GNOME applications can be installed via the `gnome-apps` package.
-
-If you require [ZeroConf](http://www.zeroconf.org/) support, install the `avahi`
-package and enable the `avahi-daemon` service.
+an Xorg session can be chosen instead.
```
ericonr, line 29: I tried making this change once, but using `X` is the correct name, since `Xorg` is "only" an implementation.

src/config/graphical-session/gnome.md
```diff
@@ -2,27 +2,28 @@
 
 ## Pre-installation
 
+GNOME supports both Xorg and Wayland sessions. Follow [Wayland](./wayland.md),
```
ericonr, line 5: `Follow the "[Wayland]()" or "[Xorg]()" sections...`

src/config/graphical-session/gnome.md
not-chicken: Reverted. I was trying to make it consistent with "Gnome on Xorg" option of GDM.

ericonr:
Merged with a small correction, thanks!

Issue #268, opened 2020-05-10
Post Installation


Duncaen:
The [Post Installation](https://docs.voidlinux.org/config/post-install.html) section contains nothing worthwhile, I think it would make sense to remove this page and add a note to the installation guides to update the system after installing if necessary.
Also not really configuration related, if someone wants to keep a page like this I think even if its "Post Installation", it would still make sense to keep it in the "Installation" section.

flexibeast:
Closed by https://github.com/void-linux/void-docs/commit/a1469911e6399bfd28b70513a6d5b88e640e50de.

PR #273, opened 2020-05-11
post-install,guide: Multiple changes.

flexibeast:
Addresses #268.


ericonr:
Technically speaking, a chroot installation using a ROOTFS image also requires an update. It probably makes more sense to include that in the advanced guide, though.

flexibeast:
@ericonr: Good point. How about just "After completing installation, be sure ..."?

ericonr:
I like the way you wrote, because "completing installation" might end up with someone updating the live system :p . I'd leave it as is.

flexibeast:
Ah okay. Well, if there's nothing else, would you be able to merge this, please?

PR #274, opened 2020-05-11
gnome: Modify example for testing that `gdm` works.

flexibeast:
As recommended by @Duncaen [here](https://github.com/void-linux/void-docs/pull/249#issuecomment-625428886).

### DIFF
--- a/src/config/graphical-session/gnome.md
+++ b/src/config/graphical-session/gnome.md
@@ -24,8 +24,5 @@ package and enable the `avahi-daemon` service.
 ## Display Manager
 
 To use a graphical display manager for logging in to GNOME, enable the `gdm`
-service. Test the `gdm` service before enabling it:
-
-```
-# runsv /etc/sv/gdm
-```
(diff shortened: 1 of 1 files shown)


ericonr:
I'd prefer to have this in a single section under `services/`, and link to that. Perhaps mention that if they disable the service they should remove the `down` file as well? Otherwise they might forget it when trying to enable the service again.

flexibeast:
@ericonr: So make creation of such a new section a blocker for #237?

ericonr:
If you're up to writing it, I don't have an issue with it being included in that PR. If not, then yes, a blocker. I can try to get to it, if you can't.

flexibeast:
Sure, i'll see what i can do. :-)

ericonr:
@flexibeast perhaps even merge this PR into that one? That way the change comes in one piece.

ericonr:
So this one can just become a link to that page, right?

PR #277, opened 2020-05-12
services/managing: Add information about testing a service.

flexibeast:


src/config/services/managing.md
```diff
@@ -70,6 +70,18 @@ currently running:
 # rm /etc/runit/runsvdir/default/<service>
 ```
 
+### Testing services
+
+To check a service works as expected before fully enabling it:
```
ericonr, line 75: Perhaps explain that this is necessary for the service to run in the environment it is expected to run?

src/config/services/managing.md
```diff
@@ -70,6 +70,18 @@ currently running:
...
+
+```
+# touch /etc/sv/<service>/down
+# ln -s /etc/sv/<service> /var/service
+# sv once <service>
+```
+
+If everything works, remove the `down` file to enable the service.
```
ericonr, line 83: I'd like to have some advice that they should remove the service either way, otherwise they will forget about it when trying to enable the service some time from now.

src/config/services/managing.md
flexibeast: Hmm, yes. Any suggested wording?

src/config/services/managing.md
flexibeast: Sorry, i don't follow? The use-case i'm thinking of is: "I want to enable this service, but I need to check it works first. Yes, it does. Okay, I can remove `down`, which will leave the service enabled."

ericonr: Your use case is the "happy path", so to speak. If the service has any issues, and the user decides to simply drop it, they should "remember to remove the `down` file as well, otherwise enabling the service at a later time won't actually start it"

src/config/services/managing.md
ericonr: "To check if a service is working correctly when started by the service supervisor, it is possible to run it once, before fully enabling it:"

flexibeast: Would you be okay with: "To check if a service is working correctly when started by the service supervisor, run it once before fully enabling it:"?

src/config/services/managing.md
flexibeast: Ah, okay. How about something like: "If everything works, remove the \`down\` file to enable the service. If not, and you don't wish to enable the service, remove both the \`down\` file and the symlink in \`/var/service/\`."?

src/config/services/managing.md
ericonr: yup!

flexibeast: Done. :-)

src/config/services/managing.md
ericonr: Hmm, this is on the same page that has all the other information about service management. I re-read it and perhaps this warning isn't really necessary.

ericonr:
In general, I think the header is in the correct position too.

src/config/services/managing.md
flexibeast: So leave it as-is? If so, i'll force-push the other change, then i guess it's ready for you to merge. :-)

ericonr: I think so, yeah!

flexibeast:
@ericonr: PR updated, and ready to be merged. Thanks for your review!

PR #177, opened 2020-04-07
installation via chroot guide

jeffayle:
This guide is not quite complete. I would like feedback on what is and isn't included, and whether any steps are confusing (or wrong).

There are three areas in particular that I am unsure of (they are currently marked in the text with TODO)

- what is the purpose of the base-voidstrap package, when would you choose it over base-minimal and when would you choose it over base-system ? (are there other metapackages that I'm unaware of someone might want to use instead?)
- installing grub for uefi can be done with only `grub-install`. The guides in the wiki include a number of switches that are either default or automatically configured. are any of these automatic configurations liable to break, should the switches be included in the instructions?
- xbps-reconfigure is used to run kernel post-install hooks to make dracut and grub work. is there a way to do this without having to lookup the proper package to reconfigure?

PR #278, opened 2020-05-13
users and groups: basic tasks to manage users and groups

jeffayle:
This pull request expands the users and groups page to cover basic tasks like creating and deleting users.

I've also changed "Users and groups" to "Users and **G**roups".

### DIFF
--- a/src/config/users-and-groups.md
+++ b/src/config/users-and-groups.md
@@ -1,6 +1,37 @@
-# Users and groups
+# Users and Groups
 
-## Default groups
+The [useradd(8)](https://man.voidlinux.org/useradd.8),
+[userdel(8)](https://man.voidlinux.org/userdel.8) and
+[usermod(8)](https://man.voidlinux.org/usermod.8) commands are used to add,
+delete and modify users respectively. The
+[passwd(1)](https://man.voidlinux.org/passwd.1) command is used to change
+passwords.
+
+The [groupadd(8)](https://man.voidlinux.org/groupadd.8),
+[groupdel(8)](https://man.voidlinux.org/groupdel.8) and
+[groupmod(8)](https://man.voidlinux.org/groupmod.8) commands are used to add,
+delete and modify groups respectively. The
+[groups(1)](https://man.voidlinux.org/groups.1) command lists all groups a user
+belongs to.
+
+## sudo
+
+> Note: [sudo(8)](https://man.voidlinux.org/sudo.8) is installed by default but
+> may not be configured. It is only necessary to configure sudo if you wish to
+> use it.
+
+Use [visudo(8)](https://man.voidlinux.org/visudo.8) as root to edit the
+[sudoers(5)](https://man.voidlinux.org/sudoers.5) file.
+
+To create a superuser, uncomment the line
+
+```
+#%wheel ALL=(ALL) ALL
+```
+
+and add users to the `wheel` group.
+
+## Default Groups
 
 Void Linux defines a number of groups by default.
 
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -16,7 +16,7 @@
 - [Configuration](./config/index.md)
    - [musl](./config/musl.md)
    - [Locales](./config/locales.md)
-   - [Users and groups](./config/users-and-groups.md)
+   - [Users and Groups](./config/users-and-groups.md)
    - [Services And Daemons](./config/services/index.md)
       - [Managing Services](./config/services/managing.md)
       - [Configuring Services](./config/services/configuring.md)


src/config/users-and-groups.md
```diff
@@ -1,6 +1,104 @@
...
+
+This configuration is automatic if no root password is set during an interactive
+installation.
+
+### Lock & Unlock a Password
+
+A user can be prevented from logging into the system by locking their password.
+
+```
+# passwd -L username
```
Duncaen, line 68: this is `-l`

Duncaen:
I feel like this is a bit too verbose, I think it would be enough to provide a short section like "user accounts can be created with useradd(1), modified with usermod(1) and deleted with userdel(1). Passwords can be changed, locked and unlocked with passwd(1).", I don't think its really necessary to show examples for every single tool and/or flags for passwd.

Also don't really think the sudo wheel group stuff belongs here.

jeffayle:
> Also don't really think the sudo wheel group stuff belongs here.

is there somewhere else in the docs you think it belongs? I figured it should be in the docs since sudo comes with void, but if you do a chroot install or set a root password in the interactive installer you end up with an unconfigured sudo.

flexibeast:
Yeah, as the ["About this Handbook" section](https://docs.voidlinux.org/about-handbook/index.html) of the Handbook notes, the Handbook is not intended to be a general guide to using Linux:

> This handbook is not an extensive guide on how to use and configure common Linux software. The purpose of this document is to explain how to install, configure, and maintain Void Linux systems, and to highlight the differences between common Linux distributions and Void.

So the most we should add here is the sort of text @Duncaen described, and the `sudo` stuff doesn't need to mentioned in the Handbook at all - not everyone is going to need to specifically configure it (i certainly haven't), and anyway, a number of users don't use `sudo`, but `doas` instead.

jeffayle:
okay, I've removed everything except a list of commands with links to man pages.

flexibeast:
@jeffayle: Thanks, looks good.

@ericonr: Any objections to me merging this?

ericonr:
Well, this is weird, but I think the sudo part should be mentioned here. As @jeffayle wrote, sudo comes as part of base-system for base users, and they should have an idea of what they need to setup for the basic sudo functionality, at least. At least link to `sudo(1)` and possibly `sudoers(5)`, IMHO.

flexibeast:
@ericonr: i asked @Duncaen about this on IRC, and his response was:

> I guess we could add a section about sudo in this page that  references the man pages and mentions that there is a rule that can be uncommented to allow users of the wheel group
>
> not sure if its worth it to have a separate page for that i guess
>
> in general, a lot of pages are really short, which I think is not really a nice reading experience 
>
> so adding that to users and groups seems ok if it would otherwise be a 2 sentence page

jeffayle:
Something like

> Use visudo(8) to configure sudo(8).

Should I leave the instruction of what line to uncomment to reproduce the configuration you get from the live installer, or just leave it at this?

ericonr:
@jeffayle I would add the specific line, making it clear that if you don't want to, you can do it some other way.

@flexibeast definitely not another page! I think this way works well :)

jeffayle:
Okay, I added the lines about sudo but now I'm starting to think this should go into #177 instead of here because it describes how to do something that's done already if you'd used the interactive installer. This section here would be replaced with a note that sudo is installed by default, and if you want to *stop* it from granting superuser privileges, do this.

ericonr:
@jeffayle we had actually talked about this in that PR, and I had thought including it in users and groups made more sense. It's fair to point from that page to this one, perhaps, but I think it fits better here.

jeffayle:
oh right, that's where this PR came from in the first place.

how is what I have now on sudo? is it too much?

src/config/users-and-groups.md
```diff
@@ -1,6 +1,37 @@
...
+> Note: [sudo(8)](https://man.voidlinux.org/sudo.8) is installed by default but
+> may not be configured. It is only necessary to configure sudo if you wish to
+> use it.
+
+Use [visudo(8)](https://man.voidlinux.org/visudo.8) (as root) to edit the sudo
+configuration file.
+
+To create a superuser, find the line `#%wheel ALL=(ALL) ALL` in the
+configuration file and uncomment it by removing the `#`. Add any users you wish
+to make superusers to the `wheel` group.
```
ericonr, line 28: I prefer simply "uncomment the line" and show it in a code block like

```
# %wheel ALL=(ALL) ALL
```

src/config/users-and-groups.md
```diff
@@ -1,6 +1,37 @@
...
+
+To remove superusers, remove them from the `wheel` group. To disable the group
+altogether, use visudo to comment out the aforementioned line by adding `#` at
+the beginning.
```
ericonr, line 32: I think this paragraph is a logic conclusion of the former, so it doesn't have to be included.

src/config/users-and-groups.md
jeffayle: yeah, that's better

src/config/users-and-groups.md
jeffayle: yeah, that's pretty straight forward

ericonr:
Awesome! Looks good to me.

src/config/users-and-groups.md
```diff
@@ -1,6 +1,37 @@
...
+belongs to.
+
+## sudo
+
+> Note: [sudo(8)](https://man.voidlinux.org/sudo.8) is installed by default but
+> may not be configured. It is only necessary to configure sudo if you wish to
+> use it.
+
+Use [visudo(8)](https://man.voidlinux.org/visudo.8) (as root) to edit the sudo
+configuration file.
```
flexibeast, line 24: Since there's more than one config file related to `sudo` (`/etc/sudo.conf` as well as `/etc/sudoers`), i'd prefer this say:

"Use \[visudo(8)\](https://man.voidlinux.org/visudo.8) as root to edit the \`/etc/sudoers\` file."

flexibeast:
@jeffayle, @ericonr: Thank you both for your work on this. Other than [one small change](https://github.com/void-linux/void-docs/pull/278#discussion_r426219860), i think this is ready to merge.

src/config/users-and-groups.md
jeffayle: how about "use [visudo(8)] (as root) to edit the [sudoers(5)] file.", since that file has a manpage?

flexibeast: You're right, that's better. :-) But my preference is to not have parentheses around "as root".

jeffayle: sure thing

flexibeast:
@ericonr: i think this is now ready to merge. Any objections?

ericonr:
None! Thanks :D

PR #280, opened 2020-05-16
wpa_supplicant: Remove extraneous example text.

flexibeast:


ericonr:
I don't use enough wpa_supplicant to know what parts of that are ok to remove. Seems fine to me (and much better), though.

flexibeast:
@ericonr: Thanks. :-) i've just merged this; we can always address any issues later.

PR #285, opened 2020-05-19
style-guide: Multiple changes.

flexibeast:
* Further specify link text formatting.
* Mention use of colon prior to examples.
* Describe appropriate usage of notes.
* Fix some incorrect capitalisation.
* Remove unnecessary vertical space.

This PR doesn't (yet) address whether we should specify use of "Void" or "Void Linux". From `void-docs/src`:

    $ ag -c 'Void' | cut -d':' -f2 | awk '{s+=$1} END {print s}' 
    104
    $ ag -c 'Void Linux' | cut -d':' -f2 | awk '{s+=$1} END {print s}'
    30

### DIFF
--- a/src/contributing/void-docs/style-guide.md
+++ b/src/contributing/void-docs/style-guide.md
@@ -63,13 +63,13 @@ For example:
 # vi /etc/fstab
 ```
 
-And not:
+and not:
 
 ```
 $ sudo vi /etc/fstab
 ```
 
-And also not:
+and also not:
 
 ```
 vi /etc/fstab
@@ -77,20 +77,37 @@ vi /etc/fstab
 
 ## Links
 
-### Internal links
-
-Links to other sections of the Handbook must be relative. For example:
+Link text should not include sentence-level punctuation. For example:
 
 ```
-[example](./example.html#heading)
-
+[Visit this site](https://example.org).
 ```
 
 and not:
 
 ```
-[example](example.html#heading)
+[Visit this site.](https://example.org)
 ```
+
+### Internal links
+
+Links to other sections of the Handbook must be relative. For example:
+
+```
+[example](./example.md#heading-text)
+```
+
+and not:
+
+```
+[example](example.md#heading-text)
+```
+
+Command code-blocks should be introduced with a colon (':'), i.e.:
+
+> For example:
+> 
+> `$ ls -l`
 
 ### Man Page Links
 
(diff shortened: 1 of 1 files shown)


src/contributing/void-docs/style-guide.md
```diff
@@ -63,27 +63,38 @@ For example:
...
+```
+[Visit this site.](https://example.org)
+```
+
 ### Internal links
 
 Links to other sections of the Handbook must be relative. For example:
 
 ```
 [example](./example.html#heading)
```
jeffayle, line 97: (from your last style-guide PR) should this be example.md#heading ? Most current links point to .md files

flexibeast: i wasn't sure whether, when referring to anchors in links, we need to specify '.html' or '.md' in the source; istr some places in the doc sources specifying '.html'. If `./example.md#heading` definitely works, yes, we should certainly change it (and i'll do so in this PR).

jeffayle: yes, you can find an internal link to .md with a fragment in src/config/microcode.md.

flexibeast: @jeffayle: i've just tried building with links that use '.md' instead of '.html', and that seems to work, so i've updated this PR accordingly. Thanks for reminding me to check this!

ericonr:
What *would* be the cases for quotes, then? Or do you think we should fully remove them?

```
> Something
```

flexibeast:
@ericonr: Semantically, blockquotes should be used to show either actual or possible text.

For example:

    The manual for program X states that:
 
    > This option should be used with caution!

or:

    Make sure each sentence ends with a full stop:

    > This is a complete sentence.

i feel it shouldn't be used as a styling mechanism.

If we want to bring attention to a particular note, we could use bolding: `**Note:**`

flexibeast:
And yes, i feel we should discourage the use of "Note" text; if the information is important, it shouldn't be just a note, and if it's not, it doesn't need to be in a callout that interrupts the general flow of text. (i found an ORA book on Python to be particularly bad in this regard.)

ericonr:
I see your point. What do you think should be the policy in this regard, then?

flexibeast:
@ericonr: Do you mean, policy for use of blockquotes? If so, then i think something along the lines of: "Blockquotes should only be used when quoting text from another source." i use "should" here to allow us some leeway, but i think that probably covers it.

ericonr:
More in the sense of the places where are already using block quotes. Should we go on a fixing spree or leave it as something that will be fixed as pages are are modified?

flexibeast:
@ericonr: Oh i see. No, let's not go on a fixing spree; i think at this point we should focus on adding new content and improving existing content, rather than making stylistic modifications. That said, maybe what we can do is create an issue "Ensure Handbook follows style guide", and label it as "good first issue" - it could be a good way to get new contributors involved.

ericonr:
Im in favor of that.

flexibeast:
@ericonr: Great. :-) i'll open that issue shortly.

i've just updated this PR with a bit about appropriate usage of block quotes. Okay to merge?

ericonr:
Looks good! Thanks :D

PR #288, opened 2020-05-19
src/theme/index.hbs: skip to content link skips table of contents

jeffayle:
The skip to content link currently points to the beginning of the table of contents. it should point to the beginning of the article.


flexibeast:
Nicely spotted. Thanks!

PR #294, opened 2020-05-20
src/theme/css/default.css: fix design for 700-890px

PaperMountainStudio:
Before this commit, if you resized your window to be 700-890px wide, you would get all the text in a narrow column on the left. This fixes it, but I am not sure if I fixed the root of the problem or if this is only a workaround.

### DIFF
--- a/src/theme/css/general.css
+++ b/src/theme/css/general.css
@@ -305,7 +305,7 @@ body {
 }
 
 /* whatever, just hide the nav */
-@media only screen and (max-width: 700px) {
+@media only screen and (max-width: 890px) {
 	#void-nav ul.right {
 		display: none;
 	}


ericonr:
I tested this locally and it looks good (and works much better, thanks!), so I will see if we have someone who can verify it or suggest some other workaround. I don't know much about CSS.

SevereOverfl0w:
The real problem here is that we're using javascript (somewhere) to determine whether the sidebar should be visible by default.  That javascript is not responding to the screen size changing.  If you resize your browser to 750px and then open the site from fresh, you'll notice no such problems.

SevereOverfl0w:
As an aside, it seems that mdbook has a similar problem (although not exactly the same).  https://rust-lang.github.io/mdBook/ resize to 400 from a wider browser.

ericonr:
Hmm yeah, it's not hiding the sidebar on that one either.

PaperMountainStudio:
This is a hack, proper solution by @SevereOverfl0w is in #298

PR #295, opened 2020-05-20
theme/css/general.css: use dark theme if prefered

PaperMountainStudio:
This PR uses the new prefers-color-scheme media query to determine if user's OS is set to dark mode (or browser is manually set to dark mode) and uses dark theme.

Let me know if I should improve something.

### DIFF
--- a/src/theme/css/general.css
+++ b/src/theme/css/general.css
@@ -325,11 +325,11 @@ body {
 	}
 
 	code {
-		background: #264434;
+		background: inherit;
 	}
 	pre {
-		background: #264434;
-		border: 1px solid #192d22;
+		background: #514749;
+		border: 1px solid #222;
 	}
 
 	blockquote {
@@ -337,13 +337,13 @@ body {
 		border: 1px solid #222;
 	}
 	table td {
-		border: 1px #353535 solid;
+		border: 1px #3c3c3c solid;
 	}
 	table thead {
-		background: #353535;
+		background: #3c3c3c;
 	}
 	table tbody tr:nth-child(2n) {
-		background: #353535;
+		background: #3c3c3c;
 	}
 
 	/* nav */
--- a/src/theme/css/general.css
+++ b/src/theme/css/general.css
@@ -310,3 +310,74 @@ body {
 		display: none;
 	}
 }
+
+@media (prefers-color-scheme: dark) {
+	body {
+		color: #ccc;
+		background-color: #333;
+	}
+	h1, h2, h3, h4, h5, h6 { color: #ccc; }
+	a {
+		color: #62b086;
+	}
+	a:hover {
+		color: #ccc;
+	}
+
+	code {
+		background: #264434;
+	}
+	pre {
+		background: #264434;
+		border: 1px solid #192d22;
+	}
+
+	blockquote {
+		background: #514749;
+		border: 1px solid #222;
+	}
+	table td {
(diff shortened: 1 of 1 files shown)


Vaelatern:
Cool, I think we should merge as-is unless someone has a substantive issue.

ericonr:
@Vaelatern have you tested it locally? I don't use dark mode too much so I don't have any stakes on this. It looks reasonable for me, and could make life better for those who do use dark mode in other places.

PaperMountainStudio:
I am not sure about 2 things:
1) is the background dark enough? I have seen pages looking like this, but I am not sure if other people will find it sufficient.
2) I am not sure about the blockquote design, it seems a little bit out of place. This is the best I came up with though.

If someone is a designer, I will be happy to change the colors.

ericonr:
For blockquote, did you consider any other color? I have to agree, the green looks somewhat weird. I'm trying to think of some color that would be compatible with the current style.

PaperMountainStudio:
I tried using the color which is already in the light theme, but making it lighter or darker. I think you are talking about the code blocks though. That's one more thing I am not happy with. Blockquote is the light grey rectangle.

ericonr:
Yes, sorry! The blockquotes look good to me, I think the issue is with how it matches the code block inside it.

PaperMountainStudio:
preview2:

jeffayle:
The two boxes on the page are a blockquote and a pre---I don't think those two elements should be styled so similarly. I think there should also be some way to identify <code> visually other than the monospaced font.

Vaelatern:
tweak: can you please make the header bar color darker? Closer to the code block color. When using dark mode I do like when the website is actually dark.

PaperMountainStudio:
preview3:
preview 3

@Vaelatern I made all the colors a bit darker. I don't want to go with even darker color for the nav bar, because it would not fit in the original colorscheme. Do you think this is enough?

ericonr:
I like this last version! @Vaelatern can you give it a last look, please?

hippi777:
hi folks! :) its really neat, but how can i check it out in action?!!! :D thx!

ericonr:
@hippi777 you have to clone this repository, install `mdBook` and run `mdbook serve` inside it. It will tell you what you need to open in your browser to access this (I believe it's `localhost:3000`).

hippi777:
ahh, thx 4 ur effort, but actually i couldnt figure out where the switch was hidden on the docs.void pages and i only wanted that... :D will it be there, or this will be the only way?

PaperMountainStudio:
A switch is not something I personally need, but I might work on it in
the near future. Currently, your browser has to be set to dark mode, if
you are using firefox, use this:
1. Go to about:config
2. Add a new variable ui.systemUsesDarkTheme of type number with value 1

And all pages supporting this media query will change to dark. I think 
firefox will make this accessible from the settings in next versions or
they will read if the OS is set to dark on Linux (they do on Mac and
Windows).

hippi777:
WOW THX!!!! :D

ericonr:
@flexibeast can this be merged? Or should it go in the `book.css` file you mentioned in #298 ?

flexibeast:
@ericonr: No, i don't think that needs to be a blocker here. i'll open a separate issue for that shortly, and merge this PR now.

ericonr:
Cool! Thanks @PaperMountainStudio !

PR #296, opened 2020-05-21
session-management: Clarify info about system bus and session bus.

flexibeast:


flexibeast:
No comments/objections, so i'm merging this.

PR #298, opened 2020-05-21
Fix size issues when navbar grows

SevereOverfl0w:
There's 2 aspects to these changes:

1. Making the docs work OK on a size that currently doesn't.  There's
some kind of overflow with the navbar, which causes the width to get
very small.
2. Making the navbar fully responsive with the left/right side.

The navbar will now automatically wrap as the screen can no longer
accommodate the right hand side.

*Before submitting a pull request, please read all the sections of the Handbook listed in [CONTRIBUTING](./CONTRIBUTING.md).*

### DIFF
--- a/src/theme/css/general.css
+++ b/src/theme/css/general.css
@@ -84,17 +84,17 @@ svg {
 	padding: 1em;
 }
 
-.right {
-	float: right;
-}
-
 /* void navigation */
 
 #void-nav {
 	width: 100%;
-	height: 50px;
+	min-height: 50px;
 	background: #478061;
 	font-size: 14px;
+
+	display: flex;
+	flex-direction: row;
+	flex-wrap: wrap;
 }
 #void-nav button,
 #void-nav label {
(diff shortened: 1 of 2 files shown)


flexibeast:
CSS changes should probably not be made by modifying `general.css`; my understanding is that they'll get clobbered when we update `mdBook`. According to [the mdBook documentation](https://rust-lang.github.io/mdBook/format/theme/index.html), it seems that `book.css` should be modified instead. But are these changes something that should preferably be made [upstream](https://github.com/rust-lang/mdBook)?

SevereOverfl0w:
My understanding is that these styles are mostly custom to us due to the navbar.

SevereOverfl0w:
I don't see a book.css?

flexibeast:
@SevereOverfl0w: Yeah, `book.css` would need to be created, and local customisations should be moved there. However, i'd like to get your changes in place as soon as possible, so if you could put them in `book.css`, i'll open a separate issue about moving other local customisations to that file.

flexibeast:
@SevereOverfl0w: i'd like to merge this, and move things to `book.css` later (cf. #323). Could you please resolve the conflicts?

SevereOverfl0w:
Sorry, dropped the ball on this.  I'll try and get to it today or tomorrow.

SevereOverfl0w:
PR is updated with the changes you requested.

flexibeast:
Great, thanks!

Issue #267, opened 2020-05-10
musl in configuration section.


Duncaen:
* I feel like this is more relevant to the installation section, there is not a single thing that would be related to configuring already installed systems.
* Examples on how to create void linux chroots don't really belong here, I think it would make sense to add a chroot page to the installation section and link from the musl page to it.

flexibeast:
There's currently a PR for a chroot installation guide: https://github.com/void-linux/void-docs/pull/177. Once that's finalised, @ericonr is planning to work on further installation-related content that leverages that page.

That PR also has a [discussion](https://github.com/void-linux/void-docs/pull/177#pullrequestreview-403985709) about the appropriate `mount` commands; istr someone saying that setting up a glibc chroot on an existing musl system is a different situation than setting up a chroot for installation. If that is indeed the case, then it seems to me that the chroot content could maybe be retained; if not, then it could be removed.

At any rate, we can certainly move this page to the "Installation" section.

Duncaen:
I was thinking more about a "How to setup Void Linux container/chroot rootfs", which I think is useful and quite different from installing void linux on bare metal using a chroot.

travankor:
Some musl-related things that could be documented:

- gcompat
- utmps (utmp/wtmp for musl)
- libexecinfo (backtraces)
- [musl-locales](https://gitlab.com/rilian-la-te/musl-locales)

flexibeast:
@travankor:

* re. `gcompat`, it wasn't included in the current `musl` page due to @q66's comment [here](https://github.com/void-linux/void-docs/pull/163#discussion_r395958344).
* re. `utmps`, it's not currently packaged, but i've starting putting together a package for it, though i've not yet submitted a PR - if i created a WIP PR, would you be able to help with that?

travankor:
@flexibeast Yes, I can help package utmps. I'm a little busy in the next 2-3 days, though, so maybe wait a few days.

flexibeast:
@travankor: Thanks! i've just opened a WIP PR, https://github.com/void-linux/void-packages/pull/21869, but certainly no rush, as i've got lots of other things on my plate myself. :-)

ericonr:
@flexibeast regarding gcompat for musl, I know of at least one case where it works perfectly. The toolchain that `arduino-cli` downloads is linked to glibc but works in a musl system as long as you have gcompat. I recently added it as a dependency, even. Therefore, I believe it might make sense to mention it, because it can work.

[...]

Issue #85, opened 2019-03-09
Advanced installation guides


bobertlo:
I plan to start working on this, but am looking for input on how to handle this. There are SO many installation guides on the wiki, with SO much duplication. My proposal is to create a fleshed out "advanced installation" guide, along with other specific guides, which will only offer differences from the main guide.

### Advanced Installation Guide
This guide will detail "manual" installation from start to finish, more in the style of the "live image installer" guide. This guide would be formal and fleshed out, mentioning alternatives to the given procedures. Topics would include:

- preparing and mounting partitions
- manually installing a system with `xbps-install`
- basic configuration of the system afterwards, up to the point of being able to use the rest of the manual.
- bootloader: only cover "regular" bootloader in this guide for BIOS/UEFI systems

### Special Installation Guides

These guides will in every instance possible just reference the "main" advanced installation guide, and not duplicate those steps. They will only present *changes* to that procedure, and enough context for the user to be able to follow what is happening. Initial topics could include:

- encrypted disks
- raspberry pi
- other embedded systems
- dual booting
- alternative bootloaders
- etc

### Proposed Process
1. Write a *minimal* installation guide. It should cover anything needed to go from nothing to a working system, but not try to cram everything from the wiki into it. A minimum viable product to manually install a system.
2. Write Alternative Guides for specific use cases and hardware, referencing them as needed from the main guide.
3. ???
4. Profit

bobertlo:
@the-maldridge @nilium any opinions on this? I want to make sure we're on the same page before I start writing this. I'd like to start the main guide and possibly recruit others to write other sections as needed/wanted, because that's a lot of work and I don't want to write about things I'm not familiar with.

Duncaen:
I think the if you can split it into nice logical steps like "Disk setup", with separate pages where you can go in depth into encrypted disks etc you can avoid having "alternative" guides.

The raspberrypi and other embedded systems are too different to setup and basically just hey put this image here and boot, I think they are more like using the installer script and not part of the "Advanced installation" unless they are setup similiar by bootstrapping using xbps-install.

bobertlo:
Thanks @Duncaen those are good points! Put like that I agree. Since the images are not the same procedure, they should be handled separately, and this might remove the need for different sections.

I'll try writing up the advanced guide and see how disk encryption does or does not fit in that guide.

Also, Im not sure what other things there are to cover within the scope of this document.

nilium:
[...]

PR #300, opened 2020-05-22
[WIP] chroot-rootfs: Add section.

flexibeast:
Further to @Duncaen's [comment](https://github.com/void-linux/void-docs/issues/267#issuecomment-626345328) on #267.

### DIFF
--- a/src/config/chroot-rootfs.md
+++ b/src/config/chroot-rootfs.md
@@ -1,0 +1,26 @@
+# Creating a chroot rootfs
+
+As root, create a directory that will contain the chroot, and install a base
+system in it via the `base-voidstrap` package. If network access is required,
+copy `/etc/resolv.conf` into the chroot; `/etc/hosts` may need to be copied as
+well.
+
+Several directories then need to be mounted as follows:
+
+```
+# mount -t proc none <chroot_dir>/proc
+# mount -t sysfs none <chroot_dir>/sys
+# mount --rbind /dev <chroot_dir>/dev
+# mount --rbind /run <chroot_dir>/run
+```
+
+Use [chroot(1)](https://man.voidlinux.org/chroot.1) to change to the new root,
+then run programs as usual. Once you've finished using it, unmount the chroot
+using [umount(8)](https://man.voidlinux.org/umount.8).
+
+## PRoot
+
+An alternative to the above is [proot(1)](https://man.voidlinux.org/proot.1), a
+user-space implementation of chroot, mount --bind, and binfmt_misc. By
+installing the `proot` package, unprivileged users can utilize a chroot
+environment.
--- a/src/installation/musl.md
+++ b/src/installation/musl.md
@@ -25,28 +25,5 @@ which provide their own libc in the image.
 
 ### glibc chroot
 
-Software requiring glibc can be run in a glibc chroot.
-
-Create a directory that will contain the chroot, and install a base system in it
-via the `base-voidstrap` package. If network access is required, copy
-`/etc/resolv.conf` into the chroot; `/etc/hosts` may need to be copied as well.
-
-Several directories then need to be mounted as follows:
-
-```
-# mount -t proc none <chroot_dir>/proc
-# mount -t sysfs none <chroot_dir>/sys
-# mount --rbind /dev <chroot_dir>/dev
-# mount --rbind /run <chroot_dir>/run
-```
-
-Use [chroot(1)](https://man.voidlinux.org/chroot.1) to change to the new root,
-then run glibc programs as usual. Once you've finished using it, unmount the
-chroot using [umount(8)](https://man.voidlinux.org/umount.8).
-
-#### PRoot
-
-An alternative to the above is [proot(1)](https://man.voidlinux.org/proot.1), a
-user-space implementation of chroot, mount --bind, and binfmt_misc. By
-installing the `proot` package, unprivileged users can utilize a chroot
-environment.
+Software requiring glibc can be run in a glibc
+[chroot](../config/chroot-rootfs.md].
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -58,6 +58,7 @@
    - [External Applications](./config/external-applications.md)
    - [Printing](./config/print/index.md)
    - [Manual Pages](./config/man.md)
+   - [chroot rootfs](./config/chroot-rootfs.md)
 - [XBPS Package Manager](./xbps/index.md)
    - [Advanced Usage](./xbps/advanced-usage.md)
    - [Repositories](./xbps/repositories/index.md)


ericonr:
We do have xtool's `xchroot`, which does basically those steps. How do you feel about recommending it?

ericonr:
I'd mention you should create and populate the directory as root, otherwise permissions tend to get weird. And I'm not entirely sure how we could fit `xbps-uunshare` as a non-privileged tool, or even if we need to.

st3r4g:
In order to run graphical applications (imagine a steam chroot on a host musl system) one must also mount `/tmp` (for X11 and Xwayland) and/or `/run` (for Wayland)

Piraty:
how about extending it to be more generic regarding target architecture?
building a bootable image (either same arch or foreign arch, like rpi ) involves some of the same steps but you install `base-system` (or `rpi-base`) instead of `base-voidstrap`

src/installation/musl.md
```diff
@@ -25,28 +25,5 @@ which provide their own libc in the image.
...
-then run glibc programs as usual. Once you've finished using it, unmount the
-chroot using [umount(8)](https://man.voidlinux.org/umount.8).
-
-#### PRoot
-
-An alternative to the above is [proot(1)](https://man.voidlinux.org/proot.1), a
-user-space implementation of chroot, mount --bind, and binfmt_misc. By
-installing the `proot` package, unprivileged users can utilize a chroot
-environment.
+Software requiring glibc can be run in a glibc chroot. For further information,
```
Piraty, line 28: just put the reference under `chroot` keyword

jeffayle:
> but you install base-system (or rpi-base) instead of base-voidstrap

I don't think this is true if you're not building a bootable system. `rpi-base` includes a kernel, which you definitely don't need in this situation. There might be something else in there that you need to get it working though?

flexibeast:
Just to be clear, @Duncaen's comment that led me to create this section was:

> I was thinking more about a "How to setup Void Linux container/chroot rootfs", which I think is useful and quite different from installing void linux on bare metal using a chroot.

There's a good chance i've misinterpreted @Duncaen's comment, but i assumed what was meant was a page about setting up a chroot on an existing Void installation (e.g. a glibc chroot on a musl system, currently in the Handbook [here](https://docs.voidlinux.org/config/musl.html)). Installations via a chroot will be covered by #177 (which is almost complete, and which i feel will be a crucial addition to the Handbook), and then by further work leveraging #177 to address #85.

@ericonr: i wasn't aware of `xchroot` as part of `xtools`, as even though `man xchroot` takes me to the `xtools` man page, there seems to be no mention of it? (And the layout of the `xtools` man page [on man.voidlinux.org](https://man.voidlinux.org/xtools.1) is impressively broken.) Given what i've written above, is it a tool that can be used in this context? If so, the `xtools` man page will need to be updated, and of course, the `xmandump` stuff will need to be resolved so that we can point to the updated man page.

src/installation/musl.md
flexibeast: Sure, will do.

jeffayle:
To what extent should the handbook document *what* to do with a chroot? glibc compatibility on musl definitely makes sense (apart from already being in the handbook) because very few distributions support musl. Other uses might include running x86 binaries on x86_64 distribution (maybe this should go in the multilib page?) and creating jails for non-trusted users (if this is still good advice with docker and lxc around?)

ericonr:
@flexibeast master of xtools has a reference to `xchroot`. It should be usable in that context, but perhaps we can mention it as a "tool that automates this process", and still document the process. I have updated `xmandump`, and maldrige is working on getting it up and running c:

@st3r4g you'd need to bind-mount /tmp for X11 to work, right? In that case, `xchroot` is definitely not the tool.

st3r4g:
the wiki said to install `xhost` and run `xhost +local:`, it worked, but I don't know all the implications of that. Bind-mounting `/tmp` (or maybe just the X socket inside tmp) seems a better suggestion, do you agree?

ericonr:
@st3r4g if `xhost` "just works", it might be a better solution. Did you test how well bind-mounting `/tmp` worked? It *is* a more complicated step and can be easy to get wrong, I think.

st3r4g:
Yes, I'm currently using a script that binds mount `/tmp` too, that way I also get `/tmp` to be a `tmpfs` inside the chroot. I don't think is more complicated really

ericonr:
I'm not sure what I'd prefer to see here, and also don't know what's safest/more correct. Thanks for testing :p

flexibeast:
Referring to `xvoidstrap` and `xchroot` does seem to be the way to go, although there seem to be a few blockers involved:

* `xchroot` does `mount --rbind /proc "$CHROOT/proc"; mount --rbind /sys "$CHROOT/sys"` rather than the `mount` commands described by @q66 in [this comment](https://github.com/void-linux/void-docs/pull/163#discussion_r395958797). i don't know if the source for `xchroot` needs to be changed, or whether instead it's fine for the contexts in which it's intended to be used. (What are those contexts?)

* At any rate, there needs to be a release of `xtools` where the man page references `xchroot` (which it does on `xtools` [master](https://github.com/leahneukirchen/xtools)).

* The layout of [the xtools.1 man page on man.voidlinux.org](https://man.voidlinux.org/xtools.1) needs to be fixed.

ericonr:
- latest version with `xchroot` in the man page was released a few days ago
- I opened an issue about the layout here https://github.com/leahneukirchen/xtools/issues/184, but we don't know what's causing it

flexibeast:
@ericonr: Ah, thanks, i hadn't noticed the `xtools` package update.

flexibeast:
To help keep track of what needs to be done in relation to this section, a list:

[ ] Mention need to mount `/tmp`, unless it should be done via `xchroot` (cf. [this comment](https://github.com/void-linux/void-docs/pull/300#issuecomment-632864663))
[ ] Ensure section is applicable to multiple architectures (cf. [this comment](https://github.com/void-linux/void-docs/pull/300#issuecomment-632887934) and [this comment](https://github.com/void-linux/void-docs/pull/300#issuecomment-632937264))
[ ] Confirm whether the source `xchroot` needs to be modified re. `mount` commands (cf. [this comment](https://github.com/void-linux/void-docs/pull/300#issuecomment-636569842))
[ ] Confirm appropriate use of `xhost` (cf. [this comment](https://github.com/void-linux/void-docs/pull/300#issuecomment-634462437))
[ ] Fix layout of `xtools` man page on man.voidlinux.org (cf. [this issue](https://github.com/leahneukirchen/xtools/issues/184))

ericonr:
I think we should leave `/tmp` as a manual mount. `xchroot` doesn't need to always do it, and a user can choose to either have it as a separate tmpfs or as a bind mount. For this use case, it should be a bind mount.

I believe installing base-system is enough. I have a glibc chroot on my musl RPi. If you need graphics support or something, then you can install `rpi-base` - even then, I don't believe it's needed.

flexibeast:
Blocked on #413.

flexibeast:
@ericonr: Would you be interested in taking over this PR?

ericonr:
I can try!

flexibeast:
Thanks!

src/config/chroot-rootfs.md
```diff
@@ -0,0 +1,26 @@
+# Creating a chroot rootfs
+
+As root, create a directory that will contain the chroot, and install a base
```
Piraty, line 3: it is not necessary that the chroot dir is owned by root.
```
$ XBPS_ARCH=x86_64-musl xbps-install -MS --rootdir=/tmp/rootfs.void.x86_64-musl --repository=https://alpha.de.repo.voidlinux.org/current/musl base-voidstrap
$ proot -R /tmp/rootfs.void.x86_64-musl/
```

PR #302, opened 2020-05-25
src/config/kernel: add reason to remove old kernels

cinerea0:
Added information from https://wiki.voidlinux.org/System_maintenance

### DIFF
--- a/src/config/kernel.md
+++ b/src/config/kernel.md
@@ -18,8 +18,10 @@ modules.
 
 When updating the kernel, old versions are left behind in case it is necessary
 to roll back to an older version. Over time, old kernel versions can accumulate,
-consuming disk space and increasing the time taken by dkms module updates. Thus,
-it may be advisable to clean old kernels from time to time.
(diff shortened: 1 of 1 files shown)


src/config/kernel.md
```diff
@@ -18,7 +18,8 @@ modules.
 
 When updating the kernel, old versions are left behind in case it is necessary
 to roll back to an older version. Over time, old kernel versions can accumulate,
-consuming disk space and increasing the time taken by dkms module updates. Thus,
+consuming disk space and increasing the time taken by dkms module updates. If
+the boot partition fills up with old kernels, it can cause a kernel panic. Thus,
```
ericonr, line 22: Suggest: "Furthermore, if `/boot` is a separate partition and fills up with old kernels, it can cause a kernel panic."

cinerea0: Amended!

ericonr:
Awesome, thanks!

PR #304, opened 2020-05-26
xorg: explain DDX vs modesetting

### LINKED ISSUE #303: Xorg section should describe modesetting vs DDX
As suggested by @st3r4g in [this comment](https://github.com/void-linux/void-docs/pull/297#discussion_r430277148):

> I think we need a paragraph in the xorg section explaining that Xorg can run with two different kinds of graphics drivers: 'modesetting' vs 'DDX', and how to check what you're running.

st3r4g:
Closes #303
'Forcing the modesetting driver' was tested under a minimal `twm` setup with `intel` drivers, testing on different setups is welcome

### DIFF
--- a/src/config/graphical-session/xorg.md
+++ b/src/config/graphical-session/xorg.md
@@ -24,21 +24,28 @@ Void provides both open-source and proprietary (non-free) video drivers.
 
 ### Open Source Drivers
 
-The open source drivers are installed with the `xorg` package by default, or may
-be installed individually if the `xorg-minimal` was installed. Below is a table
-of device brands and their driver packages.
+Xorg can use two categories of open source drivers: DDX or modesetting.
 
-| Brand  | Type        | Driver Package       |
-|--------|-------------|----------------------|
-| AMD    | Open Source | `xf86-video-amdgpu`  |
-| ATI    | Open Source | `xf86-video-ati`     |
-| Intel  | Open Source | `xf86-video-intel`   |
-| NVIDIA | Open Source | `xf86-video-nouveau` |
+#### DDX
 
-> Note: Fourth generation Intel users may want to use the default xorg driver,
-> rather than installing `xf86-video-intel` driver package. For more
-> information, see the [Arch wiki
-> page](https://wiki.archlinux.org/index.php/Intel_graphics#Installation).
+The DDX drivers are installed with the `xorg` package by default, or may be
+installed individually if the `xorg-minimal` package was installed. They are
+provided by the `xf86-video-*` packages.
+
+For advanced configuration, see the man page corresponding to the vendor name,
+like [intel(4)](https://man.voidlinux.org/intel.4).
+
+#### Modesetting
+
+Modesetting requires the `mesa-dri` package, and no additional vendor-specific
+driver package.
+
+Xorg defaults to DDX drivers if they are present, so in this case modesetting
+must be explicitly selected: see [Forcing the modesetting
+driver](#forcing-the-modesetting-driver).
+
+For advanced configuration, see
+[modesetting(4)](https://man.voidlinux.org/modesetting.4).
 
 ### Proprietary Drivers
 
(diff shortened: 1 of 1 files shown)


src/config/graphical-session/xorg.md
```diff
@@ -24,9 +24,13 @@ Void provides both open-source and proprietary (non-free) video drivers.
...
-of device brands and their driver packages.
+Xorg can use two categories of open source drivers: DDX or modesetting.
+
+#### DDX
+
+The DDX drivers are installed with the `xorg` package by default, or may be
+installed individually if the `xorg-minimal` was installed. Below is a table of
+device brands and their driver packages.
 
 | Brand  | Type        | Driver Package       |
```
q66, line 35: I don't like this table, it's inaccurate (`xf86-video-ati` is not only for old ATi GPUs) and incomplete (we have a lot more drivers than that)

ericonr: Do you think we should try to document all of them instead? Or perhaps find some upstream docs with all of those drivers which we can link to.

st3r4g: Indeed, the table is not very useful. I wouldn't list everything, I'd just say that the DDX are provided by the `xf86-video-*` packages. The only non-obvious choice is `xf86-video-ati` vs `xf86-video-amdgpu`, which would be better documented on a future AMD subsection of Graphics Drivers.

src/config/graphical-session/xorg.md
```diff
@@ -82,6 +96,25 @@ Section "InputClass"
 EndSection
 ```
 
+### Forcing the modesetting driver
+
+Create the file `/etc/X11/xorg.conf.d/10-modesetting.conf`:
+
+```
+Section "OutputClass"
```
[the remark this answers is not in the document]
st3r4g: What do you mean with 'portable to everything'? I've chosen `OutputClass` instead of `Device` because it supports all options you can pass to `Device`, plus you can use `MatchDriver` to discriminate between GPUs based on the kernel driver, which is useful on multi-gpu setups.
But either is fine for me in this case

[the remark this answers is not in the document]
q66: On multi-GPU setups you will want an explicit set of Devices plus a Screen attached to the default, as Xorg does not deal reliably with them, often it will fail to start when you have multiple GPUs and leave it for autoconfig.

src/config/graphical-session/xorg.md
```diff
@@ -24,9 +24,13 @@ Void provides both open-source and proprietary (non-free) video drivers.
...
 
-The open source drivers are installed with the `xorg` package by default, or may
-be installed individually if the `xorg-minimal` was installed. Below is a table
-of device brands and their driver packages.
+Xorg can use two categories of open source drivers: DDX or modesetting.
+
+#### DDX
+
+The DDX drivers are installed with the `xorg` package by default, or may be
+installed individually if the `xorg-minimal` was installed. Below is a table of
```
ericonr, line 32: "if the `xorg-minimal` package"

src/config/graphical-session/xorg.md
```diff
@@ -35,10 +39,20 @@ of device brands and their driver packages.
...
+
+Modesetting requires the `mesa-dri` package, and no additional vendor-specific
+driver package.
+
+Xorg defaults to DDX drivers if they are present, so in this case modesetting
+must be explicitly selected: see [Forcing the modesetting
+driver](#forcing-the-modesetting-driver).
+
+For advanced configuration, see
+[modesetting(4)](https://man.voidlinux.org/modesetting).
```
ericonr, line 55: For future proofing, use the modesetting.4 link instead.

src/config/graphical-session/xorg.md
st3r4g: Table removed

ericonr:
Looks good, thanks!

PR #308, opened 2020-05-29
about-handbook/index, style-guide: Add text about placeholders.

flexibeast:
This PR is based on extant usage patterns in the Handbook.

### DIFF
--- a/src/about-handbook/index.md
+++ b/src/about-handbook/index.md
@@ -24,3 +24,15 @@ Examples in this guide may have snippets of commands to be run in your shell.
 When you see these, any line beginning with `$` is run as your normal user.
 Lines beginning with `#` are run as `root`. After either of these lines, there
 may be example output from the command.
+
+### Placeholders
+
+Some examples include text with placeholders. Placeholders indicate where you
+should substitute the appropriate information. For example:
+
+> ```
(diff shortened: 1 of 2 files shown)


hippi777:
hi there! :)

what about "$whatever" for placeholders instead of "<whatever>"? its better for copypaste codes and its more natural...

abenson:
Using `<placeholder>` is fairly common in industry tech-writing.

flexibeast:
@hippi777: Thank you for the suggestion, but no, for a few reasons:

* Given that many commands are intended to be run from a shell, `$whatever` looks much more like something that should be typed literally, compared to `<whatever>`.

* As [the style guide](https://docs.voidlinux.org/contributing/void-docs/style-guide.html) notes:

> Outside of the 'installation' sections, step-by-step instructions containing 'magic' commands for copying-and-pasting are strongly discouraged. Users are expected to read the canonical documentation (e.g. man pages) for individual programs to understand how to use them, rather than relying on copying-and-pasting.

* `<whatever>` is the most commonly-used format currently used in the Handbook, and i feel there's no sufficiently strong reason to get into bikeshedding an alternative, so it's enough to codify existing practice to ensure consistency of style.

hippi777:
fine, it was only an idea :D

Issue #245, opened 2020-05-06
Services documentation


ericonr:
runlevels -> default, current, single

`/etc/runit/runsvdir/current` -> link services when the system isn't running

`runsv` -> testing services

other things as well

http://smarden.org/runit/index.html is a good resource, of course

PR #314, opened 2020-05-31
[nomerge] Slim down services and update some info

ericonr:


ericonr:
Should close #245 

@Duncaen

src/config/services/index.md
```diff
@@ -32,3 +37,63 @@ Optionally, a service directory may contain:
...
 created on the first run.
+
+### Configuring services
+
+Most services can take configuration options set by a `conf` file in the service
+directory. This allows service customization without modifying the service
+directory provided by the relevant package.
+
+Check the service file for how to pass configuration parameters. Services
+usually expect `OPTS="--value ..."` in the `conf` file.
```
flexibeast, line 48: "expect something like"? Are longopts usually used?

src/config/services/index.md
```diff
@@ -32,3 +37,63 @@ Optionally, a service directory may contain:
...
+
+To make more complex customizations than provided by default, [edit the
```
flexibeast, line 50: i think we can remove "than provided by default".

src/config/services/index.md
```diff
@@ -32,3 +37,63 @@ Optionally, a service directory may contain:
...
+service](#editing-services).
+
+### Editing services
+
+To edit a service, first copy its service directory to a different directory
+name, otherwise [xbps-install(1)](https://man.voidlinux.org/xbps-install.1) can
```
flexibeast, line 56: "name. Otherwise,"

src/config/services/index.md
```diff
@@ -32,3 +37,63 @@ Optionally, a service directory may contain:
...
+
+[socklog(8)](https://man.voidlinux.org/socklog.8) is a syslog implementation
+from the author of [runit(8)](https://man.voidlinux.org/runit.8). Use socklog if
+you're not sure which syslog implementation to use. To use it, install the
+`socklog-void` package, and enable the `socklog-unix` and `nanoklogd` services.
+
+The logs are saved in sub-directories of `/var/log/socklog/`, and `svlogtail`
+can be used to help access them conveniently.
+
+The ability to reading logs is limited to `root` and users who are part of the
```
flexibeast, line 78: "read"

src/config/services/index.md
```diff
@@ -32,3 +37,63 @@ Optionally, a service directory may contain:
...
+
+## Console
+
+### Disabling default ttys
+
+Void enables [agetty(8)](https://man.voidlinux.org/agetty.8) services for `ttys`
+1 to 6 by default.
+
+To disable agetty services, remove the service symlink and create a `down` file
+in the agetty service directory to avoid that updates of the `runit-void`
```
flexibeast, line 93: Assuming i'm interpreting this correctly, "to avoid the service being re-enabled by updates of the `runit-void` package".

src/config/services/index.md
```diff
@@ -32,3 +37,63 @@ Optionally, a service directory may contain:
...
+package re-enable the service.
+
+```
+# unlink /var/service/agetty-tty6
```
flexibeast, line 97: For my own edification, why is it better to use `unlink` rather than `rm` to disable services?

src/config/services/managing.md
```diff
@@ -60,14 +60,14 @@ create a file named `down` in its service directory:
 To disable a service, remove the symlink from the running runsvdir:
 
 ```
-# rm /var/service/<service>
+# unlink /var/service/<service>
 ```
 
 Or from the `default` runsvdir if the system or the specific runsvdir is not
```
flexibeast, line 66: Hmm, maybe "if either the specific runsvdir, or the system, is not"?

src/config/services/managing.md
```diff
@@ -102,6 +102,7 @@ See [runsvdir(8)](https://man.voidlinux.org/runsvdir.8) and
 
 ### Booting a different runsvdir
 
-To boot a different runsvdir, the name of the runsvdir can be added to the
-[kernel command-line](../kernel.html#cmdline). As example adding `single` to the
-[kernel command line](../kernel.html#cmdline) will boot the `single` runsvdir.
+To boot a `runsvdir` other than `default`, the name of the desired `runsvdir`
```
flexibeast, line 105: "runsvdir" is not in backticks earlier in the section, which i think makes sense, so maybe remove the backticks here, and in the next instance?

src/config/services/user-services.md
```diff
@@ -1,11 +1,11 @@
 # Per-User Services
 
-Sometimes it would be nice to have user-specific runit services. Services that,
+Sometimes it can be nice to have user-specific runit services. Services that,
```
flexibeast, line 3: For example, you might like to open an ssh tunnel as the current user,"

src/config/services/index.md
ericonr: I count seven services locally (out of a lot). Since it's the only "standard" mechanism, I'd like to have it. I think I can change it to something like "may use something like..." instead.

src/config/services/user-services.md
```diff
@@ -1,11 +1,11 @@
...
 for example, open an ssh tunnel as your current user, run a virtual machine, or
 regularly run daemons on your behalf. The most common way to do this is to ask a
 system-level service to start a
 [runsvdir(8)](https://man.voidlinux.org/runsvdir.8) service as your user for
-your personal service directory.
+your personal [runsvdir](./index.md#runsvdirs).
 
 As an example, create a service `/etc/sv/runsvdir-<username>` with the following
```
flexibeast, line 10: "For example, you could create a service"

src/config/services/index.md
ericonr: Indeed.

src/config/services/index.md
ericonr: yup!

src/config/services/index.md
ericonr: This is actually something I wanted other's opinions on. I prefer using unlink because the action is more obvious and I think for documentation purposes it's more explicit.

src/config/services/managing.md
ericonr: This was mostly because the other page constantly used backticks for runsvdir. It probably makes more sense to not use backticks unless we are talking about the command, though. Will change!

src/config/services/index.md
flexibeast: Oh, sorry, i might have been unclear: OPTS should definitely be mentioned! i was just thinking that people may think longopts are always used for services; your suggested phrasing of "may use something like" is fine to deal with that. :-)

src/config/services/index.md
flexibeast: That's a good point. i, too, would be interested if there's any reason(s) to *not* use `unlink`.

src/config/services/managing.md
```diff
@@ -60,14 +60,14 @@ create a file named `down` in its service directory:
 To disable a service, remove the symlink from the running runsvdir:
 
 ```
-# rm /var/service/<service>
+# unlink /var/service/<service>
```
Duncaen, line 63: Why?

ericonr: > This is actually something I wanted other's opinions on. I prefer using unlink because the action is more obvious and I think for documentation purposes it's more explicit.

I can revert to `rm`, though.

Duncaen: `rm` is more commonly used and well known, "unlink" doesn't mean remove a link, both tools use the unlink syscall.

Duncaen: `unlink` is the same as `rm`, its not exclusively for removing {sym,}links and there is no reason to really prefer unlink here over rm, most users know rm and its shorter to write.

ericonr:
@Duncaen mostly because adding "Managing Services" in there would make the whole thing too big. Do you think "Logging" and "Console" should become separate pages or that everything should become a single one, with possibly only "Per-user Services" by itself?

Duncaen:
I think that configuring and managing services are the main point of the whole section, I don't think splitting them makes a lot of sense because we gain nothing from it other than slightly shorter pages.

Logging, Console and Per User services are all small edge cases I don't really see fit into a general page.

src/config/services/index.md
ericonr: Duncaen recommended leaving it as `rm` :p

ericonr:
Closed in #319

PR #318, opened 2020-06-02
xorg.md: change table to list

jeffayle:
This section previously generated an html table but it does not contain tabular
data. It has been changed to an unordered list.


ericonr:
I'd actually prefer if we just said that "these drivers are `xf86-input-*`, like we started doing for the `xf86-video-*` ones.

jeffayle:
Good point, you can get that list from xbps

ericonr:
Thanks!

PR #319, opened 2020-06-04
Slim services 2

ericonr:
Slim services 2: electric boogaloo, supersedes #314 

This one moves things around a lot, because the changes @Duncaen suggested required restructuring the whole thing. I have tried to implement all of @flexibeast 's suggestions from #314 .


src/config/services/index.md
```diff
@@ -32,3 +37,136 @@ Optionally, a service directory may contain:
...
+name. Otherwise, [xbps-install(1)](https://man.voidlinux.org/xbps-install.1) can
+overwrite the service directory. Then, edit the new service file as needed.
+Finally, the old service should be stopped and disabled, and the new one should
+be started.
+
+## Managing Services
+
+### Runsvdirs
+
+A `runsvdir` is a directory in `/etc/runit/runsvdir` containing enabled services
```
flexibeast, line 65: Upon further thought, i think we shouldn't use backticks unless we're referring to a command. In this instance, we're referring to a directory, not a command, so perhaps we should use "\*runsvdir\* instead, using emphasis because we're defining a term.

src/config/services/index.md
```diff
@@ -32,3 +37,136 @@ Optionally, a service directory may contain:
...
+in the form of symlinks to service directories. The currently running `runsvdir`
```
flexibeast, line 66: Since we're neither referring to a command or defining a term, i think we should remove the backticks here.

src/config/services/index.md
```diff
@@ -32,3 +37,136 @@ Optionally, a service directory may contain:
...
+will be linked to `/var/service` when the system is booted.
+
+The `runit-void` package comes with two runsvdir directories; `single` and
+`default`.
+
+- `single` just runs sulogin(8) and the necessary steps to rescue your system.
+- `default` the default runsvdir.
```
flexibeast, line 73: Perhaps "is the default runsvdir once the system is running"?

src/config/services/index.md
```diff
@@ -32,3 +37,136 @@ Optionally, a service directory may contain:
...
+Additional runsvdirs can be created in `/etc/runit/runsvdir/`.
+
+See [runsvdir(8)](https://man.voidlinux.org/runsvdir.8) and
+[runsvchdir(8)](https://man.voidlinux.org/runsvchdir.8) for further information.
+
+#### Booting a different runsvdir
+
++To boot a runsvdir other than `default`, the name of the desired runsvdir +can
+be added to the [kernel command-line](../kernel.html#cmdline). As an +example,
+adding `single` to the kernel command line will boot the `single` +runsvdir.
```
flexibeast, line 84: Extraneous "+"s in this para? Also, change the link to use "kernel.md".

src/config/services/index.md
```diff
@@ -32,3 +37,136 @@ Optionally, a service directory may contain:
...
+# sv down <services>
+# sv restart <services>
+# sv status <services>
+```
+
+The `<services>` placeholder can be:
+
+- Service names (service directory names) inside of the `/var/service`
+   directory.
+- The full path to the services.
```
flexibeast, line 101: "paths"

src/config/services/index.md
```diff
@@ -32,3 +37,136 @@ Optionally, a service directory may contain:
...
+
+As example the following commands show the status of a specific service and of
```
flexibeast, line 103: "For example, the"

src/config/services/index.md
```diff
@@ -32,3 +37,136 @@ Optionally, a service directory may contain:
...
+
+- `single` just runs sulogin(8) and the necessary steps to rescue your system.
+- `default` the default runsvdir.
+
+Additional runsvdirs can be created in `/etc/runit/runsvdir/`.
+
+See [runsvdir(8)](https://man.voidlinux.org/runsvdir.8) and
+[runsvchdir(8)](https://man.voidlinux.org/runsvchdir.8) for further information.
+
+#### Booting a different runsvdir
```
flexibeast, line 80: "Booting A Different runsvdir"

src/config/services/index.md
```diff
@@ -32,3 +37,136 @@ Optionally, a service directory may contain:
...
+create a file named `down` in its service directory:
+
+```
+# touch /etc/sv/<service>/down
+```
+
+The `down` file mechanism also makes it possible to disable services that are
+enabled by default, such as the [agetty(8)](https://man.voidlinux.org/agetty.8)
+services for ttys 1 to 6. This way, package updates which affect these services
+- in this case the `runit-void` package - won't re-enable them.
```
flexibeast, line 144: "case, the"

src/config/services/index.md
ericonr: I think I will only leave the first one bolded. Sounds good?

src/config/services/index.md
ericonr: I think I see what you want to make clear, but I don't know if that's the best way to phrase it.

src/config/services/index.md
ericonr: Yeah, copying from a commit diff has its disadvantages. Thanks!

ericonr:
Updated! :D

src/config/services/index.md
ericonr: "is the default runsvdir  if no runsvidr is specified in the command line" ?

src/config/services/index.md
flexibeast: Sure. :-)

src/config/services/index.md
flexibeast: Hmm. "is the default runsvdir on a running system, unless specified otherwise by the kernel command line"? Basically i'd like to make it clear that `default` is not the default runsvdir *at* startup, but *after* startup.

ericonr: Used that!

ericonr:
Updated again.

ericonr:
CI issues due to mirror being down :c

src/config/services/index.md
```diff
@@ -32,3 +37,137 @@ Optionally, a service directory may contain:
...
+Finally, the old service should be stopped and disabled, and the new one should
+be started.
+
+## Managing Services
+
+### Runsvdirs
+
+A **runsvdir** is a directory in `/etc/runit/runsvdir` containing enabled
+services in the form of symlinks to service directories. The currently running
+runsvdir will be linked to `/var/service` when the system is booted.
```
flexibeast, line 67: Suggest just: "On a running system, the current runsvdir is accessible via the \`/var/service\` symlink."

src/config/services/index.md
```diff
@@ -32,3 +37,137 @@ Optionally, a service directory may contain:
...
+
+The `runit-void` package comes with two runsvdir directories; `single` and
+`default`.
```
flexibeast, line 70: s/./:/

src/config/services/index.md
```diff
@@ -32,3 +37,137 @@ Optionally, a service directory may contain:
...
+
+## Managing Services
+
+### Runsvdirs
+
+A **runsvdir** is a directory in `/etc/runit/runsvdir` containing enabled
+services in the form of symlinks to service directories. The currently running
+runsvdir will be linked to `/var/service` when the system is booted.
+
+The `runit-void` package comes with two runsvdir directories; `single` and
```
flexibeast, line 69: s/;/,/

src/config/services/index.md
```diff
@@ -32,3 +37,137 @@ Optionally, a service directory may contain:
...
+```
+# sv up <services>
+# sv down <services>
+# sv restart <services>
+# sv status <services>
+```
+
+The `<services>` placeholder can be:
+
+- Service names (service directory names) inside of the `/var/service`
```
flexibeast, line 100: Remove "of".

src/config/services/user-services.md
```diff
@@ -1,14 +1,14 @@
 # Per-User Services
 
-Sometimes it would be nice to have user-specific runit services. Services that,
-for example, open an ssh tunnel as your current user, run a virtual machine, or
+Sometimes it can be nice to have user-specific runit services. For example, you
+might want to open an ssh tunnel as the current user, run a virtual machine, or
 regularly run daemons on your behalf. The most common way to do this is to ask a
 system-level service to start a
 [runsvdir(8)](https://man.voidlinux.org/runsvdir.8) service as your user for
```
flexibeast, line 7: s/for/via/

src/config/services/user-services.md
```diff
@@ -1,14 +1,14 @@
...
-your personal service directory.
+your personal services directory.
```
flexibeast, line 8: s/your/a/

src/config/services/index.md
ericonr: Much better!

src/config/services/index.md
ericonr: I used a ":" instead.

src/config/services/user-services.md
ericonr: The user services themselves will run via a personal services directory, but we need the system-level service **for** the user's services directory.

flexibeast: Fair enough, but still, i find the structure of this sentence quite awkward. Not enough to make it a blocker, though, so let's leave it as-is for now.

ericonr: @flexibeast I tried to make a few changes, what do you think?

flexibeast: @ericonr: Oh that's much better, thanks!

src/config/services/index.md
```diff
@@ -32,3 +37,135 @@ Optionally, a service directory may contain:
...
+
+### Runsvdirs
+
+A **runsvdir** is a directory in `/etc/runit/runsvdir` containing enabled
+services in the form of symlinks to service directories. On a running system,
+the current runsvdir is accessible via the `/var/service` symlink.
+
+The `runit-void` package comes with two runsvdirs, `single` and `default`:
+
+- `single` just runs sulogin(8) and the necessary steps to rescue your system.
```
flexibeast, line 71: Link to man.voidlinux.org here?

flexibeast:
The reference to `sulogin(8)` should probably be a link, but other than that, i think this is ready to merge. Thanks!

ericonr:
Awesome! Thanks for the review c:

PR #322, opened 2020-06-08
fde: make UUID selection clearer

ericonr:
@abenson the information checks out, right?


abenson:
I think that's clear enough, yes.

PR #330, opened 2020-06-13
wayland: update and trim.

ericonr:
- Add Enlightenment
- Remove terminal emulators section (unnecessary)
- Move media applications content to the general applications header
- Some clean up

@q66 let me know if you think I can trim some more.

Some of the information in "general desktop utilities" feels like "not our responsibility", but at the same time I think it's generally valuable. @flexibeast what do you think?

### DIFF
--- a/src/config/graphical-session/wayland.md
+++ b/src/config/graphical-session/wayland.md
@@ -10,12 +10,12 @@ the window manager and the compositor in a single application.
 
 ### Desktop Environments
 
-Both GNOME and KDE Plasma have Wayland sessions. For GNOME, Wayland is already
-the default session, while Plasma still has a few issues to iron out, so their
-default remains a Xorg session. While running these desktop environments,
-applications built with GTK+ will automatically choose the Wayland backend, but
-Qt applications might require the proper environment variable in GNOME. This is
-explained [below](#native-applications).
+GNOME, KDE Plasma and Enlightenment have Wayland sessions. GNOME uses its
+Wayland session by default. When using these desktop environments, applications
+built with GTK+ will automatically choose the Wayland backend, while Qt5 and EFL
+applications might require [setting some environment
+variables](#native-applications) if used outside KDE or Enlightenment,
+respectively.
 
 ### Standalone compositors
 
@@ -32,18 +32,26 @@ Void Linux currently packages the following Wayland compositors:
 Both GNOME and KDE Plasma have EGLStreams backends for Wayland, which means they
 can use the proprietary NVIDIA drivers. Most other Wayland compositors require
 drivers that implement the GBM interface. The main driver for this purpose is
-provided by the `mesa-dri` package. The [graphics
-section](./graphics-drivers/index.md) has more details regarding setting up
-graphics in different systems.
+provided by the `mesa-dri` package. The "[Graphics
+Drivers](./graphics-drivers/index.md)" section has more details regarding
+setting up graphics in different systems.
 
 ### Native applications
 
-Qt5 based applications require the installation of the `qt5-wayland` package and
-setting the environment variable `QT_QPA_PLATFORM=wayland-egl` to enable their
-Wayland backend. Some KDE specific applications might also require the
-installation of the `kwayland` package. GTK+ applications should do it
-automatically, but setting `GDK_BACKEND=wayland` can force the Wayland backend
-(however, this can break applications such as Chromium).
+[Qt5](https://wayland.freedesktop.org/qt5.html)-based applications require
+installing the `qt5-wayland` package and setting the environment variable
+`QT_QPA_PLATFORM=wayland-egl` to enable their Wayland backend. Some KDE specific
+applications also require installing the `kwayland` package.
+[EFL](https://wayland.freedesktop.org/efl.html)-based applications require
+setting the environment variable `ELM_DISPLAY=wl`, and can have issues without
+it, due to not supporting XWayland properly.
+[GTK+](https://wiki.gnome.org/Initiatives/Wayland/GTK%2B)-based applications
+should use the Wayland backend automatically. Information about other toolkits
+can be found in the [Wayland
+documentation](https://wayland.freedesktop.org/toolkits.html).
+
+Media applications, such as [mpv(1)](https://man.voidlinux.org/mpv.1),
+[vlc(1)](https://man.voidlinux.org/vlc.1) and `imv` work natively on Wayland.
 
 #### Web browsers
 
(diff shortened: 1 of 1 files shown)


flexibeast:
> Some of the information in "general desktop utilities" feels like "not our responsibility", but at the same time I think it's generally valuable. @flexibeast what do you think?

Definitely not what the Handbook is for; the Wayland section doesn't need to mention Wayland-specific software, any more than the Xorg section needs to mention the various stock X apps. And of course, more generally, we don't want to be providing lists of available applications (not least because these sort of lists can lead to bikeshedding over what should and shouldn't be included). So, yeah, that part should be removed entirely.

ericonr:
This is soft blocked on https://github.com/void-linux/void-packages/pull/22872

ericonr:
Should be ready for review, now.

src/config/graphical-session/wayland.md
```diff
@@ -32,18 +32,22 @@ Void Linux currently packages the following Wayland compositors:
 Both GNOME and KDE Plasma have EGLStreams backends for Wayland, which means they
 can use the proprietary NVIDIA drivers. Most other Wayland compositors require
 drivers that implement the GBM interface. The main driver for this purpose is
-provided by the `mesa-dri` package. The [graphics
+provided by the `mesa-dri` package. The [graphics drivers
```
flexibeast, line 35: This should use [the double-quoting style for referring to section titles](https://docs.voidlinux.org/contributing/void-docs/style-guide.html#internal-links).

src/config/graphical-session/wayland.md
```diff
@@ -32,18 +32,22 @@ Void Linux currently packages the following Wayland compositors:
...
 
 ### Native applications
 
-Qt5 based applications require the installation of the `qt5-wayland` package and
-setting the environment variable `QT_QPA_PLATFORM=wayland-egl` to enable their
-Wayland backend. Some KDE specific applications might also require the
-installation of the `kwayland` package. GTK+ applications should do it
-automatically, but setting `GDK_BACKEND=wayland` can force the Wayland backend
-(however, this can break applications such as Chromium).
+[Qt5](https://wayland.freedesktop.org/qt5.html) based applications require the
```
flexibeast, line 41: "\[Qt5\]\(https://wayland.freedesktop.org/qt5.html)-based"

src/config/graphical-session/wayland.md
```diff
@@ -32,18 +32,22 @@ Void Linux currently packages the following Wayland compositors:
...
+installation of the `qt5-wayland` package and setting the environment variable
+`QT_QPA_PLATFORM=wayland-egl` to enable their Wayland backend. Some KDE specific
+applications might also require the installation of the `kwayland` package.
+[EFL](https://wayland.freedesktop.org/efl.html) based applications require
```
flexibeast, line 45: "\[EFL\]\(https://wayland.freedesktop.org/efl.html\)-based"

src/config/graphical-session/wayland.md
```diff
@@ -32,18 +32,22 @@ Void Linux currently packages the following Wayland compositors:
...
+setting the environment variable `ELM_DISPLAY=wl`, and can have issues without
```
flexibeast, line 46: Is ", and can have issues ..." required?

src/config/graphical-session/wayland.md
flexibeast: "require installing the", to match "setting the" later in the sentence.

src/config/graphical-session/wayland.md
```diff
@@ -32,18 +32,22 @@ Void Linux currently packages the following Wayland compositors:
...
+it. GTK+ applications should use the Wayland backend automatically.
```
flexibeast, line 47: "\[GTK+\](https://gtk.org/)-based", to match style used above when referencing other tooklkits.

src/config/graphical-session/wayland.md
ericonr: I can be more specific about the issue, but I think it's necessary to mention it. Basically, EFL has some weird issue with XWayland, which it tries to use by default, and simply doesn't launch the application. It is necessary to set that environment variable for it to work at all (or unset `DISPLAY`, but that's a worse solution).

src/config/graphical-session/wayland.md
ericonr: I'm using links from the Wayland page about toolkits, so it might be good to link said page. Let me know what you think.

src/config/graphical-session/wayland.md
flexibeast: No, my concern is not about going into the details of the specific issues, but that you've already said setting the env var is required, so is there any need to mention that it can have issues without it? Obviously we wouldn't be saying that it's required otherwise. :-)

src/config/graphical-session/wayland.md
flexibeast: i.e. https://wiki.gnome.org/Initiatives/Wayland/GTK%2B ? Sure; my concern here is that "GTK+" be a link, like the other toolkit mentions are. (And that there be a hyphen between \<toolkit name\> and "based".)

src/config/graphical-session/wayland.md
ericonr: I wanted to avoid confusion regarding the word required: "required" to force the application to use Wayland, or "required" to make the application work at all. If you think it was clear already without all the explanation, I can certainly remove it.

src/config/graphical-session/wayland.md
ericonr: I believe that's solved, then. Thanks!

src/config/graphical-session/wayland.md
flexibeast: Oh, so like "applications *additionally* require"? If so, then i'd prefer that rather than the final phrase; but otherwise, i'm okay with leaving the latter in.

ericonr: I'm not sure I get you... Qt5 requires that stuff to use Wayland, but you can skip it and it just won't use Wayland. EFL requires that variable to work at all (or at least for cases when Xwayland is enabled, which will be most of them). I'm not sure how else I can phrase it.

flexibeast: i'm sorry i'm not being clear. i'm just saying that if we say that setting an env var is *required*, then it's already implied that *not* setting it will cause issues, so the ", and can have issues" part of the sentence is not needed. But although i feel it's unnecessary, it doesn't hurt that much to leave it in, so i'm happy to leave it as-is.

src/config/graphical-session/wayland.md
q66: The `ELM_DISPLAY` var didn't seem to be needed at least when actually running Enlightenment but shrug

ericonr: I tried to make that clear in the `desktop environment` block. If you think it could be rephrased, I'd be glad to do so. The issue presents itself with Sway when XWayland is enabled. It's this bug: https://phab.enlightenment.org/T7439

q66: I guess it may still be problematic even in E when xwayland is running, but not sure... it shouldn't hurt to set it anyway, i guess

PR #339, opened 2020-06-18
mirrors/index: add maakpain mirror, organize alphabetically.

ericonr:
Need opinions on the alphabetical organization.


flexibeast:
Organising alphabetically seems like a good idea to me. But i wonder if the contents of the "Location" field shouldn't have a more consistent format? E.g. "Asia" instead of "APAN"; and i presume we don't specify cities when unknown?

flexibeast:
(In [this comment](https://www.reddit.com/r/voidlinux/comments/hbhbsu/how_can_i_get_listed_into_tier_2_mirror/fv8whbs/), /u/Maakpain notes their mirror is in Seoul.)

Vaelatern:
I prefer the primary sort field to be Location, and only then sub-sort by URL if necessary.

Issue #337, opened 2020-06-17
Ensure relevant content from wiki FAQ page is covered by Handbook


flexibeast:
[Wiki "Frequently Asked Questions" page](https://wiki.voidlinux.org/Frequently_Asked_Questions):

- [x] 1.1 Is Void an Arch Linux fork?
   Void [landing page](https://voidlinux.org) has a "Not a fork" section.
- [x] 1.2 Is there a 'Stable' version of Void?
   #349 (merged)
- [x] 1.3 Void Linux name origin
   #349 (merged)
- [x] 2.1 How should I partition my disk?
   Covered by "Partitioning Notes" section in Handbook.
- [x] 2.2 The installer sucks, how can I do $SPECIALTHING on installation?
   Covered, and being covered, by advanced guides.
- [x] 2.3 How can I change my users default shell?
   Out-of-scope, unneeded.
- [x] 2.4 How do I remove old kernels?
   Covered by "Kernel" section in Handbook.
- [x] 2.5 How can I load kernel modules at boot?
   Covered by "Kernel" section in Handbook.
- [x] 2.6 How do I blacklist kernel modules?
   Covered by "Kernel" section in Handbook.
- [x] 2.7 How to get rid of "ataN: softreset failed (device not ready)" ?
   Out-of-scope, unneeded.
- [x] 2.8 How do I set up/start/stop services?
   Covered by "Services and Daemons - runit" section in Handbook.
- [x] 3.1 Which package contains XYZ?
   Covered by "Finding Files" section in Handbook.
- [x] 3.2 How do I get a list of installed packages?
   #343 (merged)
- [x] 3.3 I want to downgrade a package!
   Covered by "Downgrading to a specific package version" section in Handbook.
- [x] 3.4 I installed $PKGNAME and it's not in my $PATH!
   #404 (merged)
- [x] 3.5 A 'Transaction aborted due to unresolved shlibs' error is preventing me from updating!
   Covered by "Common Errors" section in Handbook.
- [x] 4.1 Valgrind needs glibc-dbg, which cannot be found. Help?
   Covered by "debug" section of Handbook.
- [x] 4.2 How do I check template files for common errors?
   Covered by void-packages/CONTRIBUTING.md.
- [x] 5.1 [Why] doesn't $MANWIDTH change the man pages' width?
   Covered by `man man` and `man man.conf`.
- [x] 5.2 [Why] is tar/Engrampa/XYZ unable to unpack *tar.xz archives?
[...]

Issue #327, opened 2020-06-12
Tracking: wiki content still to be added to the Handbook


flexibeast:
i've now completed a low-hanging-fruit pass of wiki articles to mark articles that i'm confident are either:

* satisfactorily covered by the Handbook; and/or
* clearly out-of-scope for the Handbook (e.g. don't contain any Void-specific information beyond something like `xbps-install <package>`).

i've marked such articles by renaming them to include a trailing dash, to make them visible on the [AllPages](https://wiki.voidlinux.org/Special:AllPages) wiki page. In the cases where there's a Handbook section directly corresponding to a wiki article - which is not always the case - i've removed the article contents and replaced it with the text "As this wiki is deprecated, and will soon disappear, please refer to \<link to Handbook section\>."

i have intentionally not provided redirects from the old page name to the new page name, so that search engines etc. get 404s and reduce the ranking of the wiki in search results. (A [Google search for 'void pulseaudio'](https://www.google.com/search?q=void+pulseaudio) currently ranks the wiki above the Handbook.)

The following is a list of all articles not so marked by this process:

* Acpi
* Adb
* Android
* AppArmor
* AppImages
* Arduino
* Automatic Login to Graphical Environment
* Automount
* Autox
* Beaglebone
* Bspwm
* Cinnamon
* Compiz
* Console mouse support
* Create Git Versions of Packages
* Cross Compiler
* Cubieboard2 NAND Install Hack
* Cubieboard2 SD-Card
* Cubox
* Cuda
* Disks
* Distcc
* Dmix
* Docker
* Dolphin
* Dynamic Kernel Module Support
* Embedded Images
* FAQ
* Firewall Configuration
* Firewall Guide
* Frequently Asked Questions
* Full Disk Encryption w/Encrypted Boot
* Guides
* Hashboot
* Hexchat
* Infrastructure
* Install LVM LUKS
* Install LVM LUKS on UEFI GPT
* Install Void Linux onto a USB Stick
* Install alongside Arch Linux
* Install android
* Install on a VPS
* Installation
* Installation: UEFI
* Installation On UEFI With MDRAID, via chroot
* Installation on UEFI, via chroot
* Installation via chroot
* Installer Partitioning
* Installing alongside Arch Linux
* Known Issues
* Libc
* Libinput
* Little Helpers
* Locked out - using various system recovery options to reset the password
* Locking the root account and using sudo
* Lutris
* Lyx: the document processor
* MP4 / h.264 in Firefox
[...]

PR #342, opened 2020-06-20
power-management: Add section.

flexibeast:
Partially addresses #327 and #337.

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -30,6 +30,7 @@
       - [AppArmor](./config/security/apparmor.md)
    - [Date and Time](./config/date-time.md)
    - [Kernel](./config/kernel.md)
+   - [Power Management](./config/power-management.md)
    - [Network](./config/network/index.md)
       - [iptables](./config/network/iptables.md)
       - [wpa_supplicant](./config/network/wpa_supplicant.md)
--- a/src/config/power-management.md
+++ b/src/config/power-management.md
@@ -1,0 +1,27 @@
+# Power Management
+
+## acpid
+
+The `acpid` service for [acpid(8)](https://man.voidlinux.org/acpid.8) is
+installed and enabled by default. ACPI events are handled by
+`/etc/acpi/handler.sh`, which uses [zzz(8)](https://man.voidlinux.org/zzz.8) for
+suspend-to-RAM events.
+
+## elogind
+
+The `elogind` service is provided by the `elogind` package. By default,
+[elogind(8)](https://man.voidlinux.org/elogind.8) listens for, and processes,
+ACPI events related to lid-switch activation and the *power*, *suspend* and
+*hibernate* keys. This will conflict with the `acpid` service if it is installed
+and enabled. Either disable `acpid` when enabling `elogind`, or configure
+`elogind` to `ignore` ACPI events in
+[logind.conf(5)](https://man.voidlinux.org/logind.conf.5). There are several
+configuration options, all starting with the keyword `Handle`, that should be
+set to `ignore` to avoid interfering with `acpid`.
+
+## Power Saving - tlp
+
+Laptop battery life can be extended by using
+[tlp(8)](https://man.voidlinux.org/tlp.8). To use it, install the `tlp` package,
+and enable the `tlp` service. Refer to [the TLP
+documentation](https://linrunner.de/tlp/) for details.
--- a/src/config/session-management.md
+++ b/src/config/session-management.md
@@ -23,23 +23,16 @@ software assumes the presence of a session bus.
 
 ## elogind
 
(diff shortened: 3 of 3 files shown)


src/config/power-management.md
```diff
@@ -0,0 +1,29 @@
...
+
+The `acpid` service is installed and enabled by default. ACPI events are handled
+by `/etc/acpi/handler.sh`, which uses [zzz(8)](https://man.voidlinux.org/zzz.8)
+for suspend-to-RAM events.
+
+## elogind
+
+The `elogind` service is provided by the `elogind` package. By default,
+`elogind` listens for and processes ACPI events related to lid-switch activation
+and presses of the *power*, *suspend* and *hibernate* keys. This will conflict
```
ericonr, line 17: "and the ", perhaps?

src/config/power-management.md
```diff
@@ -0,0 +1,29 @@
+# Power Management
+
+Void provides two approaches to power management:
+[acpid(8)](https://man.voidlinux.org/acpid.8) and
+[elogind(8)](https://man.voidlinux.org/elogind.8).
```
ericonr, line 5: Feels kind of weird that this paragraph mentions acpid and elogind, but not TLP. They are separate categories of applications, too, so perhaps we could have `## Power management` and `## Power saving` top level headers? Terrible names, just thinking about it.

The takeaway is that I think it makes sense for acpid and elogind to be inside a header.

src/config/power-management.md
flexibeast: Mm, that sounds more awkward to me.

ericonr: "events related to [...] the *power* [...] keys" and "events related to [...] presses of the *power* [...] keys

src/config/power-management.md
flexibeast: Good points - agreed. i'll see what i can come up with.

src/config/power-management.md
flexibeast: Okay, i'll change the sentence to:

> By default, `elogind` listens for, and processes, ACPI events related to lid-switch activation and the *power*, *suspend* and *hibernate* keys.

flexibeast: PR updated.

src/config/power-management.md
flexibeast: PR updated. i've used the heading titles you suggested as at least placeholders - let me know of any titles you think might be better. :-)

src/config/session-management.md
```diff
@@ -30,16 +30,5 @@ also be one of the mechanisms for rootless [Xorg](./graphical-session/xorg.md).
 To make use of its features, install the `elogind` package and make sure the
 [system D-Bus](#d-bus) is enabled. You might need to log out and in again.
 
-### Troubleshooting
-
 If you're having any issues with `elogind`, [enable](./services/index.md) its
 service, as waiting for a D-Bus activation can lead to issues.
```
ericonr, line 34: Still leave a comment leading back to power management? Like `elogind also does [power management](./power-management.md).`

src/SUMMARY.md
```diff
@@ -30,6 +30,7 @@
       - [AppArmor](./config/security/apparmor.md)
    - [Date and Time](./config/date-time.md)
    - [Kernel](./config/kernel.md)
+   - [Power Management](./config/power-management.md)
```
ericonr, line 33: This should become "Power", but then it's hardly descriptive. Is "Power management and saving" too big/bad?

src/config/session-management.md
ericonr: More importantly, mention that some considerations can be necessary when installing elogind, which I hope leads to people reading about the potential issues.

src/SUMMARY.md
flexibeast: i guess i regard "power saving" as an aspect of "power management". Thinking about it some more, my preferred structure for `power-management.md` would be more like:

    # Power Management
    ## acpid
    ## elogind
    ## Power Saving - tlp

WDYT?

src/config/session-management.md
flexibeast: i've just pushed some changes; let me know what you think.

src/SUMMARY.md
ericonr: That works, I think.

ericonr: If we decide to document more power saving stuff, they can get sublevel headers later.

src/config/session-management.md
ericonr: Looks good. Thanks c:

src/SUMMARY.md
flexibeast: Thanks, PR updated.

PR #297, opened 2020-05-21
Remove block quotes from all documents, except actual quotes

jeffayle:
Reflects recent changes to the style guide, this PR removes every blockquote that was being used as an aside. Every remaining blockquote is a legitimate use of the tag. #290 

```
$ grep -R '^> ' src/
src/contributing/void-docs/submitting.md:> The commit message should be in the form of `section: what was changed`
src/contributing/void-docs/style-guide.md:> For example:
src/contributing/void-docs/style-guide.md:> 
src/contributing/void-docs/style-guide.md:> `$ ls -l`
src/contributing/void-docs/style-guide.md:> Now we need to install the CUPS drivers and configure them.
src/contributing/void-docs/style-guide.md:> Install and configure the CUPS drivers, then configure them as shown.
src/contributing/void-docs/style-guide.md:> You can also use program X for this purpose.
```

I have not made any decisions regarding "should only be used sparingly, and for non-critical information". Since this commit is essentially a list of every "Note" currently in the docs, this might be a good place to make decisions about what should be labeled a note and what shouldn't.

PR #343, opened 2020-06-20
xbps/<various>: Multiple changes.

flexibeast:
* Move `packages/*` up a level, remove `packages` subdir.
* Consolidate information.
* Remove `building.md` ,`management.md`, `packages/index.md`.
* Add links to man.voidlinux.org.
* Remove unnecessary `xbps-install` code block.
* Shorten `xlocate` example.
* Add example of how to list installed packages.
* Improve prose.

Partially addresses #337.


src/config/locales.md
```diff
@@ -9,8 +9,8 @@ $ locale -a
 ## Enabling locales
 
 To enable a certain locale, un-comment or add the relevant lines in
-`/etc/default/libc-locales` and [reconfigure](../xbps/packages/management.md)
-the `glibc-locales` package.
+`/etc/default/libc-locales` and reconfigure the `glibc-locales` package using
+[xbps-reconfigure(1)](https://man.voidlinux.org/xbps-reconfigure.1).
```
ericonr, line 13: The reason I like to link to the reconfiguration page is that it mentions the need for the `-f` flag.

src/xbps/updating.md
```diff
@@ -1,47 +1,37 @@
...
 ```
 # xbps-install -Su
 ```
 
-> Note: XBPS must use a separate transaction to update itself. If your first
-> update includes the package `xbps`, you will need to run an additional update
-> for the rest of the system.
+Note: XBPS must use a separate transaction to update itself. If your update
+includes the `xbps` package, you will need to run the above command a second
+time to apply the rest of the updates.
```
ericonr, line 12: As I mentioned in #297, I believe this shouldn't be a note.

src/xbps/updating.md
```diff
@@ -1,47 +1,37 @@
...
-```
-
 ```
 $ xcheckrestart
 11339 /opt/google/chrome/chrome (deleted) (google-chrome)
 ```
 
 `xcheckrestart` will print out the PID, path to the executable, status of the
-path that was launched (almost always deleted) and the process name.
+path that was launched (almost always 'deleted') and the process name.
```
ericonr, line 29: single quotes or an inline code block?

src/config/locales.md
flexibeast: Ah, fair point. i'll change this back.

src/xbps/updating.md
flexibeast: Fixed.

src/xbps/updating.md
flexibeast: i wasn't sure about this one; i'm happy to change it to an inline code block.

flexibeast:
@ericonr: Thanks for the review; PR updated.

src/SUMMARY.md
```diff
@@ -61,11 +61,8 @@
    - [Microcode](./config/microcode.md)
 - [XBPS Package Manager](./xbps/index.md)
    - [Updating](./xbps/updating.md)
-   - [Packages](./xbps/packages/index.md)
-      - [General](./xbps/packages/management.md)
-      - [Downgrading](./xbps/packages/downgrading.md)
-      - [Building](./xbps/packages/building.md)
-      - [Finding Files](./xbps/packages/files.md)
+   - [Downgrading](./xbps/downgrading.md)
```
ericonr, line 64: Should the downgrading title be the same as the one in the page?

src/xbps/files-and-packages.md
```diff
@@ -0,0 +1,35 @@
...
+Fetching objects: 11688, done.
+From https://alpha.de.repo.voidlinux.org/xlocate/xlocate
+ + e122c3634...a2659176f master     -> master  (forced update)
+$ xlocate xlocate
+xtools-0.59_1   /usr/bin/xlocate
+xtools-0.59_1   /usr/share/man/man1/xlocate.1 -> /usr/share/man/man1/xtools.1
+```
+
+It is also possible to use
+[xbps-query(1)](https://man.voidlinux.org/xbps-query.1) to find files, though
```
ericonr, line 19: though -> although ?

src/xbps/index.md
```diff
@@ -1,5 +1,33 @@
...
+
+Most general package management is done with the following commands:
+
+- [xbps-query(1)](https://man.voidlinux.org/xbps-query.1) searches for and
+   displays information about packages installed locally, or, if used with the
+   `-R` flag, packages contained in repositories.
+- [xbps-install(1)](https://man.voidlinux.org/xbps-install.1) installs and
+   updates packages, and syncs repository indexes.
+- [xbps-remove(1)](https://man.voidlinux.org/xbps-remove.1) removes installed
+   packages, and can remove cached package files.
```
ericonr, line 15: Mention it can remove orphan packages as well?

src/xbps/updating.md
```diff
@@ -1,47 +1,37 @@
...
 
-Your system likely ran out of space in `/boot`. XBPS installs kernels and
-requests that hooks such as DKMS and Dracut be run, but it doesn't remove
-kernels that are obsolete. This is left as a task for the administrator to
-permit the retention of obsolete but still booted or known working kernels.
-
-Remove kernels with [vkpurge(8)](https://man.voidlinux.org/vkpurge.8).
+If you get a kernel panic after an update, it is likely your system ran out of
+space in `/boot`. Refer to "[Removing old
+kernels](../config/kernel.md#removing-old-kernels)" for further information.
```
ericonr, line 37: If this is the recommended style, I think it should be in the style guide. Usually I don't use quotes when linking to a page using its title, but I think it does look nicer.

src/SUMMARY.md
flexibeast: It should indeed. :-) But i think the title of the page should be changed to simply "Downgrading", so i've just done that, and added an introductory sentence.

src/xbps/files-and-packages.md
flexibeast: i considered that, and "though" sounds better to me; but if you'd prefer "although", i can certainly change it.

src/xbps/index.md
flexibeast: Done.

src/xbps/updating.md
flexibeast: Mm, i think it depends on the specific context. If literally referring to a section, then yes, it should be in quotes. If the text is not doing that, then it doesn't need to be in quotes. For example:

> For more information, please read the "[Power Management](...)" section.

vs
> Void provides facilities to assist with [power management](...).

If that makes sense to you, i'll update the style guide accordingly.

src/xbps/files-and-packages.md
ericonr: I am not entirely sure. If you're okay with "though", it's good for me.

src/xbps/updating.md
ericonr: Definitely! Thanks c:

PR #344, opened 2020-06-20
kernel: Multiple changes.

flexibeast:
* Add information about newer-than-default kernels.
* Describe how to build DKMS modules for newer-than-default kernels.
* Fix typos.

Partially addresses #337.

### DIFF
--- a/src/config/kernel.md
+++ b/src/config/kernel.md
@@ -3,15 +3,21 @@
 ## Kernel series
 
 Void Linux provides many kernel series in the default repository. These are
-named `linuxX.X`, for example `linux4.19`. You can query for all available
+named `linuxX.Y`: for example, `linux4.19`. You can query for all available
 kernel series by running:
 
 ```
 $ xbps-query --regex -Rs '^linux[0-9.]+-[0-9._]+'
 ```
 
-The `linux` meta package which is installed by default depends on one of the
-kernel packages, usually the latest mainline kernel that works with all dkms
+The `linux` meta package, installed by default, depends on one of the kernel
+packages, usually the latest mainline kernel that works with all DKMS modules.
+Newer kernels might be available in the repository, but are not necessarily
+considered stable enough to be the default; use these at your own risk. If you
+wish to use a more recent kernel and have DKMS modules that you need to build,
+install the relevant `linuxX.Y-headers` package, then use
+[xbps-reconfigure(1)](https://man.voidlinux.org/xbps-reconfigure.1) to
+reconfigure the `linuxX.Y` package you installed. This will build the DKMS
 modules.
 
 ## Removing old kernels
@@ -102,7 +108,7 @@ kernel series is installed for the first time or updated, but can also be run
 manually:
 
 ```
-# xbps-reconfigure --force linuxX.X
+# xbps-reconfigure --force linuxX.Y
 ```
 
 If run manually, they serve to apply initramfs configuration changes to the next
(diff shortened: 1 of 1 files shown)


ericonr:
Needs some more changes, if you could put them in this PR too c:

```
      ericonr │ duncaen: mind explaining what's wrong in https://github.com/void-linux/void-docs/commit/e478095076155005cc24c9cb7aecf6449979f9c4#r40050286 ? I don't think we had any complaints about that
      ericonr │ I can open an issue about it, after I know what's wrong
      duncaen │ I wanted to link to it for the ""How do I update grub" PR I just closed
      duncaen │ but its wrong
      duncaen │ xbps-query -p install-script grub
      duncaen │ first it would need to force reconfigure
      duncaen │ but also, there is nothing actually regenerating the config in grubs configure scripts
      ericonr │ I see
      ericonr │ it seems the grub kernel hook would regenerate the config, right?
      ericonr │ though I guess that's not really the correct way to go about it
      duncaen │ yes you could reconfigure a kernel package, but yea that kinda useless
      duncaen │ update-grub or what cat /usr/bin/update-grub does
```

flexibeast:
To clarify, instead of:

> then reconfiguring the `grub` package.

it should say:

> then running:
> ```
> # grub-mkconfig -o /boot/grub/grub.cfg
> ```

? Am i understanding correctly?

ericonr:
I think running `update-grub` is better, since it does just that. But yes, reconfiguring the grub package isn't enough.

flexibeast:
@ericonr: Thanks, PR updated.

src/config/kernel.md
```diff
@@ -3,16 +3,21 @@
 ## Kernel series
 
 Void Linux provides many kernel series in the default repository. These are
-named `linuxX.X`, for example `linux4.19`. You can query for all available
+named `linuxX.X`: for example, `linux4.19`. You can query for all available
```
ericonr, line 6: I'm always unsure. Should we prefer `linuxX.Y` or `linuxX.X`?

src/config/kernel.md
```diff
@@ -126,7 +131,7 @@ repositories.
 
 Kernel command line arguments can be added through the GRUB bootloader by
 editing `/etc/default/grub`, changing the `GRUB_CMDLINE_LINUX_DEFAULT` variable
-and then reconfiguring the `grub` package.
+and then running `update-grub`.
```
ericonr, line 134: I hate that we can't link to a man page here :p

src/config/kernel.md
flexibeast: Fair point - i can't see any reason *not* to change it to `X.Y`, so i'll do that.

src/config/kernel.md
```diff
@@ -3,16 +3,21 @@
...
-kernel packages, usually the latest mainline kernel that works with all dkms
-modules.
+The `linux` meta package, installed by default, depends on one of the kernel
+packages, usually the latest mainline kernel that works with all DKMS modules.
+Newer kernels might be available in the repository, but are not necessarily
+considered stable enough to be the default; use these at your own risk. If you
+wish to use a more recent kernel and have DKMS modules that you need to build,
+install the relevant `linuxX.Y-headers` package, then use
+[xbps-reconfigure(1)](https://man.voidlinux.org/xbps-reconfigure.1) to
+reconfigure the `linuxX.Y` package you installed.
```
ericonr, line 20: Make it clearer that the reconfiguration step will build the DKMS modules? Not sure if it's needed.

flexibeast: i don't think it'll hurt, so i've just added "This will build the DKMS modules" to the end of that para.

flexibeast:
@ericonr: Thanks for the review. Merging.

PR #349, opened 2020-06-22
about/index: Multiple changes.

flexibeast:
* Mention Void focusing on stability rather than on being bleeding-edge.
* Mention runit being used for service supervision as well as init.
* Mention origin of the "Void" name.

Partially addresses #337.

### DIFF
--- a/src/about/index.md
+++ b/src/about/index.md
@@ -6,8 +6,9 @@ documentation effectively.
 
 Void is an independent, [rolling
 release](https://en.wikipedia.org/wiki/Rolling_release) Linux distribution,
-developed from scratch and not as a fork of any other distribution. There are a
-few features that make Void unique:
+developed from scratch rather than as a fork, with a focus on stability over
+[bleeding-edge](https://en.wikipedia.org/wiki/Bleeding_edge_technology). In
+addition, there are several features that make Void unique:
 
 - The [XBPS](https://github.com/void-linux/xbps) package manager, which is
    extremely fast, developed in-house, and performs checks when installing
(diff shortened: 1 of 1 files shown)


src/about/index.md
```diff
@@ -6,8 +6,10 @@ documentation effectively.
 
 Void is an independent, [rolling
 release](https://en.wikipedia.org/wiki/Rolling_release) Linux distribution,
-developed from scratch and not as a fork of any other distribution. There are a
-few features that make Void unique:
+developed from scratch and not as a fork of any other distribution. In addition
```
bobertlo, line 9: maybe?: not a fork [..] with a focus on stability over bleeding edge. In addition [...]

src/about/index.md
```diff
@@ -22,12 +24,15 @@ few features that make Void unique:
...
 - [runit](../config/services/index.md) is used for
-   [init(8)](https://man.voidlinux.org/init.8). This allows Void to support musl
-   as a second libc choice, which would not be possible with
-   [systemd](https://www.freedesktop.org/wiki/Software/systemd/). A side effect
-   of this decision is an init system with clean and efficient operation, and a
-   small code base.
+   [init(8)](https://man.voidlinux.org/init.8) and service supervision. This
+   allows Void to support musl as a second libc choice, which would not be
+   possible with [systemd](https://www.freedesktop.org/wiki/Software/systemd/).
+   A side effect of this decision is an init and service supervision system with
```
bobertlo, line 30: init and service supervision seems redundant here? i can't think of a great alternative at this moment.

flexibeast: i've added this due to a discussion with someone who thought that runit was only used for init, rather than service supervision. (i guess this is what happens when the arguments around systemd get framed as being a debate about init systems.)

src/about/index.md
flexibeast: Yeah, that's definitely better .... How about "Void is an independent, rolling release Linux distribution, developed from scratch rather than as a fork, with a focus on stability over bleeding-edge. In addition, there are ..."?

src/about/index.md
bobertlo: My comment only addresses the duplication of "init" and "service supervision system". I wonder if the second mention could just say "core system with [...]"? I'm not sure just a guess.

src/about/index.md
bobertlo: Yeah.

src/about/index.md
flexibeast: oic. Sure, PR updated.

src/about/index.md
flexibeast: Thanks, PR updated.

ericonr:
I like this! Thanks c:

Issue #266, opened 2020-05-10
iptables and nftables


Duncaen:
Nftables is currently completely missing.

The current iptables documentation seems low quality imho.

* The section [IP6 firewall rules](https://docs.voidlinux.org/config/network/iptables.html#ip6-firewall-rules) could be merged with "Applying the rules at boot", the `iptables-runit` service works already for both, the snipped for `/etc/rc.local` could do exactly the same.
* [Firewall rules](https://docs.voidlinux.org/config/network/iptables.html#firewall-rules) and its subsection [Adjusting the rules](https://docs.voidlinux.org/config/network/iptables.html#adjusting-the-rules) explain how to edit files, which shouldn't be necessary. They also miss to document the file names of the rules which would be loaded on boot with the `rc.local` snipped, the runit "service" and the `iptables-runit` core service.

ericonr:
The only firewall I've ever used was ufw, so I don't think I'm qualified to write about them. I can do some research, however, if no one manages to write this one.

The stylistic part I can do, but if anyone wants to claim it, I'd be happy to help them!

PR #352, opened 2020-06-24
iptables: Multiple changes.

flexibeast:
* Rename to 'firewalls.md`, to prepare for `nftables` content.
* Reorganise and rewrite section contents:
  - Include reference to iptables/ip6tables man page.
  - Remove extraneous code blocks.
  - Note that `rc.local` is called in runit stage 2.
  - Use contents of `91-iptables.sh` as code for `rc.local`.
  - Add `ip6tables -L` to example code block.
  - Add examples of flushing and restoring rules via runiservice.

Partially addresses #266.


ericonr:
Can I merge? LGTM

src/config/network/firewalls.md
```diff
@@ -0,0 +1,50 @@
...
+```
+# iptables -L
+# ip6tables -L
+```
+
+### Applying the rules at runtime
+
+`iptables` comes with a runit service to quickly flush or restore the rules on
+demand. Once the service is [enabled](../services/index.md#enabling-services),
+you can flush the rules with:
```
ericonr, line 40: Which rules does this use? If it's the ones mentioned in the section above, it might make sense to specify the rules used in the main `## iptables` part. Something like "The rules used in the solutions below are ..."

flexibeast: Good point. i've just updated the PR; let me know what you think.

ericonr: LGTM

PR #363, opened 2020-06-28
Add instructions for Installing the NVIDIA CUDA toolkit

chili-b:
There are some problems with the installer reporting missing ncurses libraries even if all the ncurses packages are installed. I couldn't find any writing about this anywhere else, so I wrote up the few things I did to make it work. Maybe this is the wrong place to put these notes, but I think they are helpful, so if that is the case please point me to a more appropriate place.

### DIFF
--- a/src/theme/css/general.css
+++ b/src/theme/css/general.css
@@ -276,7 +276,7 @@ body {
 /* 300px + 800px + 2*90px + 15px */
 @media only screen and (min-width: 1295px) {
 	.sidebar-visible #nav-wide-wrapper {
-		max-width: auto;
+		max-width: none;
 		margin: 0;
 	}
 	.sidebar-visible .nav-chapters {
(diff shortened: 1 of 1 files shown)


flexibeast:
Thanks for your contribution!

@st3r4g, @ericonr: Thoughts on this?

ericonr:
I was the one who suggested on IRC that they try to write something about this issue, that way we could discuss inclusion. On the one hand, it is useful info for users, and doesn't seem entirely too well documented. On the other hand, it might lead up to the external applications page being a big list of tricks for dealing with external stuff.I t isn't entirely bad, but it would be hard to maintain.

flexibeast:
@ericonr: Well, info about CUDA is on [the list of wiki pages that possibly need inclusion in the Handbook](https://github.com/void-linux/void-docs/issues/327); i've mentioned this PR there. My thought is that CUDA content could accompany the other NVIDIA sections, rather than going in the "External Applications" section. (E.g. we could reorganise the existing NVIDIA sections to be under a general "NVIDIA" section, and CUDA content could go on the latter 'intro' page.)

st3r4g:
why not packaging this instead?

flexibeast:
@st3r4g: No reason - if this is something that can be addressed by packaging, then i'm all for it. :-)

Issue #358, opened 2020-06-26
Ensure relevant content from `usage/xbps/` on Web site is covered by Handbook


flexibeast:
Needed to partially address #329.

[usage/xbps/](https://voidlinux.org/usage/xbps/):

- [x] Introduction
  Covered by "XBPS Package Manager" section of Handbook.
- [x] Repositories
  Covered by "Custom Repositories" section of Handbook.
- [x] Official Repositories
  Covered by "Official Repositories" subsection of Handbook.
- [x] Official Repositories (musl)
   #394 (merged)
- [x] Subrepositories
  Covered by "Repositories" section of Handbook.
- [x] Download static binaries
  Covered by "Static XBPS" section of Handbook.
- [x] xbps-install(1)
  Covered by man page and "XBPS Package Manager" section of Handbook. 
- [x] Synchronize remote repository data
  Covered by man page.
- [x] Installing/updating a single package
  Covered by man page.
- [x] Reinstalling/downgrading to a specific package version
  Covered by man page and "Downgrading" subsection of Handbook.
- [x] Updating your system
  Covered by man page and "Updating" subsection of Handbook.
- [x] xbps-query(1)
  Covered by man page and "XBPS Package Manager" section of Handbook.
- [x] Listing registered repositories
  Covered by man page.
- [x] Listing installed packages
  Covered by man page.
- [x] Showing information for a package
  Covered by man page and "XBPS Package Manager" section of Handbook.
- [x] Showing the files list for a package
  Covered by man page.
- [x] Showing the required dependencies for a package
  Covered by man page.
- [x] Showing the reverse dependencies for a package (packages that depend on it):
  Covered by man page.
- [x] Searching for packages matching its package name/version and/or description
  Covered by man page.
- [x] Searching for packages matching a filename
  Covered by man page.
- [x] xbps-remove(1)
  Covered by man page and "XBPS Package Manager" section of Handbook.
- [x] Removing a single package
  Covered by man page.
- [x] Removing a single package and recursively all packages that were installed as dependencies
  Covered by man page.
- [x] Cleaning up the cache directory
  Covered by man page.
- [x] Removing all package orphans
  Covered by man page.
- [x] Removing all package orphans and clean the cache directory
  Covered by man page.
- [x] xbps-reconfigure(1)
  Covered by man page and "XBPS Package Manager" section of Handbook.
- [x] Configure a package that is in unpacked state
  Covered by man page.
- [x] Configure all packages that are in unpacked state
[...]

Issue #201, opened 2020-04-20
Document non obvious package management mechanisms


bobertlo:
You can create a virtualpkg to install a different version of the kernel without breaking dependencies. There should be a cannonical method documented for this.

`virtualpkg=linux:linux5.6` in `/etc/xbps/newerkernel.conf?

ericonr:
I'm highjacking this issue to include everything related to XBPS that it might not be obvious that the option even exists:

- virtualpkg
- ignorepkg (less useful, this one)
- mention vkpurge (debatable)
- ...

flexibeast:
@ericonr: With `ignorepkg` and `virtualpkg` now mentioned in the "Advanced Usage" section, and `vkpurge` mentioned in the "Kernel" section, i'd like to close this. Any objections?

ericonr:
No problem! Closing now.

PR #368, opened 2020-06-29
microcode: Rename to `firmware`, expand section.

flexibeast:
* Mention how to find available firmware.
* Mention how to remove unneeded firmware.

Addresses #129.

@Duncaen, is this what you were after?


src/config/firmware.md
```diff
@@ -21,7 +21,24 @@ Install the AMD package, `linux-firmware-amd`, which contains microcode for both
...
 
 The `/proc/cpuinfo` file has some information under `microcode` that can be used
 to verify the microcode update.
+
+## Adding firmware
+
+A list of firmware available in the Void repositories can be found by running:
+
+```
+$ xbps-query -Rs firmware | awk '{ print $2 }' | grep firmware
```
ericonr, line 34: I think `xbps-query -p pkgver -Rs firmware | awk '{ print $2 }'` is a bit more explicit, what do you think?

src/config/firmware.md
```diff
@@ -1,12 +1,12 @@
-# Microcode
+# Firmware
```
ericonr, line 1: I think we need to mention that some firmware packages might only be available from nonfree repos.

src/config/firmware.md
```diff
@@ -21,7 +21,24 @@ Install the AMD package, `linux-firmware-amd`, which contains microcode for both
...
+```
+
+## Removing firmware
+
+By default, `linuxX.Y` packages and the `base-system` package install a number
+of firmware packages. To remove unneeded firmware packages, use `ignorepkg`
+entries in an XBPS configuration file, as described in
+[xbps.d(5)](https://man.voidlinux.org/xbps.d.5), to specify which packages are
+unneeded, then use [xbps-remove(1)](https://man.voidlinux.org/xbps-remove.1) to
+remove them.
```
ericonr, line 44: I feel like this needs to make it clearer that it is not necessary to remove the firmware that aren't in use.

And the explanation regarding `xbps.d` seems like something we could mention in the XBPS pages instead? Then make a link here.

src/config/firmware.md
ericonr: And this sort of trick could perhaps be mentioned in XBPS pages too.

flexibeast: That's definitely better; i was using this because `xbps-query -Rs firmware -p pkgname` didn't work. What am i misunderstanding?

ericonr: I'm not sure eithter... I think some of the properties can't be queried for.

I couldn't query: pkgname, alternatives, automatic-install, install-date, install-script, metafile-sha256, repository and state. There's probably some difference in how these properties are generated/read from the packages.

src/config/firmware.md
flexibeast: i'll certainly mention the first point.
Re. the second point, probably, but whereabouts specifically?

ericonr: Not sure how big of a change this would be, but perhaps rename `files-and-packages` from "Finding..." to just "Files and Packages" and adding some tricks there?

flexibeast: Hmm, how about "General Package Management" instead? Then it could be the place to address both #201 and #358.

ericonr: The one reason I don't like this as an option was because index already listed the `xbps-*` utilites, which are the main part of package management.

src/config/firmware.md
```diff
@@ -1,12 +1,16 @@
-# Microcode
+# Firmware
+
+Void provides a number of firmware packages in the repositories. Some firmware
+is only available in the [nonfree](../xbps/repositories/index.md#nonfree)
```
ericonr, line 4: "is only available if you have enabled the [nonfree] ..."

What do you think?

src/config/firmware.md
flexibeast: Following our discussions on IRC, i've now opened #373.

ericonr: Thanks! I think you need to link to that and this will be good to go too c:

flexibeast: With #373 now merged, i've updated this PR to refer to the subsection added, about how to ignore packages.

ericonr:
LGTM, thanks c:

flexibeast:
i'd now like to wait for an ok from @Duncaen, unless he's already done so via IRC?

ericonr:
Not that I've seen, no. OK!

Duncaen:
lgtm, not sure if the command to search for packages with firmware in the name is necessary.

flexibeast:
@Duncaen: Thanks; i've removed the subsection about finding available firmware. Merging.

ericonr:
We could have that documented in advanced usage, perhaps?

PR #373, opened 2020-07-01
package-management: Mention how to ignore packages.

flexibeast:
Needed for #368; partially addresses #201.

### DIFF
--- a/src/xbps/package-management.md
+++ b/src/xbps/package-management.md
@@ -111,3 +111,22 @@ To get a list of all installed packages, without their version:
 ```
 $ xbps-query -l | awk '{ print $2 }' | xargs -n1 xbps-uhelper getpkgname | fmt
 ```
+
+## Ignoring Packages
+
+Sometimes you may wish to remove a package whose functionality is being provided
+by another package, but will be unable to do so due to dependency issues. For
+example, you may wish to use [doas(1)](https://man.voidlinux.org/doas.1) instead
+of [sudo(8)](https://man.voidlinux.org/sudo.8), but will be unable to remove the
+`sudo` package due to it being a dependency of the `base-system` package. To
+remove it, you will need to *ignore* the `sudo` package.
+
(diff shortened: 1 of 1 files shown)


src/xbps/package-management.md
```diff
@@ -111,3 +111,22 @@ To get a list of all installed packages, without their version:
 ```
 $ xbps-query -l | awk '{ print $2 }' | xargs -n1 xbps-uhelper getpkgname | fmt
 ```
+
+## Ignoring Packages
+
+Sometimes you may wish to remove a package whose functionality is being provided
+by another package, but will be unable to do so due to dependency issues. For
```
ericonr, line 118: "functionality is being provided by another package" is not the only case of dependency removal being valid. I'm not sure we want to mention other cases, though, because we don't want to support people stripping down their systems unnecessarily.

src/xbps/package-management.md
```diff
@@ -111,3 +111,22 @@ To get a list of all installed packages, without their version:
...
+
+To ignore a package, add an appropriate `ignorepkg` entry in an
+[xbps.d(5)](https://man.voidlinux.org/xbps.d.5) configuration file. For example:
+
+```
+ignorepkg=sudo
+```
+
+You could then remove the `sudo` package using
+[xbps-remove(1)](https://man.voidlinux.org/xbps-remove.1).
```
ericonr, line 132: "You will then be able to remove..."

src/xbps/package-management.md
flexibeast: Agreed on both points. So leave it as-is?

ericonr: I think so, yeah.

flexibeast: Thanks. i've amended the PR to address your other comment. Ok to merge?

ericonr: Ok! Thanks c

PR #378, opened 2020-07-02
wpa_supplicant: Mention need to enable service in opening paragraph.

flexibeast:


flexibeast:
CI check failed due only to linkcheck failing for infradocs.voidlinux.org, so i'm merging this.

PR #379, opened 2020-07-03
power-management: acpid is not always enabled.

ericonr:
Not sure this is the best way to phrase it, just wanted to get started on it.

### DIFF
--- a/src/config/power-management.md
+++ b/src/config/power-management.md
@@ -3,9 +3,9 @@
 ## acpid
 
 The `acpid` service for [acpid(8)](https://man.voidlinux.org/acpid.8) is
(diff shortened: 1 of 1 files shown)


flexibeast:
Is it only enabled by default when installing from an ISO, or are there other cases as well? (Or does whether or not it's enabled depend on the flavour as well?)

ericonr:
I actually don't know :p 

I can try to take a look at how mklive is configured for this, but I'm not sure we should try to list where it happens and where it doesn't.

ericonr:
Updated.

flexibeast:
LGTM.

ericonr:
Thanks!

PR #380, opened 2020-07-03
podman.md: Added containers and VMs section with starting podman doc …

Frick-David:
I am looking for a couple suggestions regarding this doc.
1) Should I add a section about the podman man pages? Seems like a good move.
2) Is it okay that the podman command to test the installation is something relating to fedora. It seems to be more an apache web server, and it is the one in the podman docs on their website. Did not know if the void team wanted it to be something a little more distro agonist.
   -- NOTE to self: add the command to stop the test container --> `sudo podman stop -l`
3) I do not like the error section in this. Saying to run a command as sudo to fix an error makes my blood pressure raise at least 10 points. I will absolutely edit that section (i see it more of a placeholder). I could use some guidance on what to put. The podman doc says
```
The code samples are intended to be run as a non-root user, and use sudo where root escalation is required.
```
They list the command for the test container as 
```
$ podman run -dt -p 8080:8080/tcp registry.fedoraproject.org/f29/httpd
```
Emphasis on the $. Note that command does return the documented error on my system. Their doc makes no reverence to editing permissions, groups etc to allow podman to run the above command as user. I could use some guidance on cleaning up that section.

After I have feedback, I will edit accordingly and resubmit!

### DIFF
--- a/src/config/containers_and_vms/podman.md
+++ b/src/config/containers_and_vms/podman.md
@@ -1,0 +1,33 @@
+# Podman
+
+**Podman** is a daemonless engine for developing, managing, and running pods,
+containers, and container images.
+
+## Installation
+
+Install the `podman` package.
+
+## Checking installation
+
+To check your installation of podman, run the following command:
+
+`$ podman run -dt -p 8080:8080/tcp registry.fedoraproject.org/f29/httpd`
+
+Then, open a Internet browser and navigate to `0.0.0.0:8080` in the address bar.
+If you see a message about a test page and apache server running, podman is
+running correctly.
+
+To see more podman commands, and walk through basic tutorials, see the [getting
+started](https://podman.io/getting-started/) page.
+
+## Troubleshooting
+
+### OCI runtime error
+
+You may receive the following message when you issue a podman command:
+
+```
+Error: container_linux.go:349: starting container process caused "process_linux.go:449: container init caused \"process_linux.go:378: setting rlimits for ready process caused \\\"error setting rlimit type 7: invalid argument\\\"\"": OCI runtime error
+```
+
+If this occurs, rerun the command with `sudo` privileges.
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -50,6 +50,8 @@
       - [Configuring a New Printer](./config/print/config.md)
    - [Manual Pages](./config/man.md)
    - [Microcode](./config/microcode.md)
+   - [Containers and VMs](./config/containers_and_vms/index.md)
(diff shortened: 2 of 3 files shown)


flexibeast:
As much as possible, we should be trying to combine things into a single page; i've been doing a lot of consolidation of late, to try to reduce the number of short pages (e.g. the "XBPS Package Manager" section, which has been heavily consolidated). @ericonr and i feel that around 150 lines is the threshold at which we should start considering whether or not to split the page. Would combining the current Docker and Podman content exceed that threshold?

> Should I add a section about the podman man pages? Seems like a good move.

Rather than having a distinct subsection, there should be links to the relevant man pages as part of the text, e.g.

    To remove a package, use [xbps-install(1)](https://www.voidlinux.org/xbps-install.1):

    # xbps-install <unneeded-package>

> Is it okay that the podman command to test the installation is something relating to fedora. It seems to be more an apache web server, and it is the one in the podman docs on their website. Did not know if the void team wanted it to be something a little more distro agonist.

Is that the only way to test the installation? If so, that's fine; if not, i feel the command should be something more general.

As a general point, don't use `sudo` in commands; instead, use `#` or `$` to indicate whether or not the command needs to be run with root privileges (cf. [the style guide](https://docs.voidlinux.org/contributing/void-docs/style-guide#commands.html)). Unfortunately, i don't have any experience with Podman to know how to best handle the specific error issue you mentioned.

Frick-David:
Fair enough on the `sudo`. I have forgotten about that! that is an easy fix. Let me look into some other basic commands. That was the one off the website but you are right there has to be a similar and more direct one. Docker had something that printed hello world from docker so let me look for something like that.

I think docker and podman together might exceed that threshold just a little bit, but let me check on that. If it is a little over, they probably could still go together

src/config/containers_and_vms/podman.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Installation
+
+Install the `podman` package.
+
+## Checking installation
+
+To check your installation of podman, run the following command:
+
+`$ podman run -dt -p 8080:8080/tcp registry.fedoraproject.org/f29/httpd`
```
sgn, line 14: I think it's better to use code block instead of inline code?

```
    ```sh
    $ podman run -dt ....
    ```
```

Frick-David: Okay, will change it

Duncaen:
Is there anything void specific to document? Otherwise I don't really see the need to document how to use the podman command.

Frick-David:
It is less about how to use the command and more about testing that podman is indeed working.

PR #384, opened 2020-07-05
xbps/<various>: Reorganise section.

flexibeast:
* Move most of `package-management` into `index`.
* Rename `package-management` to `advanced-usage`.


src/xbps/index.md
```diff
@@ -31,3 +31,115 @@ together with the [xbps.d(5)](https://man.voidlinux.org/xbps.d.5) man page.
...
+```
+
+Then downgrade with `xbps-install`:
+
+```
+# xbps-install -R /var/cache/xbps/ -f pkg-1.0_1
+```
+
+The `-f` flag is necessary to force downgrade/re-installation of an already
+installed package.
```
ericonr, line 109: I would put downgrading in advanced usage, I think. What do you think?

src/xbps/index.md
```diff
@@ -31,3 +31,115 @@ together with the [xbps.d(5)](https://man.voidlinux.org/xbps.d.5) man page.
...
+```
+
+This requires `xbps-query` to download parts of every package to find the file.
+`xlocate`, however, queries a locally cached index of all files, so no network
+access is required.
+
+To get a list of all installed packages, without their version:
+
+```
+$ xbps-query -l | awk '{ print $2 }' | xargs -n1 xbps-uhelper getpkgname | fmt
```
ericonr, line 144: This feels kind of hacky, I would like XBPS to have this functionality.

That said, do we need `fmt` in the end?

flexibeast: Agreed!

No, we probably don't need `fmt`; that was just included in the command i moved from the old wiki. Without it, the output is considerably less compressed, but people can use pipe the output to `less` if they want. `fmt` removed.

PR #386, opened 2020-07-06
check.sh: Add script.

flexibeast:


check.sh
```diff
@@ -0,0 +1,76 @@
...
+
+    if [ ! $NO_XBPS ]; then
+        echo "\nTrying to install '$PKG' ... "
+        if [ ! -z $SU_CMD ]; then
+            $SU_CMD xbps-install $PKG
+
+            # 'xbps-install' exits with 0 when the user aborts, so we can't
+            # check return status; use 'which' to check status instead.
+
+            if ( ! which mdbook 2>&- 1>&- ); then
```
ericonr, line 13: I think this which is checking for the wrong package, isn't it?

check.sh
```diff
@@ -0,0 +1,76 @@
...
+        fi
+    else
+        echo "Can't find 'xbps-install'; please install '$PKG' manually."
+        exit 1
+    fi
+
+}
+
+echo -n "Checking for mdbook ... "
+if ( ! which mdbook 2>/dev/null ); then
```
ericonr, line 29: I prefer to use `command -v` instead of `which`, since it's a shell built-in.

check.sh
```diff
@@ -0,0 +1,76 @@
...
+
+if [ $INSTALL_LINKCHECK ]; then
+    PKG='mdbook-linkcheck'
+    try_installing
+fi
+
+if [ $INSTALL_VMDFMT ]; then
+    PKG='vmdfmt'
+    try_installing
+fi
```
ericonr, line 73: This one can end up running a `xbs-install` command for each package, right? Couldn't we put everything into a single command, and then in the end of `try_installing` check if all programs now exist?

check.sh
```diff
@@ -0,0 +1,76 @@
...
+
+vmdfmt -l -w src/
+mdbook build
```
ericonr, line 76: One can also run `mdbook-linkcheck -s`, to avoid having to install `mdBook`. Do you think users should attempt to build the whole handbook, then?

src/contributing/void-docs/submitting.md
```diff
@@ -31,11 +31,20 @@ $ cd void-docs
...
 
-Once you are satisfied with your changes, commit them and push them to the
-forked repository:
+## Making changes
+
+To serve the docs locally and view your changes, run `mdbook serve` from the
+root of the repository.
+
+Once you are satisfied with your changes, run the `check.sh` script provided in
+the repository root. This will run the `vmdfmt` command, which will flow the
```
ericonr, line 42: flow -> wrap

check.sh
flexibeast: No; the *executable* is `mdbook`, even though the *package* is `mdBook`.

check.sh
flexibeast: Oh, i didn't know one could do that .... Well, i guess my question would be, might there be issues picked up by doing a build? If you feel `vmdfmt` and `mdbook-linkcheck` cover the possible issues, i'm happy to change this.

src/contributing/void-docs/submitting.md
flexibeast: Changed.

check.sh
flexibeast: Oh, sorry, i misunderstood. Fixed.

check.sh
flexibeast: Sure, i'll see what i can do.

check.sh
flexibeast: i was having difficulties controlling the output of `command -v` as i wanted, particularly when checking for `xbps-install`, for which i didn't see the need to have a "Checking ..." line. But i suppose having the latter would be a good idea, so i'll change all uses of `which` to `command -v`.

check.sh
ericonr: It can work as a smoketest, since if it errors out we can catch if before it happens in deployment. We can perhaps make it conditional? Force installation of `mdbook-linkcheck`, and if the person has `mdbook` as well, run `mdbook build`, else run `mdbook-linkcheck -s`.

flexibeast: Sure, done.

flexibeast:
PR updated: the script has been reworked, and i've added changes to the style guide (including removing what i feel is unnecessary text about obtaining things via `go get` and `cargo`).

@ericonr: Do you think the style guide needs to be trimmed further? i think it's fine as-is - i think it's good to let people know about the existence of `vmdfmt` and `mdbook-linkcheck`, and how to use them - but let me know if you think there's more that could be done. :-)

check.sh
```diff
@@ -0,0 +1,66 @@
+#!/bin/sh
+
+echo -n "Checking for mdbook-linkcheck ... "
+if ( ! command -v mdbook-linkcheck ); then
```
sgn, line 4: I'm late but why do we need to put this check (and others) in a subshell?

check.sh
```diff
@@ -0,0 +1,66 @@
+#!/bin/sh
+
```
sgn, line 2: We probably need `MISSING=` here,
or this could be a trouble:

```sh
MISSING=not-exist
export MISSING
./check.sh
```

ericonr: Good idea, thanks.

check.sh
ericonr: Not sure, actually. Is it too much of an issue?

sgn: I don't like `fork(2)` a new process for the subshell while we can check in the main shell just fine. Save a bit time, and typing :p

PR #394, opened 2020-07-09
repositories/index,guides/chroot: Multiple changes.

flexibeast:
* index: Mention locations of main repository for different system types.
* chroot: Refer to new "main repository" subsection in `index`.

Partially addresses #358.

### DIFF
--- a/src/xbps/repositories/index.md
+++ b/src/xbps/repositories/index.md
@@ -8,7 +8,16 @@ which may also be signed.
 Note that, while local repositories do not require signatures, remote
 repositories *must* be signed.
 
-## Official Repositories
+## The main repository
+
+The locations of the main repository in relation to a base [mirror
+URL](./mirrors/index.md) are:
+
+- glibc: `/current`
+- musl: `/current/musl`
+- aarch64 and aarch64-musl: `/current/aarch64`
+
+## Subrepositories
 
 In addition to the main repository, which is enabled upon installation, Void
 provides other official repositories maintained by the Void project, but not
--- a/src/installation/guides/chroot.md
+++ b/src/installation/guides/chroot.md
@@ -81,14 +81,10 @@ Follow only one of the two following subsections.
 
 ### The XBPS Method
 
-[Select an XBPS mirror](../../xbps/repositories/mirrors/index.md). For
-simplicity, save this URL to a shell variable for later use. Append `/current`
-to this repository URL to install a glibc distribution, or `/current/musl` for a
-musl distribution.
-
-Note: musl binaries are not available for i686.
-
-For example, this is the URL for glibc from a Tier 1 server:
+Select a [mirror](../../xbps/repositories/mirrors/index.md) and use the
+[appropriate URL](../../xbps/repositories/index.md#the-main-repository) for the
+type of system you wish to install. For simplicity, save this URL to a shell
+variable for later use, e.g.:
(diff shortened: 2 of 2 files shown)


src/xbps/repositories/index.md
```diff
@@ -8,7 +8,20 @@ which may also be signed.
...
+## The main repository
+
+The locations of the main repository in relation to a base [mirror
+URL](./mirrors/index.md) are:
+
+| System type | Location           |
+|-------------|--------------------|
+| glibc       | `/current`         |
+| musl        | `/current/musl`    |
+| aarch64     | `/current/aarch64` |
```
ericonr, line 20: Not sure these should go in the same table... Or perhaps have `aarch64` and `aarch64-musl`  in the last line?

src/xbps/repositories/index.md
```diff
@@ -8,7 +8,20 @@ which may also be signed.
...
+
+Note: musl binaries are not available for i686.
```
ericonr, line 22: Do we need to have this info?

flexibeast: This came from the existing "Installation via chroot" page; if you think it can be removed, i'm happy to do so. :-)

src/xbps/repositories/index.md
flexibeast: Fair point. So two tables, one of which is for `aarch64` specifically, and which lists both `/current/aarch64` and `/current/aarch64/musl`?

ericonr: I don't think so, because i's `/current/aarch64` for both. Perhaps `aarch64 and aarch64-musl | /current/aarch64` ?

src/xbps/repositories/index.md
ericonr: I'm not entirely sure. It is documented in the musl page already.

flexibeast: Okay, removing.

src/xbps/repositories/index.md
flexibeast: i've changed the text as you say, and presented the information as a list instead of a table - WDYT?

src/xbps/repositories/index.md
```diff
@@ -8,7 +8,16 @@ which may also be signed.
...
+## The main repository
+
+The locations of the main repository in relation to a base [mirror
+URL](./mirrors/index.md) are:
+
+- glibc: `/current`
+- musl: `/current/musl`
+- aarch64 and aarch64-musl: `/current/aarch64`
+
+## Subrepositories
```
ericonr, line 20: Is subrepos the correct name for them?

src/xbps/repositories/index.md
ericonr: Looks good to me, thanks!

src/xbps/repositories/index.md
flexibeast: That term is used on the current usage/xbps page on the Web site, so i assume so. :-)

ericonr: Fair enough :p

ericonr:
You may ~~fire~~ merge when ready.

flexibeast:
Heh, thanks. :-) CI failing due only to issues at www.libressl.org, so i'm merging this.

PR #396, opened 2020-07-11
icons: Add section.

flexibeast:
This PR addresses an issue frequently encountered by new Void users (including myself!).

### DIFF
--- a/src/config/graphical-session/icons.md
+++ b/src/config/graphical-session/icons.md
@@ -1,0 +1,18 @@
+# Icons
+
+## GTK
+
+By default, GTK-based applications try to use the Adwaita icon theme for
+application icons. Consequently, installation of the `gtk+3` package will also
+install the `adwaita-icon-theme` package. If you wish to use a different theme,
+install the relevant package, then specify the theme in
+`/etc/gtk-3.0/settings.ini` or `~/.config/gtk-3.0/settings.ini`.
+`adwaita-icon-theme` can be removed after
+[ignoring](../../xbps/advanced-usage.md#ignoring-packages) the package.
+
+For information about how to specify a different GTK icon theme in
+`settings.ini`, refer to [the GtkSettings
+documentation](https://developer.gnome.org/gtk3/stable/GtkSettings.html#GtkSettings.properties),
+in particular the
+"[gtk-icon-theme-name](https://developer.gnome.org/gtk3/stable/GtkSettings.html#GtkSettings--gtk-icon-theme-name)"
+property.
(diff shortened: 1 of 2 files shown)


src/config/graphical-session/icons.md
```diff
@@ -0,0 +1,18 @@
...
+
+## GTK
+
+In the absence of any other configuration (i.e. via `/etc/gtk-3.0/settings.ini`
+or `~/.config/gtk-3.0/settings.ini`), GTK-based applications will try to use the
+Adwaita icon theme for application icons. However, as users might prefer to use
+a different theme, the `adwaita-icon-theme` package is not installed by default,
+resulting in GTK applications not displaying the appropriate icons. If you have
+no preference for a different icon theme, install the `adwaita-icon-theme`
+package to ensure application icons are displayed.
```
ericonr, line 11: This feels like something that should be solved in packaging. I would like to raise it again with @void-linux/pkg-committers

flexibeast: Yes, my preference would be for the GTK packages (certainly `gtk+3`, maybe `gtk+`?) to pull in `adwaita-icon-theme`.

ahesford: We'll see if anybody objects to `gtk+3` depending on `adwaita-icon-theme` in [void-packages #23530](https://github.com/void-linux/void-packages/pull/23530). I think I was bitten by this when setting up my first Void system.

If everybody agrees and the dependency is added, the docs should be updated to note that the theme can be removed after an `ignorepkg` provided users are willing to manage their own themes.

flexibeast: That seems like a good approach, thanks.

ericonr:
I believe the main part is now solved. @flexibeast would you still like to include anything else?

Thanks @ahesford !

flexibeast:
@ericonr: Yes, thanks to @ahesford for solving this long-standing issue, and thanks to you, @fosslinux and @Chocimier for helping to check the solution. :-)

PR updated - i think it's ready to merge. Ok?

ericonr:
I'm not fully comfortable with dedicating a whole page to this, but I can't think of a better place to put it either...

flexibeast:
@ericonr: Fair enough, but then, i certainly don't think it's any worse than the presence of the "Shells" section. :-) i'd like to merge this, as we can always move the section contents elsewhere if a clear opportunity arises.

(i'd prefer to remove "Shells", not least for containing a "packages in the repos" list. We could instead briefly mention of `chsh(1)` somewhere else - maybe `config/index`, which at this point contains only a single, not-particularly-helpful, sentence.)

ericonr:
I'm in favor of removing "Shells" as well. I think mentioning `chsh` can go in user account stuff instead?

Feel free to merge, then. Thanks!

ericonr:
Fair enough.

Issue #392, opened 2020-07-08
documentation Request: documentation on setting up a void packages mirror


Frick-David:
Is there any chance of documentation on how to set up a mirror for helping out with getting people packages via xbps? I was thinking of setting up a virtual server to use with xbps for people in my area.

flexibeast:
My first thought is "no", given that how to mirror something isn't Void-specific information, and there are already existing guides to the process available elsewhere (e.g. [here](https://www.eff.org/keeping-your-site-alive/mirroring-your-site)). @ericonr, thoughts?

ericonr:
I mostly agree. I think we could link to one such guide (EFF <3), as an example. Not sure if it's easy to fit into the current structure.

flexibeast:
Could we perhaps add a subsection to `mirrors/index` saying something like (very roughly):

> If you'd like to set up a mirror, follow one of the many guides available (e.g. \<EFF guide\>), then submit a PR to the `void-docs` repo which adds your mirror to this page.

?

ericonr:
I like that option.

Frick-David:
Nice!

PR #401, opened 2020-07-14
mirrors/index: Add "Creating a mirror" subsection.

flexibeast:
Addresses #392.

### DIFF
--- a/src/xbps/repositories/mirrors/index.md
+++ b/src/xbps/repositories/mirrors/index.md
@@ -24,12 +24,13 @@ not managed by Void and do not have any guarantees of freshness or completeness
 of packages, nor are they required to sync every available architecture or
 sub-repository.
 
+### Globally-available mirrors
+
 | Repository                                         | Location          |
 |----------------------------------------------------|-------------------|
 | <https://mirror.ps.kz/voidlinux/>                  | Asia: Almaty, KZ  |
 | <https://mirrors.bfsu.edu.cn/voidlinux/>           | Asia: China       |
 | <https://mirrors.cnnic.cn/voidlinux/>              | Asia: China       |
-| <https://mirrors.hushan.tech:44300/voidlinux>      | Asia: China       |
 | <https://mirrors.tuna.tsinghua.edu.cn/voidlinux/>  | Asia: China       |
 | <https://mirror.maakpain.kro.kr/void/>             | Asia: Seoul, SK   |
 | <https://void.webconverger.org/>                   | Asia: Singapore   |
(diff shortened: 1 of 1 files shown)


src/xbps/repositories/mirrors/index.md
```diff
@@ -53,3 +53,12 @@ sub-repository.
...
 Mirrors](./tor.md) for more information.
+
+## Creating a mirror
+
+If you'd like to set up a mirror, follow one of the many guides available (e.g.
+[this EFF
+guide](https://www.eff.org/keeping-your-site-alive/mirroring-your-site)), then
+submit a pull request to [the void-docs
+repository](https://github.com/void-linux/void-docs) to add your mirror to this
+page. A mirror requires at least 1.1TB of storage space.
```
ericonr, line 64: "a full mirror requires around 1TB of storage", to allow for increases. Does adding "full" there help with making it clear that the mirror can have less than all the repositories?

flexibeast: Oh, well, i thought that "at least" covered the possibility of increases. :-) But i'll change it. And yes, adding "full" is good, i think. PR updated.

ericonr:
There's also this answer from @the-maldridge in one of the issues about mirrors:

> Running a mirror is pretty straightforward. All you have to do is setup rsync and keep it reasonably up to date. If your mirror is going to be exposed outside of your region, this should be noted in the documentation. If your mirror is internal to your region, i.e. behind some kind of carrier NAT, then this should also be noted.
> Please keep in mind that we pay bandwidth for all data sent out from the T1 mirrors. You can respect this by only mirroring if your use case for your mirror will offset the network throughput consumed by your mirror syncing.

I think we should try to include some of it.

flexibeast:
Thanks - i've incorporated that information and updated the PR. One thing to note is that the tables on this page currently don't include information about whether or not a mirror is only internal to a region.

src/xbps/repositories/mirrors/index.md
```diff
@@ -53,3 +53,21 @@ sub-repository.
...
+
+If you'd like to set up a mirror, and are confident you can keep it reasonably
+up-to-date, follow one of the many guides available for mirroring with
+[rsync(1)](https://man.voidlinux.org/rsync.1), then submit a pull request to
+[the void-docs repository](https://github.com/void-linux/void-docs) to add your
+mirror to this page.
+
+If your mirror is going to be exposed outside of your region, this should be
+noted in the documentation. If your mirror is internal to your region, i.e.
+behind some kind of carrier NAT, this should also be noted.
```
ericonr, line 67: @the-maldridge would you be fine with changing this to apply only for mirrors that are internal to their region? Most of the mirrors seem to be available worldwide (or at least from Brazil), so not being available everywhere would be the exception.

src/xbps/repositories/mirrors/index.md
```diff
@@ -53,3 +53,21 @@ sub-repository.
...
+
+A full mirror requires around 1TB of storage.
+
+Please keep in mind that we pay bandwidth for all data sent out from the Tier 1
+mirrors. You can respect this by only mirroring if your use case for your mirror
+will offset the network throughput consumed by your mirror syncing.
```
ericonr, line 73: "It is also possible to mirror only part of the repositories. Excluding debug packages is one way of decreasing the load on the Tier 1 mirros that has low impact on users."

ericonr:
> One thing to note is that the tables on this page currently don't include information about whether or not a mirror is only internal to a region.

IIRC, the only mirror I couldn't access was `hushan.tech`.

src/xbps/repositories/mirrors/index.md
flexibeast: Added to the previous para, where it seems to fit best, with some small changes.

ericonr:
>     maldridge │ then there should be a new section of region-locked mirrors

So a separate table for region locked Tier 2 mirrors.

flexibeast:
@ericonr: So create a new table with the hushan.tech mirror as the only entry? Or should we first confirm that it is indeed region-locked, rather than not being available for some other reason? If the latter, i'd prefer to merge this PR first, and open a separate issue about the hushan.tech mirror.

the-maldridge:
@flexibeast new table; it is region locked.

flexibeast:
@the-maldridge: Thanks, PR updated.

Is there anything else, or is this now ok to merge?

src/xbps/repositories/mirrors/index.md
ericonr: I think we need to update this paragraph then, right?

flexibeast: Gah, yes, thanks. i've removed this para entirely, and added the relevant text to the preceding one. WDYT?

ericonr: Looks good to me, thanks!

PR #385, opened 2020-07-05
raspberry-pi: Add section.

flexibeast:
This is mostly a simple port of [the old wiki page](https://wiki.voidlinux.org/Raspberry_Pi), with some light editing to fit Handbook style. i've never used an RPi, and don't have access to one, so i can't make any technical comments; but i'm happy to change the PR to reflect technical comments from those who have and/or do. :-)

Issue #133, opened 2019-11-07
Add documentation for arm/aarch64


flexibeast:
Cf. @Johnnynator's comment on /r/voidlinux [here](https://www.reddit.com/r/voidlinux/comments/dsr2bz/installing_on_aarch64/f6rpyqe/).

valadaa48:
fresh void aarch64 install failed to exit runit/2 stage due to having console=/dev/ttyFIQ0 in kernel command line. Once removed it worked.

https://old.reddit.com/r/voidlinux/comments/dwcste/aarch64_on_rock64_hanging_on_boot/

ericonr:
From IRC, agrisis was having trouble with the `udevd` service not working, as well as issues in dhcpcd and others. I believe they removed that line to get it to work.

```
      agrisis │ omg I figured it out
      agrisis │ kernel command line: console=/dev/ttyS0
      agrisis │ that messes up runit I guess
      agrisis │ https://old.reddit.com/r/voidlinux/comments/dwcste/aarch64_on_rock64_hanging_on_boot/
```

Fluffkin:
Couple of ideas for things to add:

The current default /boot/cmdline.txt setting for elevator is "noop". Changing that to "deadline" may improve apparent performance.

In /boot/config.txt if you set gpu_mem to anything less than 32 (MB), you will get alarming looking VCHI related errors in your boot log. They are apparently harmless, but do mean you are using a cut down version of the firmware.

If you've plugged in an LCD monitor to your HDMI on the Pi, you may get weird overscan / resolution options. You'll need to set hdmi_group to 2 in /boot/config.sys and lookup the correct values for hdmi_mode / your monitor here from the DMT group: https://www.raspberrypi.org/documentation/configuration/config-txt/video.md

Installing the package "fake-hwclock" is very useful for maintaining a roughly correct system time between reboots.

amak79:
Another idea for RPi.

The default CPU governor for the RPi kernel is `powersave`. This can make the system sluggish under certain loads as it will force the lowest frequency. A better default is `ondemand` which can be set for all cores with:
```
echo ondemand > /sys/devices/system/cpu/cpufreq/policy0/scaling_governor
```
I put the above in `/etc/rc.local`

PR #402, opened 2020-07-15
Create ARM devices page

ericonr:
Still work in progress, and I'm not sure I love the current structure. The `preparing your storage` header seems to be in a weird place. I wanted to add all the boards in this page as well, but it might be best to have a separate page to document their quirks.

Supersedes #385 , but I still haven't applied all comments from it.

### DIFF
--- a/src/installation/guides/arm-devices/index.md
+++ b/src/installation/guides/arm-devices/index.md
@@ -1,0 +1,170 @@
+# ARM Devices
+
+Void Linux provides packages and images for several ARM devices. Installing Void
+on such devices can be done in several ways:
+
+- [Pre-built images](#pre-built-images): images that can be flashed directly
+   onto an SD card or other storage medium, but give you a limited partition
+   layout and require manual expansion if you wish to increase the partitions;
+- [Tarball installation](#tarball-installation): PLATFORMFS and ROOTFS tarballs
+   that can be extracted to a previously prepared partition scheme;
+- [Chroot installation](#chroot-installation): follows most of the steps
+   outlined in [the chroot guide](../chroot.md).
+
+Since most of the commands on this guide will be run on external storage, it is
+important to run [sync(1)](https://man.voidlinux.org/sync.1) before removing the
+device.
+
+Further [configuration steps](#configuration) that are mostly specific to such
+devices are also outlined.
+
+## Installation
+
+If you are installing Void Linux on one of the ARM devices covered in the
+"[Supported platforms](./platforms.md)" page, make sure to read its section
+thoroughly.
+
+### Pre-built images
+
+After [downloading and verifying](../../index.md#downloading-installation-media)
+an image, it can be flashed onto an SD card in `/dev/mmcblk0`, for example, with
+the [cat(1)](https://man.voidlinux.org/cat.1),
+[pv(1)](https://man.voidlinux.org/pv.1) or
+[dd(1)](https://man.voidlinux.org/dd.1) command:
+
+```
+# dd if=<image>.img of=/dev/mmcblk0 bs=4M status=progress
+```
+
+### Custom partition layout
+
+It is possible to customize an installation further, for example with a custom
+partition layout, but it requires a more involved process. Two available options
+are:
+
+- [Tarball installation](#tarball-installation); and
+- [Chroot installation](#chroot-installation)
+
+To prepare the storage for these installation methods, it is necessary to
+partition the storage medium to then mount the partitions in the correct
+mounting points.
+
+The usual partitioning scheme for ARM devices requires at least two partitions
+on a drive formatted with an MSDOS partition table:
+
+- one formatted as FAT32 with partition type `0c`, which will be mounted on
+   `/boot`;
+- one that can be formatted as any file system that Linux can boot from, such as
+   ext4, which will be mounted on `/`. If you are using an SD card, for example,
+   it might be interesting to create the ext4 file system with the
+   `^has_journal` option, which disables journaling and might increase the
+   drive's life, at the cost of a higher chance of data loss.
+
+This can be done with the [cfdisk(8)](https://man.voidlinux.org/cfdisk.8)
+utility, for example.
+
+To access the newly created file systems, it is necessary to mount them. This
+guide will assume that the second partition will be mounted on `/mnt`, but you
+may mount it elsewhere. To mount these filesystems, you can use the commands
+below, replacing the device names with the appropriate ones for your setup.
+
+```
+# mnt /dev/mmcblk0p2 /mnt
+# mkdir /mnt/boot
+# mount /dev/mmcblk0p1 /mnt/boot
+```
+
+#### Tarball installation
+
+First, [download and verify](../../index.md#downloading-installation-media) a
+PLATFORMFS or ROOTFS tarball for your desired platform and [prepare your storage
+medium](#preparing-your-storage). Then, it is necessary to unpack the tarball
+onto the file system, which can be done using
+[tar(1)](https://man.voidlinux.org/tar.1):
+
+```
+# tar xvfp <image>.tar.xz -C /mnt
+```
+
+#### Chroot installation
+
+It is also possible to perform a chroot installation, which requires using the
+`qemu-user-static` package together with either the `binfmt-support` or `proot`
+package. This guide explains how to use the `qemu-<platform>-static` program
+from `qemu-user-static` with [proot(1)](https://man.voidlinux.org/proot.1).
+
+First, [download and verify](../../index.md#downloading-installation-media) a
+PLATFORMFS or ROOTFS tarball for your desired platform and [prepare your storage
+medium](#preparing-your-storage). Then, follow the [XBPS chroot installation
+steps](../chroot.md#the-xbps-method) using the appropriate architecture and base
+packages, some of which are listed in the "[Supported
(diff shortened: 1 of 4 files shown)


flexibeast:
Just noting, for cross-referencing purposes, that this PR address #133.

ericonr:
Ok, I think this is closer to a merge-able state. The additions specific to each board should go in `platform.md`, since `index.md` is just shy of our soft line limit :D 

I would like to make these additions in separate PRs, however.

If possible, review for everything:

- clarity
- correctness
- relevance
- anything else you can think of

Some notes, regarding differences from #385 and the wiki page:

- I removed mentions of configuring `module-load` for the random device in RPis, since it seems to be a built-in module in our kernels
- I removed mentions of the user groups for Xorg, and instead linked to the graphical session page, since RPis can also run Wayland compositors. If these groups are necessary, they should be mentioned in the Xorg page.

Pinging @flexibeast @Fluffkin @jnbr @Duncaen @Piraty @fosslinux @pbui

ericonr:
For the specific device configuration, linking to the official rpi foundation docs might be the best course of action. I will consider removing those.

src/installation/guides/arm-devices/index.md
```diff
@@ -0,0 +1,137 @@
...
+
+### System time
+
+Several of the ARM devices supported by Void Linux don't have battery powered
+real time clocks (RTC), which means they won't keep track of time once powered
+off. This issue can present itself as HTTPS errors when browsing the web or
+using the package manager. It is possible to set the time manually using the
+[date(1)](https://man.voidlinux.org/date.1) utility. In order to fix this issue
+for subsequent boots, install and enable [an NTP
+server](../../../config/date-time.md#ntp). Furthermore, it is possible to
```
jnbr, line 132: Nitpick: "client" or "daemon" would be a better fit.

ericonr: Thanks!

src/installation/guides/arm-devices/index.md
```diff
@@ -0,0 +1,137 @@
...
+thoroughly.
+
+### Pre-built images
+
+After [downloading and verifying](../../index.md#downloading-installation-media)
+an image, it can be flashed onto an SD card in `/dev/mmcblk0`, for example, with
+the [dd(1)](https://man.voidlinux.org/dd.1) command:
+
+```
+# dd if=<image>.img of=/dev/mmcblk0 bs=4M status=progress
```
Piraty, line 34: 1. mention `sync`?
2. in fact you can just `cat` or `pv` the image to the blockdevice

src/installation/guides/arm-devices/index.md
```diff
@@ -0,0 +1,137 @@
...
+
+## Configuration
+
+Some additional configuration steps need to be followed to guarantee a working
+system. Configuring a [graphical
+session](../../../config/graphical-session/index.md) should work as normal.
+
+### Logging in
+
+The `root` password is `voidlinux`.
```
Piraty, line 112: This is only true for the prebuilt images as well as the tarballs. not for installation via chroot where you set it yourself

src/installation/guides/arm-devices/index.md
```diff
@@ -0,0 +1,137 @@
...
+
+### fstab
+
+The `/boot` partition must be added to `/etc/fstab`, with an entry similar to
```
Piraty, line 116: You can boot rpi wihtout having mounted the fat partition at /boot, so `must` is not the right word here.

src/installation/guides/arm-devices/index.md
```diff
@@ -0,0 +1,137 @@
...
+
+After [downloading and verifying](../../index.md#downloading-installation-media)
+an image, it can be flashed onto an SD card in `/dev/mmcblk0`, for example, with
+the [dd(1)](https://man.voidlinux.org/dd.1) command:
+
+```
+# dd if=<image>.img of=/dev/mmcblk0 bs=4M status=progress
+```
+
+### Preparing your storage
```
Piraty, line 37: This section is not on the same logical level as the other 3 (it is necessary information to the tarball/chroot way)

ericonr: Do you have suggestions for where to put it? I agree with you, but I couldn't think of a better placement.

src/installation/guides/arm-devices/index.md
ericonr: I have this in the beginning:

```
Since most of the commands on this guide will be run on external storage, it is
important to run [sync(1)](https://man.voidlinux.org/sync.1) before removing the
device.
```

I always do `dd` for images. For `cat` it's just `cat <image> > /dev/mmcblk0`, right? I can add that.

src/installation/guides/arm-devices/index.md
ericonr: Oh good point, thanks.

src/installation/guides/arm-devices/index.md
ericonr: I remember I pretty much broke my installation when I updated without having `/boot` mounted, because the kernel became incapable of loading any modules. But you're right that it isn't required to boot. Will have to think of a better wording.

src/installation/guides/arm-devices/index.md
ericonr: Added cat.

src/installation/guides/arm-devices/index.md
ericonr: Fixed, I think.

src/installation/guides/arm-devices/index.md
ericonr: Tried to add a bigger explanation about it.

src/installation/guides/arm-devices/index.md
ericonr: Tried a new placement, lmk what you think.

src/installation/guides/arm-devices/index.md
Piraty: of course it's not wise to update kernel etc when /boot is not mounted correctly :-)

src/installation/guides/arm-devices/index.md
Piraty: suggestion: `with the cat, pv or  [dd(1)](https://man.voidlinux.org/dd.1) command:`
For sure we don't want to list all possible code snippets.
I'm open to not mention `cat` at all too

src/installation/guides/arm-devices/index.md
Piraty: suggested section layout:
```
# ARM Devices
<intro>
## Installation
<intro>
### Using pre-built image
<dd command etc>
### Custom partition layout
<intro>
<Preparing your storage>
#### rootfs tarball
#### chroot
## Configuration
...
```

src/installation/guides/arm-devices/index.md
ericonr: Indeed! It was a case of a faulty configuration, and then I completely borked the system when updating. Just want to be sure people don't commit the same one.

src/installation/guides/arm-devices/index.md
ericonr: This layout is very close to what @flexibeast suggested, so I'm switching to it. Thanks to both :)

ericonr:
Need to find out what's up with `xf86-video-fpturbo`, and if it's needed.

PaperMountainStudio:
`chrony` user is missing in some of 201911* prebuilt ARM images [1]. Should it be mentioned? RPi3's musl PLATFORMFS has this problem too, but the glibc version is ok. It will be fixed in the next image build [1].

[1]: https://github.com/void-linux/void-mklive/issues/4

ericonr:
Interesting, thanks! Oh, that was probably why chrony crashed constantly on my first RPi install! Yeah, I guess it bears mentioning.

@flexibeast this seems to be what that weird paragraph in the wiki was about ^^

Piraty:
This PR sits for nearly a month now. It looks good to me. I propose to merge, anything left can be fixed after (so docs. finally show the ARM page)

ericonr:
Agree. I have added @PaperMountainStudio's suggestion now, so that's the only new part. If someone could take a look?

@flexibeast are you good with merging this one?

flexibeast:
@ericonr: In the context of @Piraty's comment, this reads okay to me for now, although there are definitely a number of places where i'd change the language to more idiomatic English (e.g. from "mount the partitions in the correct mounting points" to "mount the partitions at the correct mount points"). i can prepare a separate PR for this once this is merged.

Additionally, there have been some recent comments ([here](https://github.com/void-linux/void-docs/issues/133#issuecomment-648431260) and [here](https://github.com/void-linux/void-docs/issues/133#issuecomment-669976804)) in https://github.com/void-linux/void-docs/issues/133 about possible content to include after this has been merged.

ericonr:
> i can prepare a separate PR for this once this is merged.

Ok, thanks! If you'd like to leave some comments, I can try to fix them in this PR. Otherwise, feel free to merge :D

flexibeast:
@ericonr: Mm, think i'll just merge this, as my suggested phrasing changes might inadvertently change the meaning from what was intended. So i'd prefer to do that in a separate PR. After *that's* merged, i'd like to announce these new sections on /r/voidlinux, as i feel they're a particularly significant addition. :-)

PR #404, opened 2020-07-16
xbps/index: Mention `xbps-query -f`.

flexibeast:
Partially addresses #337.

### DIFF
--- a/src/xbps/index.md
+++ b/src/xbps/index.md
@@ -74,6 +74,14 @@ kernels](../config/kernel.md#removing-old-kernels)" for further information.
 
 ## Finding Files and Packages
 
+If you can't find a file or program you expected to find after installing a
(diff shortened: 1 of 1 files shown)


src/xbps/index.md
```diff
@@ -74,6 +74,13 @@ kernels](../config/kernel.md#removing-old-kernels)" for further information.
 
 ## Finding Files and Packages
 
+To list all the files provided by an installed package, use
+[xbps-query(1)](https://man.voidlinux.org/xbps-query.1):
```
ericonr, line 78: Perhaps add something like "If you can't find a file or program you expected to find when installing a package, you can list all the files provided ..." ?

flexibeast: Good idea - PR updated.

PR #405, opened 2020-07-17
shell: Remove section, move relevant content to `config/index`.

flexibeast:


ericonr:
I see this page was brought over when porting one of the wiki pages into the docs by @Duncaen. Are you ok with the removal?

src/config/index.md
```diff
@@ -1,3 +1,13 @@
 # Configuration
```
ericonr, line 1: Perhaps make it clear that there are pages under this section? I think section contents would be too much, but at least make it explicit that the configuration section isn't only the shell part?

flexibeast: Yeah, i was thinking the same thing, but couldn't immediately come up with wording i didn't find awkward. (At one point i had: "This section and its subsections provide information about configuring your Void
system.")

ericonr: That doesn't sound too bad, and it's definitely better than nothing at all.

flexibeast: Oh okay - well, i've put it back in. :-)

Duncaen:
Would it make sense to move this to user configuration instead?

flexibeast:
That was suggested by @ericonr [here](https://github.com/void-linux/void-docs/pull/396#issuecomment-659426802), so i've updated the PR accordingly.

PR #412, opened 2020-07-22
[RFC] submitting: Mention suggesting changes/additions on r/voidlinux.

flexibeast:
i'd like to get feedback from @void-linux/doc-writers and @void-linux/pkg-committers about this. i'm personally happy to take contributions this way, to try to lower barriers to contributing, but i understand that other members of the Void team might not want Void to commit to such a policy.

### DIFF
--- a/src/contributing/void-docs/submitting.md
+++ b/src/contributing/void-docs/submitting.md
@@ -2,10 +2,13 @@
 
 Proposed changes should be submitted as pull requests to the
 [void-docs](https://github.com/void-linux/void-docs) repository on
-[GitHub](https://github.com/). Please note that, unlike a wiki, submissions will
-be reviewed before they are merged. If any changes are required they will need
-to be made before the pull request is accepted. This process is in place to
-ensure the quality and standards of the Handbook are sustained.
+[GitHub](https://github.com/). However, you can also suggest changes and/or
+additions on [the voidlinux subreddit](https://www.reddit.com/r/voidlinux/).
+
+Please note that, unlike a wiki, submissions will be reviewed before they are
+merged. If any changes are required they will need to be made before the pull
+request is accepted. This process is in place to ensure the quality and
+standards of the Handbook are sustained.
 
 ## Requirements
 


ericonr:
I feel like the mailing list, even when barely used, has a better contribution format than reddit. Contributing there would require quite a lot of work from the people bringing it to GH. Opening issues about stuff people commented on reddit works, tho.

flexibeast:
> I feel like the mailing list, even when barely used, has a better contribution format than reddit.

Agreed (although i'm biased, as someone who prefers mailing lists over Web fora). But in light of @the-maldridge [not wanting to link to the list](https://github.com/void-linux/void-docs/issues/337#issuecomment-662185268), i'm assuming we shouldn't be doing so in the Handbook.

ericonr:
It would be disjointed, but we could consider having mailing list only for void-docs. The issue is that it would depend mainly on you, so it isn't ideal.

flexibeast:
Yeah, a bus number of 1 is Not Good.

bobertlo:
I'm personally inclined to think that posting an issue on github is not too steep of a barrier. I am not saying we couldn't take suggestions on reddit, but I don't know how I feel about directing people from *this page* to post on reddit about the docs.

flexibeast:
@bobertlo: \*nod\* Fair enough. My experience has been that, for a number of people, the issue is not technical, but political: GH is owned by MS, GH has a nonfree UI, etc.

the-maldridge:
The *official* way to contribute to Void is here on GitHub.  As a project that values simplicity, some maintainers are willing to accept patches from the ML if they are of exceptionally high quality but this is not an official channel of the project.  While I recognize and value that some folks have principals that are incompatible with our current hosting, this is not sufficient reason to make it significantly harder to keep Void running in a sane way.

arp242:
> Fair enough. My experience has been that, for a number of people, the issue is not technical, but political: GH is owned by MS, GH has a nonfree UI, etc.

Doesn't Reddit have pretty much the same issues? (personally, I don't think it's much an issue though, but Reddit and GitHub both seem "big corp" and are "non-free".)

One big problem with Reddit is that the "new" UI is incompatible with old.reddit.com; specifically, triple-quoted code blocks only work with the new UI and not the old one.

PR #413, opened 2020-07-22
chroot.md: Update mount syntax for /dev, /proc, /sys

ahesford:
Apparently `mount` accepts `--rbind` and `--make-rslave` in the same
invocation, so we can merge the two commands per mount into one.

### DIFF
--- a/src/installation/guides/chroot.md
+++ b/src/installation/guides/chroot.md
@@ -137,9 +137,9 @@ methods.
 Mount the pseudo-filesystems needed for a chroot:
 
 ```
-# mount --rbind /sys /mnt/sys && mount --make-rslave /mnt/sys
-# mount --rbind /dev /mnt/dev && mount --make-rslave /mnt/dev
-# mount --rbind /proc /mnt/proc && mount --make-rslave /mnt/proc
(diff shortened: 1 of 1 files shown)


ericonr:
I could swear I read something about `--make-rslave` being run in a separate invocation for a reason.

ahesford:
Apparently this requires some further investigation.

ericonr:
Here https://unix.stackexchange.com/questions/120827/recursive-umount-after-rbind-mount

ahesford:
The [StackExchange answer](https://unix.stackexchange.com/a/371208) and subsequent comments are not consistent with my experience. The first comment says it's OK to combine, but the second comment suggests that combining the two will make it non-recursive. However, if I do
```
mount --make-rslave --rbind /dev /mnt
```
on my running system, I definitely have `/mnt`, `/mnt/pts` and `/mnt/shm`.

ericonr:
I have observed the same thing as you. Perhaps there are some edge cases?

ahesford:
Given the possibility that this does something unexpected, I don't see value in changing the docs to shave off a few keystrokes.

PR #414, opened 2020-07-22
iwd: Tidy section.

flexibeast:
* Remove now-unnecessary note about minimal kernel version.
* Convert note about wheel group membership to plain text.
* Improve phrasing.

Using `pandoc` to convert this section to a PDF results in an error due to the presence of '≥', but i assume that the note containing that character is no longer needed.


bobertlo:
LGTM

Issue #407, opened 2020-07-19
PDF version of Handbook


flexibeast:
Suggested by /u/YouNeverWalkAlone [here](https://www.reddit.com/r/voidlinux/comments/htcroz/void_documentation_progress_report/fyghp6w/).

i'm not sure what the best approach to this would be. Use `pandoc` to export to LaTeX, then use `pdflatex` to produce the PDF?

ericonr:
We'd have to check the results, but now that we have texlive packages that might be a viable solution and we could even include in the package. I believe @abenson has some templates for markdown -> LaTeX stuff we could try to use.

flexibeast:
#416 is now basically ready to be merged, modulo any feedback from @void-linux/pkg-committers. However, some things to follow up in the future:

* The ToC structure follows LaTeX conventions, rather than the Web site conventions. Some code could be added to create a ToC along the lines of the latter.
* Lines with filepaths are currently handled by using `sloppypar` for the entire document, which is a probably an overly-blunt tool.
* The Tor repo table is over-indented.

ericonr:
[handbook.pdf](https://github.com/void-linux/void-docs/files/4985723/handbook.pdf)

If anyone wants to get their hands on it without building

bobertlo:
I am not posting these in the PR because I don't view anything as blocking. I don't feel great about the paragraph indentation being non-uniform. The first paragraph of a section lacking indentation and then the following having it really sticks out to me. This is highlighted by the use of short single sentence paragraphs. Seeing it in this format makes this stick out more. The handbook feels more like a wiki and in that display format it seems natural.

Another thing is just that the paragraphs are single spaced from eachother while lists seem to be double-spaced between their elements and other paragraphs. I don't really know anything about latex, but that is the only other note I saw.

The short sentences and stub sections taking up whole pages I view as more a product of our writing than the PDF.

I do not think these things are blockers from publishing this, just more issues to consider in the future if we are going to do tagged releases. I am starting to have a lot more free time lately and will try to take a look at some of this stuff with you guys. I love this and keep up the good work!

bobertlo:
If anything, I think in light of this we could merge InfraDocs into the section above as like a see also... just so it doesn't sit alone on the page, or could be expanded to elaborate more on the infrastructure of void in general (and maybe indirectly ease common questions about a former public figure?)

[...]

PR #416, opened 2020-07-22
mdbook-latex: Add mdBook backend script.

flexibeast:
Addresses #407.

This script makes use of `pandoc` and `pdfunite`, the latter being provided by the `poppler` package. There's some temporary special-casing to deal with issues that are addressed in #414 and #415 

The resulting PDF has some layout problems (e.g. on pp. 9 and 10).

i imagine we might want to incorporate this script into `build.sh`, but i've kept it separate for now, whilst it's in development. (We could, of course, keep it separate, and simply call it from `build.sh` if/when required.)

### DIFF
--- a/mdbook-latex
+++ b/mdbook-latex
@@ -1,0 +1,396 @@
+#!/usr/bin/perl
+
+use strict;
+use warnings;
+
+use Cwd;
+use JSON;
+
+my $data;
+{
+  local $/;
+  undef $/;
+  $data = <>;
+}
+my $json = JSON->new->decode($data);
+my $sections = $json->{book}->{sections};
+
+my $date = `date '+%d %B %Y'`;
+chomp $date;
+
+my $latex;
+my %code_inline;
+my $code_inline_index = 0;
+my %codeblocks;
+my $codeblock_index = 0;
+
+# Opening.
+
+$latex .= "\\documentclass{article}\n";
+$latex .= "\\usepackage[margin=4cm]{geometry}\n";
+$latex .= "\\usepackage[T1]{fontenc}\n";
+$latex .= "\\usepackage{hyperref}\n";
+$latex .= "\\usepackage{listings}\n";
+$latex .= "\\lstdefinestyle{void}{\n";
+$latex .= "  basicstyle=\\ttfamily,\n";
+$latex .= "  breaklines=true,\n";
+$latex .= "  showstringspaces=false,\n";
+$latex .= "  aboveskip=\\bigskipamount,\n";
+$latex .= "  belowskip=\\bigskipamount,\n";
+$latex .= "}\n";
+$latex .= "\\def\\code#1{\\texttt{#1}}\n";
+$latex .= "\\begin{document}\n";
+$latex .= "\\title{Void Handbook}\n";
+$latex .= "\\date{$date}\n";
+$latex .= "\\maketitle\n";
+$latex .= "\\setcounter{tocdepth}{5}\n";
+$latex .= "\\tableofcontents\n";
+$latex .= "\\begin{sloppypar}\n"; # Ugly; need to find better solution
+
+# Contents.
+
+foreach my $i (@$sections) {
+    process_json($i);
+}
+
+# Postprocessing.
+
+$latex =~ s/config-network-wpa\\_supplicant/config-network-wpa_supplicant/g;
+$latex =~ s/config-network-wpa_supplicant-the-wpa\\_supplicant-service/config-network-wpa_supplicant-the-wpa_supplicant-service/g;
+$latex =~ s/(handbook-codeblock-\d+)/$codeblocks{$1}/eg;
+$latex =~ s/(handbook-codeinline-\d+)/$code_inline{$1}/eg;
+
+# Closing.
+
+$latex .= "\\end{sloppypar}\n";
+$latex .= "\\end{document}\n";
+
+open(my $fh, ">", "handbook.tex")
+    or die "Can't open handbook.tex for writing: $!";
+print $fh $latex;
+close($fh)
+    or die "Couldn't close handbook.tex after writing: $!";
+
+exit 0;
+
+sub process_json {
+
+    my $i = shift;
+    my $type = ref($i);
+    if ($type eq 'HASH') {
+        if (defined $i->{content}) {
+            process_content($i->{path},$i->{content});
+        }
+        if (defined $i->{Chapter}) {
+            process_json($i->{Chapter});
+        }
+        if (defined $i->{sub_items}) {
+            foreach my $j ($i->{sub_items}) {
+                process_json($j);
+            }
+        }
+    }
+    if ($type eq 'ARRAY') {
+        foreach my $j (@{$i}) {
+            process_json($j);
+        }
+    }
+}
+
+sub process_content {
+
+    my $path = shift;
+    my $content = shift;
+    my $base = convert_path_to_hypertarget_base($path);
+
+    # Create sections.
+    $content =~ s/^# ([^\n]+)/create_sectioning("section",$1,$base)/eg;
+
+    # Create subsections.
+    $content =~ s/\n## ([^\n]+)/create_sectioning("subsection",$1,$base)/eg;
+
+    # Create subsubsections.
+    $content =~ s/\n### ([^\n]+)/create_sectioning("subsubsection",$1,$base)/eg;
+
+    # Create paragraphs.
+    $content =~ s/\n#### ([^\n]+)/create_sectioning("paragraph",$1,$base)/eg;
+
+    # Create subparagraphs.
+    $content =~ s/\n##### ([^\n]+)/create_sectioning("subparagraph",$1,$base)/eg;
+
+    # Create codeblocks.
+    $content =~ s/```\n(.+?)\n```/create_codeblock($1)/esg;
+
+    # Create internal links.
+    $content =~ s/\[([^]]+?)\]\(((?:\..?.+?)|(?:#.+?))\)/create_internal_link($1,$2,$path)/esg;
+
+    # Create external links.
+    $content =~ s/<(http[^>]+)>/\\href{$1}{$1}/sg; # bare links
+    $content =~ s/\[([^]]+?)\]\((.+?)\)/\\href{$2}{$1}/sg;
+
+    # Create lists.
+    $content =~ s/\n\n(- .+?)\n(?:\n|\z)/create_list($1)/esg;
+
+    # Create numbered lists.
+    $content =~ s/\n\n(\d+\. .+?)\n(?:\n|\z)/create_numbered_list($1)/esg;
+
+    # Create blockquotes.
+    $content =~ s/((^>(?:\s*\n|[^\n]+))+)/create_blockquote($1)/emsg;
+
+    # Bold text.
+    $content =~ s/\*\*([^*]+)\*\*/{\\bfseries $1}/sg;
+
+    # Escape LaTeX special characters.
+    $content =~ s/#/'\#'/eg;
+    $content =~ s/\$/'\$'/eg;
+    $content =~ s/_/'\_'/eg;
+    $content =~ s/~/'\~'/eg;
+    $content =~ s/&/'\&'/eg;
+    $content =~ s/\^/'\^'/eg;
+
+    # Create tables.
+    $content =~ s/((?:(?:\| +[^|]+ +\| +(?:[^|]+ +\|)+\n)|(?:\|-[^\n]+\n))+)/create_table($1)/esg;
+
+    # Create inline code.
+    $content =~ s/(?: |\n)`([^`]+)`/create_code_inline($1)/esg;
+
+    $latex .= $content;
+
+}
+
+sub convert_heading_to_fragment {
+
+    my $heading = shift;
+
+    $heading =~ s/ /-/g;
+    $heading =~ s/\.//g;
+    $heading =~ s/\(|\)//g;
+    $heading =~ s/(.+)/lc($1)/e;
+
+    return $heading;
+
+}
+
+sub convert_path_to_hypertarget_base {
+
+    my $base = shift;
+
+    if ($base =~ m|index.md|) {
+        $base =~ m|(.+/index).md|;
+        $base = $1;
+        $base =~ s|/|-|g;
+    } else {
+        $base =~ m|(.+).md|;
+        $base = $1;
+        $base =~ s|/|-|g;
+    }
+
+    return $base;
+
+}
+
+sub create_blockquote {
+
+    my $blockquote = shift;
+    my $result;
+
+    $blockquote =~ s/>//g;
+    $blockquote =~ s/\n//g;
+
+    $result .= "\\begin{quote}\n";
+    $result .= $blockquote;
+    $result .= "\\end{quote}\n";
+
+    return $result;
+
+}
+
+sub create_codeblock {
+
+    my $codeblock = shift;
+    my $label;
+    my $result;
+
+    $codeblock_index++;
+    $label = 'handbook-codeblock-' . $codeblock_index;
+
+    $result .= "\\begin{lstlisting}[style=void]\n";
+    $result .= $codeblock;
+    $result .= "\n\\end{lstlisting}\n";
+
+    $codeblocks{$label} = $result;
+
+    # Mark places in document where codeblock will be re-inserted
+    # during postprocessing.
+    return $label;
+
+}
+
+sub create_code_inline {
+
+    my $code = shift;
+    my $label;
+    my $result;
+
+    $code_inline_index++;
+    $label = 'handbook-codeinline-' . $code_inline_index;
+
+    $result = " \\code{$code}";
+
+    $code_inline{$label} = $result;
+
+    # Mark places in document where code will be re-inserted
+    # during postprocessing.
+    return $label;
+
+}
+
+sub create_internal_link {
+
+    my $text = shift;
+    my $destination = shift;
+    my $current_section = shift;
+    my $current_directory;
+    my $base;
+    my $fragment;
+    my $hypertarget;
+    my $result;
+
+    if ($destination =~ /^#/) {
+        $base = $current_section . $destination;
+        $base =~ s|.md#.+$||;
+    } else {
+        $current_section =~ m|^(.+/)|;
+        $current_directory = $1;
+        $base = $destination;
+        $base =~ m|^(.+\.md)|;
+        $base = $1;
+        $base = $current_directory . $base;
+        $base = Cwd::abs_path("../../src/" . $base);
+        $base =~ s|^.+/void-docs/src/||;
+        $base =~ s|.md||;
+    }
+    $base =~ s|/|-|g;
+
+    if ($destination =~ /#/) {
+        $destination =~ m|#(.+)$|;
+        $fragment = "-" . $1;
+    } else {
+        $fragment = "";
+    }
+
+    $text =~ s/\n/ /;
+    $hypertarget = $base . $fragment;
+    $result = "\\hyperlink{$hypertarget}{$text}";
+
+    return $result;
+
+}
+
+sub create_list {
+
+    my $list = shift;
+    my $result;
+
+    $result .= "\n\\begin{itemize}\n";
+    $list =~ s/^- /\\item /mg;
+    $result .= $list;
+    $result .= "\n\\end{itemize}\n";
+
+    return $result;
+
+}
+
+sub create_numbered_list {
+
+    my $list = shift;
+    my $result;
+
+    $result .= "\n\\begin{enumerate}\n";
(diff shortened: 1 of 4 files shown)


flexibeast:
#414 and #415 merged; removed special-casing.

ericonr:
Can pandoc itself generate PDFs? I thought it needed LaTeX available somewhere on the system. That's awesome!

Since it's a bigger script, I don't see too much of an issue with running it separately. Will try it in a bit!

flexibeast:
Oh, sorry, no, i installed `texlive-20200406_2` to provide `pdflatex` for `pandoc`.

ericonr:
No problem! We should add these deps to the readme, then.

flexibeast:
PR updated with changes to README.

ericonr:
Hmm I wonder if we could somehow append the markdown documents together instead... Besides the formatting issues you pointed out, the page numbers are a bit weird, and we don't have clickable links :/

flexibeast:
@ericonr: Sure, i'll rework it to concatenate the .md files first; i imagine that would mean that `pdfunite` would no longer be required. This might also fix the page numbering issue.

Yeah, i'm not sure why there aren't internal links; maybe they need to be actively enabled via `hyperrefoptions`?

ericonr:
I'm afraid it might make weird things to the headers by concatenating everything. Let's see what happens!

Have you tested the alternative mdbook backend for ebooks? Maybe we could print a PDF from that? I know it looked weird when I tried it, but idk if it was caused by foliate or by the backend.

ericonr:
Btw, when you're done remember to add the generated files to `.gitignore` :)

flexibeast:
No, the headers seem fine when concatenating the Markdown sources (as does the page numbering). The problem is that new sections don't appear on new pages, and inserting FF or FS characters in the Markdown don't work. i'm about to try converting to LaTeX, inserting LaTeX page breaks where appropriate, then running `pandoc`. i'd prefer not to  use the EPUB backend unless we really need to, since `pandoc` is still going to need `texlive` installed to convert the EPUB to PDF, so i'd like to avoid a possibly-unnecessary dependency.

ericonr:
The reason I would consider the ebook backend is exactly that it would have already done the processing to split the pages where necessary as well as make the links work. But if you manage to make it work without that, all the better!

flexibeast:
Trying to put together a LaTeX file suitable for direct use by `pdflatex`, involving adding `\documentclass{book}`, various `\usepackage`s and the manual definition of a command defined by `pandoc` (`tightlist`), isn't yet producing anything satisfactory. (Not only is the layout utterly broken in terms of horizontal space, but the hyperlinks are looking for individual `.md` files.) But trying to use `mdbook-epub`, i get:

```
Error: Incompatible mdbook version, expected 0.3.7 but got 0.4.0
```
:-(

flexibeast:
(If the EPUB stuff doesn't work out either, i can work on a Perl mdBook backend to generate the necessary LaTeX from the JSON representation - i've actually made an exploratory start on that.)

ericonr:
Hmm I think `mdbook-epub` hasn't been updated in a while :/

https://github.com/rust-lang/mdBook/issues/88 has some stuff, including software to collate the markdown files!

ericonr:
https://github.com/cetra3/mdcollate this one seems quite interesting!

flexibeast:
Actually, my Perl backend is coming along nicely: here's [an excerpt of the output it's producing](https://pastebin.com/4rBRvMvg). i'm in the middle of working out how to do internal links; once that's done, i can work on producing LaTeX lists and tables.

If you have no objections, i'd like to focus on continuing to develop this, as it seems promising: this backend could give us total control over the output, with the ability to do things like create a nice ToC, use the `listings` package for `sh` codeblocks, create an index, etc. WDYT?

ericonr:
I don't have any issues, and having full control is indeed very interesting! Good luck c:

flexibeast:
Okay, thanks. i've updated the PR to include the `mdbook-latex` backend script, rather than `pdf.sh`, and changed the PR title accordingly.

ericonr:
Oh, you're doing it the way we had discussed it a while ago, right? Using the JSON output of mdbook! Nice.

mdbook-latex
```diff
@@ -0,0 +1,120 @@
...
+    process_json($i);
+}
+
+$latex .= "\\end{document}\n";
+
+open(my $fh, ">", "handbook.tex")
+    or die "Can't open handbook.tex for writing: $!";
+print $fh $latex;
+close($fh)
+    or die "Couldn't close handboox.tex after writing: $!";
```
ericonr, line 41: handbook

flexibeast: Thanks, already fixed locally. :-)

flexibeast:
PR updated. Running `pdflatex` on the `handbook.tex` file in `book/latex/` completes, once one removes the backslash from:

```
grep -m1 '(II) modeset([0-9]\+):' /var/log/Xorg.0.log
```

in `xorg.md` (which i believe shouldn't be there anyway).

There are very obviously many visual issues still to address, and although external links are (i believe) working, internal links are not, though the infrastructure for that is now basically in place.

ericonr:
Would you like to make this into a separate packaged project or should it be kept in-tree? I can see the case for making it a package with all the dependencies and then we could talk to @the-maldridge about deploying the pdf in the mirror as well, for people to download.

ericonr:
It would also help with PATH issues, since it's complaining about not finding the latex backend.

flexibeast:
i was wondering about this also. My thinking has been that it should perhaps be kept in void-docs, as the code is (at least at this point) quite specific to the Void Handbook, and how we do things. Otoh, the arguments for a distinct repo under void-linux make sense as well.

Re. the PATH issue, yes, i've just been adding the path to my local void-docs repo to PATH. Obviously if it's a separate package, this is not an issue, but if we keep it in-tree, we'll need to work out how to best address this issue.

ericonr:
Checklist time:

- [ ] enable code block wrap
- [ ] find out why the margins change every other page (might be because it's configured for a notebook style thing?)
- [ ] ideally fix tables as well
- [ ] code blocks have a weird backslash: `\# f d i s k − l Disk / dev / sda`
- [ ] local links

This is pretty cool already, though! Thanks :D

ericonr:
> Re. the PATH issue, yes, i've just been adding the path to my local void-docs repo to PATH. Obviously if it's a separate package, this is not an issue, but if we keep it in-tree, we'll need to work out how to best address this issue.

Yeah. I needed to add the full path to PATH for it to work.

> Otoh, the arguments for a distinct repo under void-linux make sense as well.

I don't see an issue with starting out like this. I guess we can keep it in tree while we work on it, and then when we decide to ship PDFs we move to a separate project. Can the current repo license apply to code? Might make sense to add a license header on the top too.

flexibeast:
> enable code block wrap
> find out why the margins change every other page (might be because it's configured for a notebook style thing?)

Yeah, it's probably due to `documentclass` being `book`.

> ideally fix tables as well

Definitely - haven't implemented table conversion yet. :-)

> code blocks have a weird backslash: \# f d i s k − l Disk / dev / sda

Yep, this is due to `$` being a special character for LaTeX; i'm just escaping them in general right now, and will need to special-case handling `$` (and `#`) inside listings.

> This is pretty cool already, though!

Thanks!

flexibeast:
@ericonr: PR updated.

* Tables basically working (although the table for flavors will need to be rotated and presented on a landscape page).
* Blockquotes basically working (except for the one in "About this Handbook", which we could maybe change to a simple codeblock).
* Escaping of `#` and `$` in codeblocks fixed.

The next thing i'll be focusing on is getting internal links working.

ericonr:
You can add an `optional =true` key to the backend config, which means it won't fail with an error if it can't find the `mdbook-latex` backend.

flexibeast:
@ericonr: Thanks, i've now changed `book.toml` locally.

flexibeast:
@ericonr: PR updated:

* Internal links are now working in general, though (as the output of `pdflatex` shows) there are still a number of links not properly handled.
* `documentclass` has been changed from `book` to `article`.

ericonr:
Nice work! I much prefer the layout with `article`.

Most of the links seem to be working fine, too :D 

What I found: the first page has a "Content" header but nothing beneath it, and the commands in pafe 23 (FDE guide) have some weird formatting around them.

flexibeast:
> the first page has a "Content" header but nothing beneath it

That's odd. i'm getting a full ToC here. A couple of thoughts:

* LaTeX can require that it be run twice, to correctly generate references. Does the ToC appear after doing that?
* If "no", can you try removing all the generated `handbook.*` files, then re-running `pdflatex` twice, and checking if a ToC is then produced?

> the commands in pafe 23 (FDE guide) have some weird formatting around them.

Could you be more specific? i'm not noticing any particular issues with surrounding formatting (though there are definitely page overflows).

ericonr:
> LaTeX can require that it be run twice, to correctly generate references. Does the ToC appear after doing that?

Indeed, my bad!

> Could you be more specific? i'm not noticing any particular issues with surrounding formatting (though there are definitely page overflows).

I might have been mistaken regarding this one as well :p 

Perhaps at some point we could look into making the content separation between code blocks and prose a bit clearer.

flexibeast:
@ericonr: PR updated. Now with special-casing to deal with the 'flavors' table!

* Most internal links should now work; the exceptions are those relating to `kernel.md` and `wpa_supplicant.md`. i've got a rough idea of what the problems are, but i'm not yet sure how to address them.
* The overall margins have been decreased slightly.
* The 'flavors' table is special-cased to (amongst other things) reduce the margins even further, to allow it to fit on a portrait page. i think this is probably okay for now; i'm not sure whether we want to make that table landscape somehow (which might be slightly inconvenient for readers).
* Most margin overflows are now fixed, via a less-than-ideal solution. One exception is the table containing the Swedish Tor mirror, but again, i think this is okay for now.
* The spacing before and after code listings has been increased.

i'm starting to think more about how to deploy this. Things we need to think about:

* the best way to add `mdbook-latex` to PATH.
* the best way of building the PDF in `build.sh`.
* how to deploy the built PDF, and where (perhaps `https://docs.voidlinux.org/handbook.pdf`?)

Once we've sorted out the above, the landing page can be modified to provide a link to the PDF.

ericonr:
> the best way to add mdbook-latex to PATH.

If done inside `build.sh`, it's reasonably simple. But for deployment I would like to consider the possibility of having a package. Perhaps @void-linux/pkg-committers can weigh in too?

> the best way of building the PDF in build.sh.

I know `cd book/latex` and then building it twice should work. What other way would you suggest?

ericonr:
https://docs.voidlinux.org/handbook.pdf looks good :D

flexibeast:
> What other way would you suggest?

Oh, i didn't have anything else in particular in mind .... It's just that my brain is quite frazzled at the moment, so i'm not sure that i'm not missing anything important. :-)

flexibeast:
@ericonr: PR updated. i've now special-cased the style guide, because it contains many examples that shouldn't be converted. Instead of trying to write hairy code to deal with that, i've manually created a `style-guide.tex` file that gets inserted at the appropriate point. This does mean that `style-guide.md` and `style-guide.tex` have to be kept in sync, but i feel that will generally involve less work than having to create (and potentially modify in the future) code to handle all the special cases.

The main thing left to do, that i'm aware of, is to fix the last several broken links.

flexibeast:
@ericonr: Oh, fixing the remaining broken internal links turned out to be easier than i thought. :-) All should now be working.

ericonr:
Wow, this is looking great!

I don't love the idea of the separate syle guide, but I understand the reasoning for it. I guess we should leave a comment on it that it should be kept in sync with the markdown version. We could also consider not including the style guide in the PDF version, since anyone who's contributing surely will have other ways to access the handbook than the PDF....

Do you know if we could make the code blocks have the same font as inline code?

flexibeast:
> Wow, this is looking great!

:-D

> We could also consider not including the style guide in the PDF version, since anyone who's contributing surely will have other ways to access the handbook than the PDF....

Fair enough; i'll leave it to others to make that decision.

> Do you know if we could make the code blocks have the same font as inline code?

Should be doable. i'll see what i can do.

ericonr:
Would excluding a page be simple, and would you be ok with not having it? You did go to the trouble of transposing it.

If we do omit it from the PDF, might be nice to add it as a config option in `book.toml` to make it configurable and not have to add special cases.

flexibeast:
> Would excluding a page be simple, and would you be ok with not having it?

Yes, and sure. :-)

> If we do omit it from the PDF, might be nice to add it as a config option in book.toml to make it configurable and not have to add special cases.

Sorry, i'm not sure i understand .... Could you elaborate?

ericonr:
> Sorry, i'm not sure i understand .... Could you elaborate?

afaik, the JSON output of mdbook also contains the configuration keys listed under `[output.latex]`, so perhaps we can have an `exclude = [ "style-guide", ... ]` key there, so that excluding the style guide doesn't require inserting special cases inside the script.

flexibeast:
@ericonr: PR updated with changed codeblock typeface style.

ericonr:
Nice!

I believe it's very usable now and I am almost ok with merging it (unless we decide on making it a separate project). However, I will note some nitpicks here:

- The Hashboot page is weird: the paragraph after `# hashboot index` is indented wrong; "usage" for Flashrom has a leading `#` for some reason
- The AppArmor items don't have newlines between them
- The table for the TOR repo is a bit too indented, I _think_

And non nitpicks that we should solve before shipping a PDF:

- There is an artifact in the `shlib errors` section header (page 117) - it probably got confused with some code block, and then it's showing what should be prose as code too.
- The block with an example of a repodata error should be a simple code block, but it has some weird quoting

flexibeast:
> so that excluding the style guide doesn't require inserting special cases inside the script.

Ah, okay. Yes, i agree that the script should have as little special-casing as possible; i'll check if adding an `exclude` entry works, and if so, update the PR accordingly.

Thanks for finding those nits! i'll look into them. (Though the Tor repo table should use the same table indentation as the other, non-flavors, tables ....)

Finally, re. `build.sh`, should i uncomment the currently-commented calls to `pdflatex`, and just add `PATH="$PWD:$PATH"` to the top of the file?

ericonr:
> Ah, okay. Yes, i agree that the script should have as little special-casing as possible; i'll check if adding an exclude entry works, and if so, update the PR accordingly.

Thanks!

> Finally, re. build.sh, should i uncomment the currently-commented calls to pdflatex, and just add PATH="$PWD:$PATH" to the top of the file?

For the PATH part, yes. For the pdf generation we should have it happen in a separate folder, I think. `pdflatex` generates lots of artifacts. You could do `cd book/latex`, perhaps?

flexibeast:
> For the PATH part, yes.

Done.

> For the pdf generation we should have it happen in a separate folder, I think. pdflatex generates lots of artifacts. You could do cd book/latex, perhaps?

Well, one can just pass the `-output-directory=` option to `pdflatex`.

PR updated according to the above.

flexibeast:
Hmm, i can't seem to get the `exclude` to work. Under `[output.latex]`, i tried both:

```
exclude = [ "style-guide.md" ]
```
and
```
exclude = [ "contributing/void-doc/style-guide.md" ]
```

but the style guide seems to be included nonetheless in both cases. Can you confirm?

flexibeast:
(Btw, the lack of numbering in the "AppArmor" section was because i'd not implemented handling numbered lists; the PR now includes such an implementation.)

It looks like at least some of the problems on the "Common Errors" page are also caused by the presence of particular quotes in the error messages. Can you please confirm for me that the two "ERROR" messages and the "DEBUG" message on that page accurately reflect XBPS output re. quotes?

flexibeast:
@ericonr: i believe i might have solved the codeblocks issue in general. :-D As a bonus, the style guide no longer requires special-casing overall (though i can spot at least one small issue). Could you please confirm? i've just pushed an update.

ericonr:
> Well, one can just pass the -output-directory= option to pdflatex.

That works, but I didn't know about it.

> (Maybe the backend is supposed to handle excludes?)

Yes, sorry. I should have made that clear. I guess it's no longer needed, though :D

ericonr:
> (though i can spot at least one small issue)

I'm guessing the upside down interrogation mark?

> i believe i might have solved the codeblocks issue in general

Indeed!

I've just noticed that the index is a bit off, e.g. the Wayland section should be under Graphical Session, not on the same level.

flexibeast:
> I'm guessing the upside down interrogation mark?

Yeah. i'll look into that.

> Indeed!

Excellent! i'll remove the `style-guide.tex` file from this PR, and create a PR to revert #430.

> I've just noticed that the index is a bit off, e.g. the Wayland section should be under Graphical Session, not on the same level.

i noticed something similar re. the "Hashboot" and "AppArmor" sections. i'll look into that too.

So, other things still left to do:

* The problem with the "If this is not run after installation ..." para in "Hashboot" seems to be partially addressed by #433, but not completely - i don't know what's going on there yet, the LaTeX looks reasonable. Not sure if you want to make this a blocker.
* The Tor mirrors table is definitely indented too far.

Anything else? Have i forgotten anything?

ericonr:
> The problem with the "If this is not run after installation ..." para in "Hashboot" seems to be partially addressed by #433, but not completely - i don't know what's going on there yet, the LaTeX looks reasonable. Not sure if you want to make this a blocker.

Not a blocker, I believe.

For the build command:

```
diff --git a/build.sh b/build.sh
index 6a8ed10..9050e6a 100755
--- a/build.sh
+++ b/build.sh
@@ -26,5 +26,5 @@ pandoc \
     -o "void-docs.7" "void-docs.md"
 
 # Build PDF
-pdflatex -output-directory=book/latex/ book/latex/handbook.tex
-pdflatex -output-directory=book/latex/ book/latex/handbook.tex
+pdflatex -output-directory=book/latex/ book/latex/handbook.tex >/dev/null
+pdflatex -output-directory=book/latex/ book/latex/handbook.tex >/dev/null
```

In `live-images/partition`, `< 2GB` is being shown as `! 2 GB`.

Oh, the default groups table is a bit screwed D: 

The Intel video acceleration one is too.

flexibeast:
Good point re. the `>/dev/null`s - i've just added them locally.

Thanks for noting the other issues - i've added them to my TODO. :-)

flexibeast:
Fixed:

* "< 2GB" being shown as "! 2 GB";
* the default groups table;
* the Intel video acceleration table;
* the style guide example.

Re. the Tor mirrors table, the LaTeX looks fine; i don't understand what's going on. Would it be okay to leave it for now, and come back to it later, as with that "Hashboot" para?

Re. the ToC. The issue is that the `\tableofcontents` looks for LaTeX `\section`s for the top level, and each Handbook section (i.e. Web page) is a 'section' in this context. In other words, unlike the HTML backend, it's not creating a ToC from the directory structure. i imagine we could manually replicate the Web site's ToC layout, though of course this would need to be manually changed every time the source structure is changed. Is this a blocker? If not, i can come back to this later. If so, i'll get to work. :-)

ericonr:
I don't think any of these issues are blockers! So we can get to them in time.

flexibeast:
Okay. :-) So is there anything else that needs to be done, or do you think this might be ready to merge?

ericonr:
I don't think so. Might want to wait a bit to see if anyone has any comments. I guess we could list the to-do items in #407 ?

ericonr:
[handbook.pdf](https://github.com/void-linux/void-docs/files/4985723/handbook.pdf)

If anyone wants to try it out :D

flexibeast:
@ericonr: So do you think we should we wait longer for any feedback, or can we now merge?

ericonr:
I asked @bobertlo to take a final look before merging, so it would be nice to wait for him.

Could you open an issue with the issues that still need to be solved in the generated PDF?

flexibeast:
> I asked @bobertlo to take a final look before merging, so it would be nice to wait for him.

Sure.

> Could you open an issue with the issues that still need to be solved in the generated PDF?

i added them in a [comment](https://github.com/void-linux/void-docs/issues/407#issuecomment-664746888) on #407, as you suggested [here](https://github.com/void-linux/void-docs/pull/416#issuecomment-664737122); do you think it would be better to have a distinct issue for them?

ericonr:
> i added them in a comment on #407, as you suggested here; do you think it would be better to have a distinct issue for them?

You can see I was making this comment at 4 in the morning :P Yes, you did list them there, and yes, it is a great place to put them. Sorry for the noise!

bobertlo:
I don't see any blockers on this. I have opinions on improvements that I will put somewhere else but I'm really surprised at the quality of this after being away for so long. I want to print this and bind it!

flexibeast:
@bobertlo: Thank you. :-) Merging.
@ericonr: Any thoughts on deployment? i'm guessing some changes need to be made to `mdbook_build.sh.j2` in the `mdbook` Ansible role?

PR #418, opened 2020-07-24
network-filesystems.md: changed wrong service names to run nfs

felipevencato:
Correction for the wrong names of the services that need to be activated to use nfs. Someone mistook the package names for the service names.


ericonr:
Thanks!

PR #420, opened 2020-07-24
history/index,xorg: Remove unneeded `/`, add final newline.

flexibeast:


ericonr:
AFAIK vmdfmt doesn't allow for empty newlines

flexibeast:
Ah. That will mean making `mdbook-latex` a bit more complicated, but okay. Thanks for the heads-up.

bobertlo:
I originally had a trailing newline but people didn't like it. A single stable output for version control is sort of the point :)

flexibeast:
@bobertlo: i.e. make sure there's never a trailing newline, rather than possibly having to insert one to keep everything consistent? If so, fair enough. :-)

bobertlo:
Yes. There are commits adding a newline to every file and then removing it lol.

Any valid markdown with the same content should produce the same output byte for byte.

PR #428, opened 2020-07-26
add details for GRUB on BIOS and GPT

### LINKED ISSUE #425: Elaborate on GPT partition table + BIOS boot
[Section in question](https://docs.voidlinux.org/installation/live-images/partitions.html#bios-system-notes)

> It is possible to install a GPT partition table on a BIOS system, but grub will need a special partition to boot properly.

I'm hoping for a little more elaboration on this or a link to an article describing what *specifically* GRUB needs to work (which is, if I understand correctly, a 1MB partition with type `BIOS boot`).

I think this improvement might save some trouble with some users (like me). :)

harrisonthorne:
Closes #425

Adds some details for GRUB with a BIOS system using a GPT partition table. The details add some clarity about what one should do if they want to install GRUB on a BIOS system with a GPT partition table instead of an MBR table.

### DIFF
--- a/src/installation/live-images/partitions.md
+++ b/src/installation/live-images/partitions.md
@@ -43,7 +43,7 @@ following table has recommendations for swap partition size.
 
 | System RAM | Recommended swap space | Swap space if using hibernation |
 |------------|------------------------|---------------------------------|
-| &lt; 2GB   | 2x the amount of RAM   | 3x the amount of RAM            |
+| < 2GB      | 2x the amount of RAM   | 3x the amount of RAM            |
 | 2-8GB      | Equal to amount of RAM | 2x the amount of RAM            |
 | 8-64GB     | At least 4GB           | 1.5x the amount of RAM          |
 | 64GB       | At least 4GB           | Hibernation not recommended     |
--- a/src/installation/live-images/partitions.md
+++ b/src/installation/live-images/partitions.md
@@ -19,9 +19,12 @@ The following sections will detail the options for partition configuration.
 ## BIOS system notes
 
 It is recommended that you create an MBR partition table if you are using a BIOS
-boot system. This will limit the number of partitions you create to four. It is
-possible to install a GPT partition table on a BIOS system, but grub will need a
-special partition to boot properly.
+boot system. This will limit the number of partitions you create to four.
+
+It is possible to install a GPT partition table on a BIOS system, but grub will
+need a special partition to boot properly. To do this, create a partition of
+1MB at the beginning of your disk, and give it the type `BIOS boot`. This
+partition shall not have a filesystem; grub should install successfully.
 
 ## UEFI system notes
 
--- a/src/installation/live-images/partitions.md
+++ b/src/installation/live-images/partitions.md
@@ -21,10 +21,10 @@ The following sections will detail the options for partition configuration.
 It is recommended that you create an MBR partition table if you are using a BIOS
 boot system. This will limit the number of partitions you create to four.
 
-It is possible to install a GPT partition table on a BIOS system, but grub will
-need a special partition to boot properly. To do this, create a partition of
-1MB at the beginning of your disk, and give it the type `BIOS boot`. This
-partition shall not have a filesystem; grub should install successfully.
+It is possible to use a GPT partition table on a BIOS system, but grub will
+require a special partition to boot properly. This partition should be at the
+beginning of your disk and have a size of 1MB, with type BIOS boot. Don't create
+any filesystem in it. GRUB should install successfully.
 
 ## UEFI system notes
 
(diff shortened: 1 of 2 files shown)


src/installation/live-images/partitions.md
```diff
@@ -19,9 +19,12 @@ The following sections will detail the options for partition configuration.
 ## BIOS system notes
 
 It is recommended that you create an MBR partition table if you are using a BIOS
-boot system. This will limit the number of partitions you create to four. It is
-possible to install a GPT partition table on a BIOS system, but grub will need a
-special partition to boot properly.
+boot system. This will limit the number of partitions you create to four.
+
+It is possible to install a GPT partition table on a BIOS system, but grub will
```
ericonr, line 24: "to use a GPT"

"but GRUB will require"

src/installation/live-images/partitions.md
```diff
@@ -19,9 +19,12 @@ The following sections will detail the options for partition configuration.
...
+need a special partition to boot properly. To do this, create a partition of
+1MB at the beginning of your disk, and give it the type `BIOS boot`. This
+partition shall not have a filesystem; grub should install successfully.
```
ericonr, line 27: "This partition should be at the beginning of your disk and have a size of 1MB, with type `BIOS boot` (I guess there's a number for that type too? Would be nice to include it). Don't create any filesystem in it. GRUB should install successfuly."

harrisonthorne:
It looks to me like the "Prepare Filesystems" section in chroot.md already has a link to the partitions.md file. Is that okay or did you have something different in mind?

src/installation/live-images/partitions.md
```diff
@@ -40,7 +43,7 @@ following table has recommendations for swap partition size.
 
 | System RAM | Recommended swap space | Swap space if using hibernation |
 |------------|------------------------|---------------------------------|
-| < 2GB      | 2x the amount of RAM   | 3x the amount of RAM            |
+| &lt; 2GB   | 2x the amount of RAM   | 3x the amount of RAM            |
```
bobertlo, line 46: Is this necessary for some of the other stuff you all have going on? Otherwise, We'd like to keep the markdown as human readable as possible.

harrisonthorne: Ah. It was causing some syntax highlighting issues with my editor. Should I change it back?

ericonr: Please do c:

harrisonthorne: Will do! Sorry about that.

ericonr:
> It looks to me like the "Prepare Filesystems" section in chroot.md already has a link to the partitions.md file. Is that okay or did you have something different in mind?

Good point, I think it's good enough.

src/installation/live-images/partitions.md
ericonr: No worries ;)

src/installation/live-images/partitions.md
```diff
@@ -19,9 +19,12 @@ The following sections will detail the options for partition configuration.
...
 
 It is recommended that you create an MBR partition table if you are using a BIOS
-boot system. This will limit the number of partitions you create to four. It is
-possible to install a GPT partition table on a BIOS system, but grub will need a
-special partition to boot properly.
+boot system. This will limit the number of partitions you create to four.
+
+It is possible to use a GPT partition table on a BIOS system, but grub will
+require a special partition to boot properly. This partition should be at the
+beginning of your disk and have a size of 1MB, with type BIOS boot. Don't create
```
bobertlo, line 26: Is `BIOS Boot` the name in fdisk/cfdisk? I wonder if it would be useful to provide the hex byte here for readers?

bobertlo: Is it `ef` - `EFI (FAT-12/16/`?

bobertlo: Sorry, I got this backwards in my head, but still I think this would be useful with a little more detail. But until there are more specific details I would rather not include this.

harrisonthorne: `BIOS boot` is the name in cfdisk, yeah. Is the hex byte the long-ish string cfdisk shows at the bottom of the partition type selection screen?

Well I guess it wouldn't be so long if it's just a byte lol

harrisonthorne: Check out [this post](https://askubuntu.com/questions/703443/gdisk-hex-codes)

I'm guessing you mean adding the GUID since this is GPT we're talking about, is that right?

bobertlo: Yes. Sorry I don't touch EFI much when I don't need to, and never run GPT without it. 😅

ericonr, line 26:
```
It is possible to use a GPT partition table on a BIOS system, but GRUB will
require a special partition to boot properly. This partition should be at the
beginning of your disk and have a size of 1MB, with type `BIOS boot`. Don't create
```

harrisonthorne: I think there may be a few other places where "GRUB" needs to be capitalized. Should I do a global search-and-replace for that?

ericonr: If you'd like to do it, yes, but do so in a separate PR. And there _are_ cases where "grub" might be correct, specially if it's inside a code block.

ericonr:
@harrisonthorne thanks! I capitalized "GRUB" and squashed your commits before merging. If you'd like to open another PR to replace "grub" around void-docs where it's wrong, that'd be awesome!

PR #430, opened 2020-07-27
common-errors: Fix incorrect quoting.

flexibeast:


src/xbps/troubleshooting/common-errors.md
```diff
@@ -35,7 +35,7 @@ files refer to [the correct repositories](../repositories/index.md).
 An "unresolvable shlib" error, such as:
 
 ```
-libllvm8-8.0.1_2: broken, unresolvable shlib `libffi.so.6'
+libllvm8-8.0.1_2: broken, unresolvable shlib 'libffi.so.6'
```
ericonr, line 38: Weirdly enough, this is copied straight from command output.

flexibeast: Ah. That's like Emacs-docs-style quoting, with " ` " and " ' " being used as leading and trailing limiters, respectively. i think changing the quoting here is the easiest solution in terms of PDF generation, even though it doesn't exactly match output. Longer-term, perhaps the XBPS output should be changed?

ericonr: Just asked about it on IRC, might open an issue later too. Changing it here is fine, but the commit message should note that it is being changed to help with PDF gen. Ideally we would be able to have arbitrary formatting inside a code block, but idk how that would work.

flexibeast: Yes, i agree that would be ideal; and the ideal way to handle that (and other things as well) would be to make use of a proper Markdown parser. (The JSON output produced by mdBook doesn't know about Markdown; the contents of a page are simply a serialised string.)

For the moment, rather than merging this, i'll try using lookbehind to check the  surrounding context, which might solve other issues along with this one.

ericonr: Yeah, the JSON output is really only useful for giving us the file structure, unfortunately.

flexibeast: Ah, i remember why i wasn't already using lookbehind: variable-width lookbehind, which is what i believe is required here, was only introduced - as an experimental feature - in Perl 5.30, which we can't assume people are running. So i've changed the commit message to indicate a temporary change, and we can handle this properly later.

ericonr: Since this is mainly something for Void, and will be deployed by Void servers, I don't have too much of an issue with requiring something that is already present in a version we ship.

ericonr: That said, if you prefer this workaround, I'm ok with it.

flexibeast: \*nod\* Fair point. Well, it turns out that the quoting in other output messages on that poge are [also causing problems](https://github.com/void-linux/void-docs/pull/416#issuecomment-664179155), and might in fact be the source of most of the problems you noted re. that page. So it you confirm that those messages are character-for-character accurate, i'll just have to work out better way of handling code blocks (and this commit could then be reverted - sorry).

ericonr: No worries!

flexibeast:
Reverted by #434.

PR #436, opened 2020-07-29
kde.md: Add section on dolphin

cinerea0:
This commit adds information from [the wiki page on dolphin](https://wiki.voidlinux.org/Dolphin) to the handbook page on KDE. The tracking issue (#327) mentions that this material might be good for the KDE section, so I ported it over as best as I could.

### DIFF
--- a/src/config/graphical-session/kde.md
+++ b/src/config/graphical-session/kde.md
@@ -17,3 +17,17 @@ instead.
 If you wish to start an X-based session from the console, use
 [startx](./xorg.md#startx) to run `startplasma-x11`. For a Wayland-based
 session, run `startplasma-wayland` directly.
+
+## Dolphin
+
+Dolphin is the default file manager of the KDE desktop environment. It can be
+installed on its own by installing the `dolphin` package, or it can be installed
+as part of the `kde5-baseapps` meta-package.
+
+### Thumbnail Previews
+
+To enable thumbnail file previews, install the `kdegraphics-thumbnailers`
+package. If you want video thumbnails, the `ffmpegthumbs` package is also
+necessary. Enable previews in "Control" -> "Configure Dolphin" -> "General" ->
+"Previews" by checking the corresponding boxes. File previews will be shown for
+the selected file types after clicking "Preview" in Dolphin's toolbar.


src/config/graphical-session/kde.md
```diff
@@ -17,3 +17,23 @@ instead.
 If you wish to start an X-based session from the console, use
 [startx](./xorg.md#startx) to run `startplasma-x11`. For a Wayland-based
 session, run `startplasma-wayland` directly.
+
+## Dolphin
+
+Dolphin is the default file manager of the KDE desktop environment. It can be
+installed on its own by installing the `dolphin` package, or it can be installed
+as part of the `kde5` or `kde5-baseapps` packages.
```
ericonr, line 25: "It can be installed on its own by installing the `dolphin` package or as a dependency of the `kde5-baseapps` meta-package."

`dolphin` isn't a dependency of `kde5`.

src/config/graphical-session/kde.md
```diff
@@ -17,3 +17,23 @@ instead.
...
+
+### Thumbnail Previews
+
+To have properly working thumbnail file previews, the packages
+`ffmpegthumbnailer` and `kdegraphics-thumbnailers` have to be installed. Enable
```
ericonr, line 30: I'm curious why `ffmpegthumbnailer` isn't a dependency for the kde graphics program. Would you mind opening an issue (or better yet, PR) for it in void-packages?

src/config/graphical-session/kde.md
```diff
@@ -17,3 +17,23 @@ instead.
...
+previews in "Control" -> "Configure Dolphin" -> "General" -> "Previews" by
+checking the corresponding boxes. File previews will be shown for the selected
+file types after clicking "Preview" in dolphin's toolbar.
```
ericonr, line 33: I guess dolphin should be capitalized? So "in Dolphin's toolbar".

src/config/graphical-session/kde.md
```diff
@@ -17,3 +17,23 @@ instead.
...
+
+### Integrated Terminal
+
+The integrated terminal function (key F4) opens a terminal emulator inside
+dolphin's window, on the same directory being currently browsed. It is
+configured to require Konsole (KDE's default terminal emulator) by default.
```
ericonr, line 39: I don't think this part is very Void specific, what do you think? Not sure we should add it.

src/config/graphical-session/kde.md
cinerea0: I've fixed it.

src/config/graphical-session/kde.md
cinerea0: I've capitalized it.

src/config/graphical-session/kde.md
cinerea0: You're right, I just moved it over because it was in the wiki. It's been removed.

cinerea0:
Whoops, forgot to commit before pushing.

src/config/graphical-session/kde.md
cinerea0: Sure thing, just opened a PR: https://github.com/void-linux/void-packages/pull/23936

ericonr: Great, thanks! Might make sense to ping th  maintainer or Johnnynator there.

ericonr:
This is soft-blocked ony your PR.

cinerea0:
@ericonr After getting some information from Johnnynator in that PR, I believe my most recent commit now contains the most accurate information. We cannot add `ffmpegthumbs` as a dependency since it isn't. Since the PR will not be accepted, it is no longer blocking.

ericonr:
> Is there a problem with the Void version of the package?

It was hanging, right? I've openened an issue with upstream to investigate!

src/config/graphical-session/kde.md
```diff
@@ -17,3 +17,17 @@ instead.
...
+installed on its own by installing the `dolphin` package, or it can be installed
+as part of the `kde5-baseapps` package.
+
+### Thumbnail Previews
+
+To have properly working thumbnail file previews, the packages `ffmpegthumbs`
+and `kdegraphics-thumbnailers` have to be installed. Enable previews in
+"Control" -> "Configure Dolphin" -> "General" -> "Previews" by checking the
+corresponding boxes. File previews will be shown for the selected file types
+after clicking "Preview" in Dolphin's toolbar.
```
ericonr, line 33: I'm a bit confused now. The necessary package is `ffmpegthumbs`, then, not `ffmpegthumbnailer`?

src/config/graphical-session/kde.md
```diff
@@ -17,3 +17,17 @@ instead.
 If you wish to start an X-based session from the console, use
 [startx](./xorg.md#startx) to run `startplasma-x11`. For a Wayland-based
 session, run `startplasma-wayland` directly.
+
+## Dolphin
+
+Dolphin is the default file manager of the KDE desktop environment. It can be
+installed on its own by installing the `dolphin` package, or it can be installed
+as part of the `kde5-baseapps` package.
```
ericonr, line 25: "meta-package"

ericonr:
Would you mind changing the commit message to `Add` instead of `Added`?

src/config/graphical-session/kde.md
cinerea0: Yes, according to Johnnynator. However, since `ffmpegthumbs` is independent of `kdegraphics-thumbnailers` and is only used to generate video thumbnails, it can't be added as a dependency.

ericonr: I see. Could you make that clear, then? Something like "To enable thumbnail file previews, install the `kdegraphics-thumbnailer` package. If you want video thumbnails, the `ffmpegthumbs` package is also necessary"

cinerea0:
I've made the changes above and changed the commit message for the last push.

ericonr:
Great, thanks!

PR #438, opened 2020-07-29
kde: add dbus dep for sddm.

ericonr:
@flexibeast does this line up with your experiments?

We had someone on reddit need to enable dbus in order to get SDDM working.


flexibeast:
TLA;DR - Too Long Ago; Don't Remember. :-) Would you like me to create a VM to test?

ericonr:
Hmm if it's easy for you to, sure. Otherwise, maldridge confirmed it on IRC so I would be ok with merging since you don't remember it for sure.

flexibeast:
i think if maldridge has confirmed it, that's probably sufficient. :-)

ericonr:
Thanks!

PR #440, opened 2020-08-02
<various>: Ensure consistency between SUMMARY titles and section titles.

flexibeast:
Needed for generating a docs.voidlinux.org-style ToC for the PDF version of the Handbook.


src/config/rc-files.md
```diff
@@ -1,4 +1,4 @@
-# rc.conf, rc.local and rc.shutdown
+# rc Files
```
ericonr, line 1: I think I prefer the original name. What do you think?

src/installation/guides/index.md
```diff
@@ -1,4 +1,4 @@
-# Advanced Installation Guides
+# Advanced Installation
```
ericonr, line 1: Same here.

src/config/rc-files.md
flexibeast: i don't mind it, it's just that it's rather long, which is not ideal for the ToC on the Web site. Happy to change it if you think this isn't an issue.

src/installation/guides/index.md
flexibeast: Changed locally.

src/config/rc-files.md
ericonr: Just tested it locally, it has the same width as the title for "Session management", so it looks like it should be fine.

PR #443, opened 2020-08-02
[RFC] history: add more interesting dates.

ericonr:
Sources:

- systemd: ccc13eae418d9f827c4505202dc3eebaccfb00aa in void-packages
- aarch64: 0f3b729b0839581b8d84b544af5d13049a15ed9a in void-packages

@leahneukirchen do you think that's a reasonable date for aarch64?

I would like to add a few more dates:

- first in person meetup - depends on someone who knows that info
- full transition to handbook - we need to decide if the wiki is being spun back up
- switch to transparent infra - suggested by maldridge, would have to find a date for it

I would like to change the date format to something more reasonable, if there's consensus for it.

### DIFF
--- a/src/about/history.md
+++ b/src/about/history.md
@@ -1,10 +1,15 @@
 # History
 
-Knowledge of the ancients, grepped from the git logs themselves:
+Knowledge of the ancients, grepped from the Git logs themselves:
 
-- Sep 26 2008: first git import of
+- 2008-09-26: first Git import of
    [void-packages](https://github.com/void-linux/void-packages)
-- Aug 17 2009: first git import of [xbps](https://github.com/void-linux/xbps)
-- Mar 1 2013: first [musl](https://www.musl-libc.org/) toolchains added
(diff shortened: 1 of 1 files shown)


flexibeast:
> I would like to change the date format to something more reasonable, if there's consensus for it.

+1 for this, preferably to ISO8601 (or some variant thereof).

leahneukirchen:
Probably, yes. I ran aarch64 for a bit with a foreign kernel.

bobertlo:
The current date format is literally copied and pasted from git logs. No other reason it is that way so no reason not to change it.

Piraty:
i like it

flexibeast:
@ericonr: Is there anything more you're thinking of adding? If not, i'd like to merge this. :-) Could you please upper-case mentions of 'Git'?

ericonr:
Not for now. I can open a new PR once we've closed the wiki :D

ericonr:
Updated c:

PR #455, opened 2020-08-07
user-services.md: Added alias to manage user services.

lordfeck:
The manual suggests setting the $SVDIR variable before calling `sv` on a user service directory. One such way to make user service management more convenient is to add an alias to your shell's init script. I've added an example of this using the alias `svl` to the relevant handbook page.

### DIFF
--- a/src/config/services/user-services.md
+++ b/src/config/services/user-services.md
@@ -45,5 +45,4 @@ ok: run: gpg-agent: (pid 19818) 0s
 ```
 
 It may be convenient to adapt the above example to a shell alias such as
-`svl="SVDIR=~/service sv"` for your user shell. This makes management of user
-services trivial.
+`svl="SVDIR=~/service sv"`. This makes management of user services trivial.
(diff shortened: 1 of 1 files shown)


flexibeast:
@ericonr: Thoughts on this?

ericonr:
I don't know whether we should add this. There are several different ways of adding shortcuts to one's shell, and explanations for how to use your shell and add aliases to it don't really feel in scope.

I'd feel better if it could be made more generic, I think.

lordfeck:
@ericonr I could strip away the explanation and reduce it to just a line or suggestion?

ericonr:
@lordfeck I think that might work better, yes!

lordfeck:
@ericonr trimmed now

flexibeast:
Merged, thanks!

Duncaen:
This is kinda useless, just set `SVDIR` in your shells profile, if you want to manage root services, you use sudo anyways and the environment variable will be unset.

PR #456, opened 2020-08-08
man: Multiple changes.

flexibeast:
* Mention manual sections.
* Mention `man.conf`.
* Mention `makewhatis` cron job.
* Move section to more prominent position in `SUMMARY`.
* Improve phrasing.


src/config/man.md
```diff
@@ -1,19 +1,44 @@
...
+There are some man pages which have the same name, but are used in different
+contexts, and are thus in a different section. You can specify which one to use
+by including the section number in the call to `man`:
+
+```
+$ man 1 printf
+```
+
+`man` can be configured via the
+[man.conf(5)](https://man.voidlinux.org/man.conf.5) file, `/etc/man.conf`.
```
ericonr, line 36: Style question: when the man page specifies where the file should be located, I tend to write something like "can be configured in man.conf(5).", without specifying the file itself. What do you think?

flexibeast: Good point; agreed. PR updated.

PR #457, opened 2020-08-08
cron,services/index: Multiple changes.

flexibeast:
* cron:
  - link to "cron" page on Wikipedia.
  - provide brief introduction.
  - mention crontab(1).
  - mention the `crond` service.
* services/index:
  - mention option of using `cron` to run something regularly, link to
  "cron" section.

### DIFF
--- a/src/config/cron.md
+++ b/src/config/cron.md
@@ -1,13 +1,27 @@
 # Cron
 
-Void Linux comes without a default cron daemon, you can choose one of multiple
-cron implementations, by installing the package and enabling the system service.
+[cron](https://en.wikipedia.org/wiki/Cron) is a daemon for running programs at
+regular intervals. The programs and intervals are specified in a `crontab` file,
+which can be edited with [crontab(1)](https://man.voidlinux.org/crontab.1).
+Running `crontab -e` as the superuser will edit the system crontab; otherwise,
+it will edit the crontab for the current user.
 
-Available choices include [cronie](https://github.com/cronie-crond/cronie/),
+By default, a cron daemon is not installed. However, multiple cron
+implementations are available, including
+[cronie](https://github.com/cronie-crond/cronie/),
 [dcron](http://www.jimpryor.net/linux/dcron.html),
 [fcron](http://fcron.free.fr/) and more.
 
-As alternative to the standard cron implementations you can use something like
-[snooze](https://github.com/leahneukirchen/snooze) or
-[runwhen](https://code.dogmap.org/runwhen/) which go hand in hand with service
-supervision.
+Once you have chosen and installed an implementation,
+[enable](./services/index.md#enabling-services) the `crond` service, which is a
+symlink to the actual service (e.g. `dcron`). If you install several
+implementations, you can choose which one to use via
+[xbps-alternatives(1)](https://man.voidlinux.org/xbps-alternatives.1); this will
+alter the `crond` symlink appropriately. Implementation documentation will be
+available in [crond(8)](https://man.voidlinux.org/crond.8).
+
+As an alternative to the standard cron implementations, you can use
+[snooze(1)](https://man.voidlinux.org/snooze.1) together with the
+`snooze-hourly`, `snooze-daily`, `snooze-weekly` and `snooze-monthly` services,
+which are provided by the `snooze` package for this purpose. Each of these
+services execute scripts in the respective `/etc/cron.*` directories.
--- a/src/config/services/index.md
+++ b/src/config/services/index.md
@@ -12,6 +12,10 @@ Some advantages of using runit include:
 - a reliable logging facility for services, where the log service stays up as
    long as the relevant service is running and possibly writing to the log.
 
+If you don't need a program to be running constantly, but would like it to run
+at regular intervals, you might like to consider using a [cron
+daemon](../cron.md).
+
 ## Section Contents
 
 - [Per-User Services](./user-services.md)


src/config/cron.md
```diff
@@ -1,13 +1,28 @@
 # Cron
 
-Void Linux comes without a default cron daemon, you can choose one of multiple
-cron implementations, by installing the package and enabling the system service.
+[cron](https://en.wikipedia.org/wiki/Cron) is a daemon for running programs at
+regular intervals. The programs and intervals are specified by using
+[crontab(1)](https://man.voidlinux.org/crontab.1) to edit a `crontab` file.
```
ericonr, line 5: `a crontab(5) file`?

src/config/cron.md
```diff
@@ -1,13 +1,28 @@
...
 [fcron](http://fcron.free.fr/) and more.
 
-As alternative to the standard cron implementations you can use something like
-[snooze](https://github.com/leahneukirchen/snooze) or
-[runwhen](https://code.dogmap.org/runwhen/) which go hand in hand with service
-supervision.
+Once you have chosen and installed an implementation,
+[enable](./services/index.md#enabling-services) the `crond` service.
+Implementation documentation will be available in
+[crond(8)](https://man.voidlinux.org/crond.8).
```
ericonr, line 24: Perhaps make sure to mention that if you have several installed it is necessary to use `xbps-alternatives` to choose between them, or at least that the `crond` service is controlled by alternatives.

src/config/cron.md
```diff
@@ -1,13 +1,28 @@
...
+[cron](https://en.wikipedia.org/wiki/Cron) is a daemon for running programs at
+regular intervals. The programs and intervals are specified by using
+[crontab(1)](https://man.voidlinux.org/crontab.1) to edit a `crontab` file.
+`crontab` files can be system-wide or user-specific; to edit the system crontab,
+run `crontab` as the superuser:
 
-Available choices include [cronie](https://github.com/cronie-crond/cronie/),
+```
+# crontab -e
+```
```
ericonr, line 11: "it is necessary to run `crontab -e` as the superuser", avoids the code block and I prefer the structure. What do you think?

src/config/cron.md
```diff
@@ -1,13 +1,28 @@
...
-[runwhen](https://code.dogmap.org/runwhen/) which go hand in hand with service
-supervision.
+Once you have chosen and installed an implementation,
+[enable](./services/index.md#enabling-services) the `crond` service.
+Implementation documentation will be available in
+[crond(8)](https://man.voidlinux.org/crond.8).
+
+As an alternative to the standard cron implementations, you can use things like
+[snooze(1)](https://man.voidlinux.org/snooze.1), which go hand-in-hand with
+service supervision.
```
ericonr, line 28: Why remove the `runwhen` mention? Perhaps add that the `snooze` package comes with `snooze-{hourly,daily,weekly,monthly}` services?

src/config/cron.md
flexibeast: Unfortunately, not all implementations install a `crontab(5)` man page; `dcron` (which i use) certainly doesn't. And the exact things allowed by crontabs can vary from impl. to impl.

src/config/cron.md
flexibeast: Sure, i've done the same thing in a cron guide i've started. :-) Will change.

src/config/cron.md
flexibeast: Because `runwhen` doesn't seem to be in the repos. Are we okay to suggest software that's not packaged?

src/config/cron.md
flexibeast: Sure, will do.

src/config/cron.md
ericonr: Oh, I see :c 

Fair enough, then.

src/config/cron.md
flexibeast: Oh, and sure, will mention those things about `snooze`.

ericonr: Lol, definitely not. Good catch!

src/config/cron.md
flexibeast: Oh and also, even when there's such a file, it's not necessarily literally called `crontab`: cf. `fcrontab`, `incrontab`.

src/config/cron.md
flexibeast: :-) Still, i'll update the commit message to explain why `runwhen` is no longer mentioned (which i should have done in the first place).

flexibeast: i've now mentioned the `snooze-*` services, listing them individually because i couldn't find a way to use the abbreviated (non-sh) shell-syntax version without it feeling awkward.

flexibeast:
@ericonr: Thanks for the review; PR updated.

src/config/cron.md
```diff
@@ -1,13 +1,28 @@
...
-supervision.
+Once you have chosen and installed an implementation,
+[enable](./services/index.md#enabling-services) the `crond` service, which is a
+symlink to the actual service (e.g. `dcron`). If you install several
+implementations, you can choose which one to use via
+[xbps-alternatives(1)](https://man.voidlinux.org/xbps-alternatives.1); this will
+alter the `crond` symlink appropriately.
+
+Implementation documentation will be available in
+[crond(8)](https://man.voidlinux.org/crond.8).
```
ericonr, line 23: Merge these two into a single paragraph? I'm not sure.

src/config/cron.md
```diff
@@ -1,13 +1,28 @@
...
+
+As an alternative to the standard cron implementations, you can use
+[snooze(1)](https://man.voidlinux.org/snooze.1), which goes hand-in-hand with
+service supervision. The `snooze` package provides the `snooze-hourly`,
+`snooze-daily`, `snooze-weekly` and `snooze-monthly` services.
```
ericonr, line 28: "services for this purpose." ? And perhaps add that they execute the scripts in the respective `/etc/cron.*` folder?

src/config/cron.md
flexibeast: i wasn't sure either, but since you had the same thought: paras merged. :-)

src/config/cron.md
flexibeast: \*nod\* i've rephrased that entire para; LMKWYT.

PR #460, opened 2020-08-11
connman.md: Add connection workflow

devlaf:
**Description**
This PR pulls in the wifi connection workflow from the old connman wiki page.

**Note**
> While this handbook does not provide a large amount of copy and paste configuration instructions, it does provide links to the man pages for the referenced software wherever possible.

Not sure if this addition is in-line with the ethos outlined here on the "About This Handbook" page, but I have found that including this in the docs can be very handy. Personally, if I'm browsing this page it's usually from mobile while I'm looking for a quick reminder to get my real system back up and connected, and it's nice to have the workflow outlined right there.

### DIFF
--- a/src/config/network/connman.md
+++ b/src/config/network/connman.md
@@ -15,10 +15,27 @@ Finally, enable the `connmand` service.
 
 ## Configuring ConnMan
 
+### ConnMan CLI
+
 The `connman` package includes a command line tool,
 [connmanctl(1)](https://man.voidlinux.org/connmanctl.1) to control network
 settings. If you do not provide any commands, `connmanctl` starts as an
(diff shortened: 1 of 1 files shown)


flexibeast:
Thanks! i think it's reasonable to include this. A couple of things:

* In the phrase "using the connmanctl interactive shell", please put backticks around "connmanctl", as in the previous paragraph.
* Does `connmanctl` work when run as a regular (i.e. non-superuser) user? If not, the `$` in the example should be changed to `#`.

devlaf:
Thanks, updated accordingly!

flexibeast:
Merged. Thanks!

PR #463, opened 2020-08-12
efi grub install instructions

anjandev:


ericonr:
Do you think you can fit this information in the `chroot` guide instead? At some point the `fde` guide will be converted into a shorter guide that refers to the chroot guide for as many steps as possible.

ericonr:
Thanks!

Issue #350, opened 2020-06-23
Add "Containers and VMs" section to Handbook


flexibeast:
Such a section should mention at least:

* LXC
* Docker
* Podman
* QEMU
* libvirt

and maybe VirtualBox.

Frick-David:
As listed I will cover podman and docker. If desired, I would also like to add a section on kubernetes. Is that agreeable?

flexibeast:
Mm, i tend to think that Kubernetes is more out-of-scope, given it's not just containerisation, but orchestration as well .... But i'd be interested to know what other Void team members think.

Frick-David:
I thought that too. Let us see!

flexibeast:
@void-linux/pkg-committers: Any thoughts on whether we should include a "VirtualBox" section, with the contents of [the old wiki page](https://wiki.voidlinux.org/VirtualBox)?

Piraty:
@flexibeast why not?

flexibeast:
@Piraty: Because it seems to me that there's a good chance it doesn't meet (my understanding of) requirements for inclusion in the Handbook.

The first part of the wiki page, about using Void as the host, is basically "Install this particular package". There's the mention of possibly needing to use `xbps-reconfigure`, but that applies to many packages that could be installed. There's the mention of possibly needing to install the relevant kernel headers, and i guess a case could be made for that being necessary information. The `virtualbox-ose` package provides the `vboxwebsrv` service, but i'm not sure that's any more crucial for us to document that particular service than any of the other hundreds of runit services Void provides, but which we don't currently document.

The second part of the wiki page is about running Void as a guest. This could be potentially useful information to provide, *if* we decide that we should also be documenting running Void in the context of other container/VM systems as well.

This all stands in contrast to documenting e.g. LXC, where the official ["getting started" information](https://linuxcontainers.org/lxc/getting-started/) gives information not relevant to Void (e.g. specifying the use of `systemd-run`).

There might be a similar issue around QEMU, where there's not even a single service file involved. If such a section is going to be more than "install `qemu`", then it's going to end up being a "getting started with QEMU" section, for which documentation is definitely available elsewhere.

i'll work with whatever the overall decision is, of course.

ericonr:
>  There's the mention of possibly needing to install the relevant kernel headers, and i guess a case could be made for that being necessary information. 

That should be documented under `dkms` already, and isn't very useful, since the `dkms` package already depends on `linux-headers`.

[...]

PR #465, opened 2020-08-12
containers-and-vms/index,libvirt: Add sections.

flexibeast:
Also move `lxc` into `containers-and-vms`.

Partially addresses #350.


src/config/containers-and-vms/libvirt.md
```diff
@@ -0,0 +1,28 @@
+# libvirt
+
+[libvirt](https://en.wikipedia.org/wiki/Libvirt) is an API and daemon for
```
ericonr, line 3: Do they have a homepage? Perhaps makes more sense to link it here.

src/config/containers-and-vms/index.md
```diff
@@ -0,0 +1,9 @@
+# Containers and Virtual Machines
+
+This section describes how to set up some of the container and virtual machine
+systems available on Void.
```
ericonr, line 4: "systems" or "tools" ?

src/config/containers-and-vms/libvirt.md
flexibeast: They do - https://libvirt.org/  - but i felt it was not particularly helpful for an introduction, at least not compared to the Wikipedia page. But yeah, you're right, it's probably more appropriate.

src/config/containers-and-vms/index.md
flexibeast: "Tools" feels odd to me here, perhaps too specific? E.g. i'd described 'grep' as a tool, but wouldn't describing (say) LibreOffice as an "office productivity tool". How about "software"?

ericonr: software sounds good!

src/config/containers-and-vms/libvirt.md
ericonr: I guess linking to the homepage is better for consistency sake? If you want to prioritize the more helpful page, I don't have any issue with it. Just wanted to bring it up.

flexibeast:
@ericonr: Thanks.
@void-linux/pkg-committers: Any comments/feedback?

src/config/containers-and-vms/libvirt.md
```diff
@@ -0,0 +1,28 @@
...
+
+The `libvirt` package provides the [virsh(1)](https://man.voidlinux.org/virsh.1)
+interface to libvirtd. `virsh` is an interactive shell and batch-scriptable tool
+for performing management tasks, including creating, configuring and running
+virtual machines, and managing networks and storage. Note that `virsh` usually
+needs to be run as root, as described in the `virsh` man page:
+
+> Most virsh commands require root privileges to run due to the communications
+> channels used to talk to the hypervisor. Running as non root will return an
+> error.
```
ahesford, line 20: This is true, but `libvirt` ships with a polkit rule `/usr/share/polkit-1/rules.d/50-libvirt.rules` that grants management privileges to users of the `libvirt` group. Thus, I suggest you add something like the following text:
```
As an alternative, if you have the `polkit` and `dbus` packages installed and you enable the `dbus` service, `libvirtd` will grant necessary privileges to any user added to the `libvirt` group.
```

flexibeast: Thanks, PR updated.

flexibeast:
@Frick-David: With this PR merged, please rebase #380 on `master`.

PR #467, opened 2020-08-15
repositories/index: Add man page link.

flexibeast:
From what i can tell, this PR will allow #107, which has now been open for a year, to be closed.

Link goes to `xtools.1`, since linking to `xdbg.1` won't work.

### DIFF
--- a/src/xbps/repositories/index.md
+++ b/src/xbps/repositories/index.md
@@ -71,8 +71,9 @@ Once enabled, symbols may be obtained for `<package>` by installing
 
 #### Finding debug dependencies
 
-The `xtools` package contains the `xdbg` utility to retrieve a list of debug
-packages, including dependencies, for a package:
+The `xtools` package contains the [xdbg(1)](https://man.voidlinux.org/xtools.1)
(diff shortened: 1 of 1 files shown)


src/xbps/repositories/index.md
```diff
@@ -71,8 +71,9 @@ Once enabled, symbols may be obtained for `<package>` by installing
 
 #### Finding debug dependencies
 
-The `xtools` package contains the `xdbg` utility to retrieve a list of debug
-packages, including dependencies, for a package:
+The `xtools` package contains the [xdbg](https://man.voidlinux.org/xtools.1)
```
ericonr, line 74: `xdbg(1)`, right?

flexibeast: i thought about that. Since the link is taking one to the `xtools.1` man page, and `xdbg.1` doesn't actually exist, i chose to omit the "(1)". i have no idea what the correct thing to do is in this case.

ericonr: That's a good point... I would prefer having the `(1)` for consistency, and it works.

PR #469, opened 2020-08-15
external-applications: Mention AppImages.

flexibeast:
Also remove blockquoting from Flatpak subsection.

@ericonr, let me know if using an AppImage on musl requires `gcompat`, and i'll update the PR accordingly.


ericonr:
`gcompat` is definitely necessary.

PR #473, opened 2020-08-15
date-time: Note that setting timezone does not set the time.

flexibeast:


flexibeast:
@ericonr: Yeah, i've noticed a few people saying "I've set the timezone, why is my clock wrong?", most recently [here](https://old.reddit.com/r/voidlinux/comments/iah3xc/time_and_date_both_inaccurate_when_time_zone_is/).

PR #474, opened 2020-08-15
kernel: Add information about using a different kernel series.

flexibeast:


src/config/kernel.md
```diff
@@ -36,6 +36,14 @@ pre-installed on every Void Linux system. This utility runs the necessary
 [hooks](#kernel-hooks) when removing old kernels. Note that `vkpurge` does not
 remove kernel *packages*, only particular *kernels*.
 
+## Using a specific kernel series
+
+To use a kernel series other than the current default, use
```
ericonr, line 41: "and be able to remove the default" ? Because this isn't necessary if you only want to install another series. It should be made clear that people can have multiple kernel series installed, because the "cleaning up my system" scenario isn't the important one here.

flexibeast: Point taken. "To change the default kernel series, and be able to remove other kernel series, ..."?

ericonr: "To be able to remove the default kernel series, after installing a different one, it is necessary to change the default" perhaps makes it clearer that this is only necessary if you want to remove stuff? I'm thinking of newbies who see this and think they need to remove the default kernel for some reason.

flexibeast: Thanks, PR updated (with changed subsection heading).

ericonr:
Given https://github.com/void-linux/xbps/issues/26 , I believe it's best we don't suggest virtual packages :/

flexibeast:
Okay, PR updated, including change to standard placeholder format.

flexibeast:
i definitely agree it's uglier! But yeah, i think it's best if we use our conventions consistently, to help readers.

Should we wait for any further comments, or can i merge?

ericonr:
Since the topic of kernel versions has a lot of divergence, we could wait a bit, I think.

PR #475, opened 2020-08-17
<various>: Remove unnecessary blockquotes.

flexibeast:
These appear to be the final few remaining blockquote usages that don't adhere to the style guide, other than the one in `external-applications.md`, which is being addressed in #469.

### DIFF
--- a/src/config/users-and-groups.md
+++ b/src/config/users-and-groups.md
@@ -28,9 +28,9 @@ output of `chsh -l`, which provides a list of installed shells.
 
 ## sudo
 
-> Note: [sudo(8)](https://man.voidlinux.org/sudo.8) is installed by default but
-> may not be configured. It is only necessary to configure sudo if you wish to
-> use it.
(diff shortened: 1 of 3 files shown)


src/config/users-and-groups.md
```diff
@@ -28,9 +28,9 @@ output of `chsh -l`, which provides a list of installed shells.
 
 ## sudo
 
-> Note: [sudo(8)](https://man.voidlinux.org/sudo.8) is installed by default but
-> may not be configured. It is only necessary to configure sudo if you wish to
-> use it.
+[sudo(8)](https://man.voidlinux.org/sudo.8) is installed by default, but might
+not be configured appropriately for your needs. It is only necessary to
+configure sudo if you wish to use it.
```
ericonr, line 33: "sudo" or "`sudo`" ?

flexibeast: i believe the former here, since we're referring to the software in general, not the specific command.

PR #478, opened 2020-08-19
about: mention void-docs package.

ericonr:


src/about/index.md
```diff
@@ -2,7 +2,8 @@
 
 Welcome to the Void Handbook! Please be sure to read the "[About This
 Handbook](./about-this-handbook.md)" section to learn how to use this
-documentation effectively.
+documentation effectively. A local copy of this handbook in several formats can
```
flexibeast, line 5: Would prefer: "handbook, in several formats, can"

ericonr: Fixed!

PR #480, opened 2020-08-20
Update fde.md

LamaRaz:
change /dev/urandom to random in order to block if entropy pool is insufficient (if setup is performed immediately after booting a live system by a script for example), else the keyfile that is generated could compromise the entire setup.
bulk up the keyfile for extra paranoia and low cost, entropy quality may not be ideal on some systems.

change the grub install line to better describe what is happening and allow for changing the grub name/id, also allow for systems that use 32 bit EFI firmware.

offer another option to get UUID.

### DIFF
--- a/src/installation/guides/fde.md
+++ b/src/installation/guides/fde.md
@@ -137,8 +137,10 @@ Do you want to import this public key? [Y/n] y
 ...
 ```
 
-UEFI systems will have a slightly different package selection. The installation
-command for a UEFI system will be as follows.
+UEFI systems will have a slightly different package selection.
+On some devices the EFI firmware is 32 bits: install grub-i386-efi instead
+To confirm EFI firmware architecture: cat /sys/firmware/efi/fw_platform_size
+The installation command for a UEFI system will be as follows.
 
 ```
 # xbps-install -Sy -R https://alpha.de.repo.voidlinux.org/current -r /mnt base-system cryptsetup grub-x86_64-efi lvm2
@@ -192,6 +194,12 @@ sda1  135f3c06-26a0-437f-a05e-287b036440a4
 ...
 ```
 
+Or in a way better suited for shell scripting:
+
+```
+# blkid | grep /dev/sda1 | cut -d'"' -f 2
+```
+
 Edit the `GRUB_CMDLINE_LINUX_DEFAULT=` line in `/etc/default/grub` and add
 `rd.lvm.vg=voidvm rd.luks.uuid=<UUID>` to it. Make sure the UUID matches the one
 for the `sda1` device found in the output of the `lsblk` command above.
@@ -201,10 +209,10 @@ configured to automatically unlock the encrypted volume on boot. First, generate
 a random key.
 
 ```
-# dd bs=512 count=4 if=/dev/urandom of=/boot/volume.key
-4+0 records in
-4+0 records out
-2048 bytes (2.0 kB, 2.0 KiB) copied, 0.000421265 s, 4.9 MB/s
+# dd bs=512 count=64 if=/dev/random of=/boot/volume.key
+64+0 records in
+64+0 records out
+32768 bytes (33 kB, 32 KiB) copied, 0.000783849 s, 41.8 MB/s
 ```
 
 Next, add the key to the encrypted volume.
@@ -235,10 +243,10 @@ a new file at `/etc/dracut.conf.d/10-crypt.conf` with the following line:
 install_items+=" /boot/volume.key /etc/crypttab "
 ```
 
-Next, install the boot loader to the disk.
+Next, install the boot loader to the disk. If the EFI firmware is 32 bits: change target to =i386-efi
 
 ```
-# grub-install /dev/sda
+# grub-install --target=x86_64-efi --efi-directory=/boot/efi --bootloader-id=void_grub --boot-directory=/boot --recheck
 ```
 
 Ensure an initramfs is generated:


flexibeast:
Thanks for your contribution! As a general point, could you please read [CONTRIBUTING](https://github.com/void-linux/void-docs/blob/master/CONTRIBUTING.md) ,and in particular, the "[Making changes](https://github.com/void-linux/void-docs/blob/master/CONTRIBUTING.md#making-changes)" subsection? CI is currently failing because it seems you've not run `./check.sh` before committing; and your commit message needs to amended to follow the format described in that subsection.

@ericonr, could you please review the content?

src/installation/guides/fde.md
```diff
@@ -235,10 +243,10 @@ a new file at `/etc/dracut.conf.d/10-crypt.conf` with the following line:
 install_items+=" /boot/volume.key /etc/crypttab "
 ```
 
-Next, install the boot loader to the disk.
+Next, install the boot loader to the disk. If the EFI firmware is 32 bits: change target to =i386-efi
 
 ```
-# grub-install /dev/sda
+# grub-install --target=x86_64-efi --efi-directory=/boot/efi --bootloader-id=void_grub --boot-directory=/boot --recheck
```
ericonr, line 249: This has been added to the `chroot` guide. Could you instead link there? We want to avoid duplicated information. And this command is only for EFI systems, while this guide covers both EFI and BIOS setups.

src/installation/guides/fde.md
```diff
@@ -137,8 +137,10 @@ Do you want to import this public key? [Y/n] y
 ...
 ```
 
-UEFI systems will have a slightly different package selection. The installation
-command for a UEFI system will be as follows.
+UEFI systems will have a slightly different package selection.
+On some devices the EFI firmware is 32 bits: install grub-i386-efi instead
+To confirm EFI firmware architecture: cat /sys/firmware/efi/fw_platform_size
+The installation command for a UEFI system will be as follows.
```
ericonr, line 143: If possible, this informatio should be transferred to the chroot guide as well. Please use inline code for commands and package names, too.

```
install `grub-i386-efi` instead
```

src/installation/guides/fde.md
```diff
@@ -201,10 +209,10 @@ configured to automatically unlock the encrypted volume on boot. First, generate
...
 
 ```
-# dd bs=512 count=4 if=/dev/urandom of=/boot/volume.key
-4+0 records in
-4+0 records out
-2048 bytes (2.0 kB, 2.0 KiB) copied, 0.000421265 s, 4.9 MB/s
+# dd bs=512 count=64 if=/dev/random of=/boot/volume.key
+64+0 records in
+64+0 records out
+32768 bytes (33 kB, 32 KiB) copied, 0.000783849 s, 41.8 MB/s
```
ericonr, line 215: Increasing key size is OK, as long as it works, but I don't know if we should recommend `/dev/random`. In a virtual machine context it might just block, and in a non virtual context it probably has enough etropy already from the hardware. https://www.2uo.de/myths-about-urandom/

LamaRaz: From the very source you linked:

> What about entropy running low?
> It doesn't matter.
> 
> The underlying cryptographic building blocks are designed such that an attacker cannot predict the outcome, as long as there was enough randomness (a.k.a. entropy) in the beginning. A usual lower limit for “enough” may be 256 bits. No more.
> 
> Considering that we were pretty hand-wavey about the term “entropy” in the first place, it feels right. As we saw, the kernel's random number generator cannot even precisely know the amount of entropy entering the system. Only an estimate. And whether the model that's the basis for the estimate is good enough is pretty unclear, too.

256 bits of true entropy is more than enough for sure, the problem is that the kernel is estimating it and if the hardware is atypical it can drastically overestimate it, that's why the larger key size - though it's an ugly hack.

I personally just take 30 seconds of /dev/random output and channel it into sha512sum and call it.

Using urandom will probably be ok but when it isn't (even the kernel is estimating it has no entropy)

/dev/random was changed btw, they removed the entropy subtraction nonsense. The only difference form urandom now I believe is blocking at the very beginning if there isn't enough entropy.

abenson: Making the "file" larger doesn't make the key larger, it's still going to just read the first (by default) 512 bits.

[cryptsetup.c](https://gitlab.com/cryptsetup/cryptsetup/-/blob/master/src/cryptsetup.c#L1730)

abenson: Really, `bs=1 count=64` would be enough.

src/installation/guides/fde.md
```diff
@@ -192,6 +194,12 @@ sda1  135f3c06-26a0-437f-a05e-287b036440a4
 ...
 ```
 
+Or in a way better suited for shell scripting:
+
+```
+# blkid | grep /dev/sda1 | cut -d'"' -f 2
```
abenson, line 200: Why not just `blkid -o value -s UUID /dev/sda1`

src/installation/guides/fde.md
Piraty: yes! the wiki suffered from too much duplication regarding "this is how to install via xbps-install into a chroot env"

src/installation/guides/fde.md
abenson: > I personally just take 30 seconds of /dev/random output and channel it into sha512sum and call it.

I don't understand what you mean by this. Do you take the actual sha512 (i.e. 512 bits) or the hex representation (meaning 128 'a-f0-9`)?

LamaRaz: The bits of course. Basically do an entropy accumulation on top of what the kernel is doing.

LamaRaz: Are you sure about the keyfile because afaik LUKs should be hashing the entire keyfile, you linked to dm-crypt which does not use LUKs key slots...

LamaRaz: I always assumed the keyfile was being hashed, if you are correct this is really poor practice.

abenson: > you linked to dm-crypt which does not use LUKs key slots

The function I linked to is `action_luksAddKey` ...

PR #483, opened 2020-08-20
config/security/hasboot: remove

ericonr:
This section is security theater. A secure boot article will supersede this.

@travankor After some discussion on IRC, it's been agreed that this commit makes sense. I would like to push it into void-docs with attribution to you, do you agree with it?

(I can make the necessary changes for it to be mergeable, too)


teldra:
Can be removed, since I plan to remove hashboot from repos because its inofficially abandoned. (And its not void specific)

flexibeast:
Once this is removed, should there be a new release of the `void-docs` package? i.e. was this section sufficiently problematic that we want to minimise the number of people trying to use it?

teldra:
Maybe, it needs to be removed from Summary.md or is this autogenerated?

fosslinux:
Why is this security theater? It certainly doesn't protect against the changing of `/boot` but it will at least alert the operator.

fosslinux:
Oh, I understand now.

PR #484, opened 2020-08-20
Good stuff from #480

abenson:
Also removed some trailing whitespaces throughout.


abenson:
Regarding the proposed changes for [#L204](https://github.com/void-linux/void-docs/blob/master/src/installation/guides/fde.md#L204), Also, given that only the first n-bits of the file is read [cryptsetup/lib/cli/cli.c](https://gitlab.com/cryptsetup/cryptsetup/-/blob/master/lib/cli/cli.c#L302), should the random keyfile size be reduced?  The default for Void (and most distros, I believe) is 256, so a 32-byte file would suffice.

src/installation/guides/fde.md
```diff
@@ -185,16 +185,13 @@ Next, the kernel needs to be configured to find the encrypted device. First,
...
-sda1  135f3c06-26a0-437f-a05e-287b036440a4
-...
+# blkid -o value -s UUID /dev/sda1
+135f3c06-26a0-437f-a05e-287b036440a4
 ```
 
 Edit the `GRUB_CMDLINE_LINUX_DEFAULT=` line in `/etc/default/grub` and add
 `rd.lvm.vg=voidvm rd.luks.uuid=<UUID>` to it. Make sure the UUID matches the one
-for the `sda1` device found in the output of the `lsblk` command above.
+for the `sda1` device found in the output of the `blkid` command above.
```
ericonr, line 194: `[blkid(8)](https://man.voidlinux.org/blkid.8)`

PR #485, opened 2020-08-21
hashboot: remove

teldra:
closes https://github.com/void-linux/void-docs/pull/483


flexibeast:
@ericonr: Agreed. Merging.

PR #494, opened 2020-08-23
locales: change title, add note about libreoffice.

ericonr:


src/config/locales.md
```diff
@@ -22,4 +22,9 @@ Some programs have their translations in a separate package that must be
 installed in order to use them. You can
 [search](../xbps/index.md#finding-files-and-packages) for the desired language
 (e.g. "german" or "portuguese") in the package repositories and install the
-packages relevant to the applications you use.
+packages relevant to the applications you use. An especially relevant case is
+when installing individual packages from the LibreOffice suite, such as
+`libreoffice-writer`, which require installing at least one of the
+`libreoffice-i18n-**` packages to work properly. This isn't necessary when
```
flexibeast, line 28: s/**/*/, i presume?

src/config/locales.md
```diff
@@ -22,4 +22,9 @@ Some programs have their translations in a separate package that must be
...
+installing `libreoffice` directly, since it already depends on the most common
```
flexibeast, line 29: "installing the `libreoffice` meta-package" is clearer, i think.

flexibeast: My overall preference is to not talk of meta-package X "depending on" package Y as a way of indicating that installing X will result in the installation of Y - it feels to me like it's exposing implementation information, rather than describing the outcome from the user's perspective. So in this instance, i'd prefer something like:

> installing the `libreoffice` meta-package, since doing so will install the most common translation packages.

flexibeast:
Also need to add a redirect in `book.toml`. :-)

src/config/locales.md
ericonr: Thanks!

ericonr:
I decided against changing the page name, didn't seem worth it... What do you think?

flexibeast:
@ericonr: i think it's fine to leave the title as-is, so feel free to update the PR to remove the title changes (and update the commit message accordingly).

PR #495, opened 2020-08-24
<various>: Add cover page to PDFs. 

flexibeast:


flexibeast:
PR updated: now with less Poppler, and more working links!

Vaelatern:
Why commit a binary build artifact?

flexibeast:
@Vaelatern: It's currently not a build artifact. Since the cover will only be changing far more rarely than the rest of the Handbook, i felt it was better to avoid adding an extra dependency such as `ImageMagick` or `inkscape` to the build process, and to instead have the PNG already available when building. But if the preference is to have the extra build dependency, rather than having a binary blob in the repo, then i can certainly change the PR to reflect that.

the-maldridge:
Holy blobs batman!

While its occasionally acceptable to bring in a blob, this is a 64k image for something that is mostly a static color.  This should be refactored to build on demand.

flexibeast:
@the-maldridge: Thanks, PR updated. i'm not sure why GitHub is claiming `handbook-cover.svg` is malformed; Emacs' nXML mode indicates it's fine. i just tried using validator.w3.org:

https://validator.w3.org/nu/?doc=https%3A%2F%2Fgithub.com%2Fvoid-linux%2Fvoid-docs%2Fblob%2Fb3ecc60dde06b4c068d57bd988cd1d9de217005d%2Fres%2Fhandbook-cover.svg

but that's returning 503.

CameronNemo:
You should not need Inkscape for SVG -> PNG. rsvg-convert from librsvg-utils should do the trick.

See for example: https://github.com/elementary/icons/commit/a4f35525fd0eb9e370fb89fb97e7a5d3f8794241

flexibeast:
@CameronNemo: Thanks, PR updated.

flexibeast:
i've now manually tidied the SVG: `rsvg-convert` is producing acceptable output, and GH is no longer reporting `handbook-cover.svg` as malformed. There's more tidying that could be done, but this should be enough for now.

However, the image isn't being properly displayed in the PDF output, so i still need to resolve that.

flexibeast:
Oh, that turned out to be easier than i expected. :-P Things seem to be in order now.

@ericonr / @the-maldridge: Could you please confirm that this is now okay to merge?

@CameronNemo: i just tried opening the tidied SVG in Inkscape, re-saving it to a new file, and converting that to PNG with `rsvg-convert` - the resulting PNG looks fine. Looking at the contents of the new file, there's some extra cruft, mainly in the form of a `sodipodi:namedview` element, but it doesn't look too bad.

flexibeast:
Gah, sorry, the letter version is broken. \*sigh\* Will look into it ....

flexibeast:
Fixed.

flexibeast:
[handbook-a4.pdf](https://github.com/void-linux/void-docs/files/5146207/handbook-a4.pdf)
[handbook-letter.pdf](https://github.com/void-linux/void-docs/files/5146208/handbook-letter.pdf)

flexibeast:
Using 300 DPI for the cover page:

[handbook-a4.pdf](https://github.com/void-linux/void-docs/files/5146271/handbook-a4.pdf)
[handbook-letter.pdf](https://github.com/void-linux/void-docs/files/5146273/handbook-letter.pdf)

CameronNemo:
This is the first I am actually seeing the handbook cover. Some quick notes:

* it is slightly offputting that the font is different for the cover page from the rest of the document.
* I think a section "About this handbook", restating the intent and scope for the document, would be useful. I know this information is already within the handbook, but a short and quick version on the cover might be appropriate.
* Unlike the docs site, the document does not include prominent links to the distribution home page or github organization. Those might be nice to add IMO.

flexibeast:
@CameronNemo:

Having those URLs seems like a good idea; i've now added them. i've included the `https://` prefix, but am happy to remove it if people prefer that.

Fair enough re. the cover page font. The font for the body text is currently the TeX Live default, Computer Modern. However, CM isn't a vector font, so it's unsuitable for use in SVGs. That said, Latin Modern is a vectorised version of CM, so i've updated the PR to use LM for both the cover page and the body text.

Here's what the cover page now looks like:

[cover-lm.pdf](https://github.com/void-linux/void-docs/files/5154717/cover-lm.pdf)

One problem with using LM on the cover page is getting the build to 'see' it. We don't seem to have an LM package independent of TeX Live, so we need to use LM from within TeX Live. Reading the [fc-cache(1)](https://man.voidlinux.org/fc-cache.1) man page, i would have thought i could use the `-y` flag to add `/opt/texlive/2020/texmf-dist/fonts/opentype/public/lm/` to the list of fontconfig dirs, but that doesn't work:

```
$ fc-cache -y /opt/texlive/2020/texmf-dist/fonts/opentype/public/lm/
Fontconfig error: Cannot load default config file
"/usr/share/fonts": scanning error
"/home/alexis/.local/share/fonts": scanning error
```

(Of course, this would be brittle even if it did work, since it would break when TeX Live moves to a 2021 version.)

What i've ending up doing is creating an `/etc/fonts/local.conf` file specifying the above directory, so that `fc-cache` finds it. But we obviously don't want to have to modify the user's system outside of their void-docs repo clone; what's the better way of handling this issue?

Finally, regarding an "About this Handbook" blurb, it seems to me that adding that in addition to the existing blurb would make the cover page rather crowded. Did you mean that such a blurb should *replace* the current one? At any rate, i'll work with whatever @the-maldridge thinks best.

the-maldridge:
Wow the cover became textually busy.  My 2c:
  * Drop "Enter The" - It sounds tacky and doesn't fit the rest of the language used in the handbook.
  * Keep the about blurb, it looks nice on the bottom.
  * Lose the URLs, they are to much text to put on the cover page.  If you want to have a "copyright" page immediately inside but before the TOC, that would be a good place for that information along with references to other parts of Void's official online presence.
  * Having removed most of the text around it, enlarge the logo by 30% to achieve a pleasing balance with the text below it.

flexibeast:
@the-maldridge: i've removed "ENTER THE", removed the URLs, enlarged the logo by 30%, and adjusted its position accordingly:

[cover-lm-2.pdf](https://github.com/void-linux/void-docs/files/5155489/cover-lm-2.pdf)

> If you want to have a "copyright" page immediately inside but before the TOC, that would be a good place for that information along with references to other parts of Void's official online presence.

Okay; i'd like to make that a separate issue/PR in which the contents can be discussed, so that we focus on finalising the cover page.

flexibeast:
@the-maldridge: Ping?

the-maldridge:
@flexibeast LGTM

flexibeast:
@the-maldridge: Great, thanks. :-) i've just updated the PR accordingly.

Last thing to resolve is the "making Latin Modern available to the build" issue described in [this comment](https://github.com/void-linux/void-docs/pull/495#issuecomment-684532594). Does anyone have any suggestions?

flexibeast:
\*sigh\* GH is again reporting that the SVG is invalid, and not only is nXML not reporting any errors, but from visual inspection, i can't see what the problem might be. Any suggestions?

flexibeast:
Just tried using the W3C validation service again:

https://validator.w3.org/check?uri=https%3A%2F%2Fraw.githubusercontent.com%2Fvoid-linux%2Fvoid-docs%2F997679a70d6f62c5ab996182d4033798edd645a8%2Fres%2Fhandbook-cover.svg&charset=%28detect+automatically%29&doctype=Inline&group=0

It reports:

> This document was successfully checked as SVG 1.1 + URL + HTML + MathML 3.0!

flexibeast:
Okay. PR updated. After much sturm und drang, i've come up with a kludge to make the Latin Modern font available to the build: temporarily install it in a temporary directory in the user's `~/.fonts/` directory, then run `fc-cache`.

Before coming up with this, i tried many combinations of making use of:

* the `-y` flag for `fc-cache`;
* the `--sysroot` flag for `fc-cache`;
* the `FONTDIR_PATH` env var;
* the `FONTDIR_FILE` env var;
* temporarily providing adding a `<dir>` element to the user's `fonts.conf`;

and many other things i've forgotten. Other users have faced similar difficulties with making use of fonts in a TeX distribution without making system-wide changes:

* https://unix.stackexchange.com/questions/355262/fc-cache-and-fc-list (this one is literally the same problem faced in this PR)
* https://tex.stackexchange.com/questions/517698/fontconfig-conf-file-with-texlive-on-microsoft-windows

So unless anyone has any better solutions, i think this kludge will do for now.

@ericonr: Could you please test this branch locally, and confirm that the output matches the last cover page sample i provided in this thread?

res/build.sh
```diff
@@ -51,3 +73,8 @@ pdflatex -output-directory=book/latex/ \
          -jobname=handbook-a4 \
          book/latex/handbook.tex \
          >/dev/null
+
+## Remove temporary font directory
+
+rm -rf $TMP_FONTDIR
```
ericonr, line 79: doesn't this break if the user already has their own `~/.fonts`?

res/build.sh
```diff
@@ -28,8 +28,28 @@ sed -e "s,@PREFIX@,$PREFIX," res/void-docs.in > res/void-docs
...
 
 # Build PDF
+
 echo "Building PDF"
+
+## Temporarily add OpenType version of Latin Modern to font cache.
+
+TMP_FONTDIR=$(mktemp -d -p ~/.fonts)
+mkdir -p $TMP_FONTDIR
+cp /opt/texlive/2020/texmf-dist/fonts/opentype/public/lm/* $TMP_FONTDIR
```
ericonr, line 38: Isn't this from the `texlive-bin` package? `texlive-core` seems to have them in `/usr/share/texmf-dist/fonts/opentype/public/lm/`

.gitignore
```diff
@@ -2,3 +2,4 @@ book/
 mandoc/
 res/void-docs
 res/void-docs.1
+/res/handbook-cover.png
```
ericonr, line 5: I don't think you need the first slash.

res/build.sh
flexibeast: This should only remove the temporary directory within `~/.fonts`, shouldn't it? Not the entire `~/.fonts` hierarchy. i have a `~/.fonts` directory, and the script certainly didn't trash it. The line:

    TMP_FONTDIR=$(mktemp -d -p ~/.fonts)

creates a temporary directory *within* `~/.fonts/`.

res/build.sh
flexibeast: Ah, that's much better; i'll change this. Is (relatively) new due to the recent TeX Live packaging changes? i've been using TeX Live since first installing Void, and not sure whether i just missed this existing prior to the changes.

.gitignore
flexibeast: Thanks, will fix.

flexibeast:
PR updated following @ericonr's comments.

res/build.sh
ericonr: Oh, I had the wrong idea about the semantics! It's good, then.

res/build.sh
```diff
@@ -28,8 +28,28 @@ sed -e "s,@PREFIX@,$PREFIX," res/void-docs.in > res/void-docs
 sed -e "s,@PREFIX@,$PREFIX," res/void-docs.1.in > res/void-docs.1
 
 # Build PDF
+
 echo "Building PDF"
+
+## Temporarily add OpenType version of Latin Modern to font cache.
+
+TMP_FONTDIR=$(mktemp -d -p ~/.fonts)
+mkdir -p $TMP_FONTDIR
```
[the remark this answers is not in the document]
flexibeast: Well, the point of the `-p` flag is to create intervening directories if they don't already exist.

    TMP_FONTDIR=$(mktemp -d -p ~/.fonts)
 
creates a temporary directory using `~/.fonts` as its base, such as `~/.fonts/tmp.YHReGrbetL`. Then 
    
   mkdir -p $TMP_FONTDIR

will create that path, creating the `~/.fonts` directory in the process if necessary.

Did the above ordering fail on your system?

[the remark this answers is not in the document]
ericonr: It fails if `~/.fonts` doesn't exist, because it can't create the directory. Then `mktemp` does the job of creating the directory inside `.fonts`, so there's no need to call `mkdir` afterwards.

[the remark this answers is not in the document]
flexibeast: Oh i see - i thought `mktemp` just created the directory name, but it actually creates the directory. Okay.

PR #496, opened 2020-08-26
external-applications.md: Add section on Steam

cinerea0:
This commit adds the relevant and current information from [the Steam wiki page](https://wiki.voidlinux.org/Steam) to the External Applications handbook page.

### DIFF
--- a/src/config/external-applications.md
+++ b/src/config/external-applications.md
@@ -47,6 +47,10 @@ Flatpak is another method for installing external proprietary applications on
 Linux. For information on using Flatpak with Void Linux, see the [official
 Flatpak documentation](https://flatpak.org/setup/Void%20Linux/).
 
+If sound is not working for programs installed using Flatpak,
+[PulseAudio](./media/pulseaudio.md) auto-activation might not be working
+correctly. Make sure PulseAudio is running before launching the program.
+
 Note that Flatpak's sandboxing will not necessarily protect you from any
 security and/or privacy-violating features of proprietary software.
 
@@ -82,3 +86,17 @@ example, to build the control package, you must install the `openblas-devel`,
 
 To use MATLAB's help browser, live scripts, add-on installer, and simulink,
 install the `libselinux` package.
+
+## Steam
+
+Steam can be installed either via a native package, which requires [enabling the
+"nonfree" repository](../xbps/repositories/index.md#nonfree), or via
+[Flatpak](#flatpak). The list of dependencies for different platforms and
+troubleshooting information for the native package can be found in its
+[Void-specific documentation](./package-documentation/index.html), while this
+section deals with potential issues faced by Flatpak users.
+
+If you are using a different drive to store your game library, the
+`--filesystem` option from
+[flatpak-override(1)](https://man.voidlinux.org/flatpak-override.1) can prove
+useful.


cinerea0:
@ericonr Is this something that would be considered for inclusion, or is the post-install message sufficient for documenting this?

src/config/external-applications.md
```diff
@@ -82,3 +82,61 @@ example, to build the control package, you must install the `openblas-devel`,
...
+
+### Installing Steam through XBPS
+
+To install Steam you first need to [enable the "nonfree"
+repository](../xbps/repositories/index.md###nonfree). Then, install the `steam`
+package; this should be enough for i686 systems. On x86_64 systems, additional
+32-bit libraries must be installed. These libraries are listed upon installing
+the `steam` package, and can be viewed in [Steam's install
+message](https://github.com/void-linux/void-packages/blob/master/srcpkgs/steam/INSTALL.msg).
+You will need to [enable](../config/services/index.md####enabling-services) the
```
ericonr, line 112: Kind of unrelated, but I think we could add a tiny item to the XBPS part about seeing install msgs (`xbps-query -p install-msg <package>`), and this would link there instead of to the git repo.

ericonr (commented):
I dislike having information split between places, so I'd prefer to have all of it either here or in void-packages. The information in void-packages I want to move to a `README.voidlinux` file, btw (this would also allow it to be more verbose, so some of the new text from here could go there).

@Chocimier what do you think? How should we determine what goes in void-docs and what goes in a readme file for the package?

ericonr:
After some more thinking, maybe the README could really only mention libraries and this section should mention the actual workarounds etc.

Chocimier:
> How should we determine what goes in void-docs and what goes in a readme file for the package?

This is question about scope of void-docs, will leave decision to you.

Readmes can contain anything, no idea how far from "system components" like network or printing docs should go.

cinerea0:
@ericonr Can you elaborate on what the README.voidlinux files are or will be? That'll help a lot in me having an opinion on where things should go.

ericonr:
@cinerea0 see https://github.com/void-linux/void-packages/blob/master/Manual.md#void-specific-documentation and the existing `README.voidlinux` files in void-packages :)

The idea is to leave printing things to the terminal during installation for important messages, and have actual docs available but without bumping the actually important stuff from view :)

cinerea0:
@ericonr Okay, I think I understand. In that case, how about, in addition to the current info in the steam install message being moved to a README.voidlinux file, the information specific to the native package in the changes I proposed be moved there as well. I'd still like to keep a Steam section in the docs for the Flatpak version, with a reference to the README file for people using the native package, because the information we had in the wiki that I copied here appears to be difficult to find otherwise.

ericonr:
I think that does sound reasonable, yes!

>As a side note, I was unaware of the functionality of xbps-query you mentioned earlier. Do you think an overview of the -p option might be a valuable addition to the docs?

Definitely! It's not exactly easy to find out about it :)

cinerea0:
@ericonr Alright, the proposed changes to the steam package have been merged into void-packages. Does the wording in my most recent push look okat?

src/config/external-applications.md
```diff
@@ -82,3 +82,26 @@ example, to build the control package, you must install the `openblas-devel`,
...
+[Flatpak](##flatpak). Dependency listings for different platforms and
+troubleshooting information for the native package can be found in its
+[Void-specific documentation file](./package-documentation/index.html), so the
+remainder of this section will deal with potential issues faced by Flatpak
+users.
+
+If you are using a different drive to store you game library run the following
+commmand, where `<storage>` is the mount point of your drive:
+
+`$ flatpak override com.valvesoftware.Steam --filesystem=<storage>`
```
ericonr, line 99:
```suggestion
If you are using a different drive to store you game library, the `--filesystem` option from [flatpak-override(1)](https://man.voidlinux.org/flatpak-override.1) can prove useful.
```

If you think an example command is really important, it should be a full code block, using three \`\`\`.

src/config/external-applications.md
```diff
@@ -82,3 +82,26 @@ example, to build the control package, you must install the `openblas-devel`,
...
+
+If sound is not working when Steam is installed in this manner, creating and
+running the following script to start Steam may fix the issue:
+
+```
+#!/bin/sh
+pulseaudio --start flatpak run com.valvesoftware.Steam
+```
```
ericonr, line 107:
```suggestion
If sound is not working when Steam is installed with Flatpak, [PulseAudio](link to pulse) auto-activation might not be working correctly. Make sure PulseAudio is running before launching Steam.
```

cinerea0: Do you think this advice is complete enough to get rid of the script that was included in the wiki? I'm not super familiar with pulse auto-activation, but my initial assumption is that it would be useful to keep the script (or maybe just the command) for users whose audio is not working even with pulse installed.

src/config/external-applications.md
cinerea0: No, I think your suggestion with a link to the man page is better. I'll include it in my next push.

src/config/external-applications.md
ericonr: I think if we recommend this at all, it should go in the PulseAudio section. But I think I'd rather see a recommendation for launching a terminal and running `pulseaudio --daemonize=no --exit-idle-time=-1` or something of the sort.

cinerea0: I understand. I'll see about adding something like that to the PulseAudio page.

cinerea0: Just opened a PR: #565.

cinerea0:
@ericonr Alright, I've incorporated your changes. Also, I made sure `check.sh` passed before pushing, the error from Travis is just it being unable to contact zeroconf.org at the time.

src/config/external-applications.md
```diff
@@ -82,3 +82,20 @@ example, to build the control package, you must install the `openblas-devel`,
 
 To use MATLAB's help browser, live scripts, add-on installer, and simulink,
 install the `libselinux` package.
+
+## Steam
+
+Steam can be installed either via a native package, which requires [enabling the
+"nonfree" repository](../xbps/repositories/index.md#nonfree), or via
+[Flatpak](##flatpak). The list of dependencies for different platforms and
```
ericonr, line 90:
```suggestion
[Flatpak](#flatpak). The list of dependencies for different platforms and
```

src/config/external-applications.md
```diff
@@ -82,3 +82,20 @@ example, to build the control package, you must install the `openblas-devel`,
...
+troubleshooting information for the native package can be found in its
+[Void-specific documentation](./package-documentation/index.html), while this
+section will deals with potential issues faced by Flatpak users.
```
ericonr, line 93:
```suggestion
section will deal with potential issues faced by Flatpak users.
```

src/config/external-applications.md
```diff
@@ -82,3 +82,20 @@ example, to build the control package, you must install the `openblas-devel`,
...
+
+If you are using a different drive to store you game library, the `--filesystem`
```
ericonr, line 95:
```suggestion
If you are using a different drive to store your game library, the `--filesystem`
```

src/config/external-applications.md
```diff
@@ -82,3 +82,20 @@ example, to build the control package, you must install the `openblas-devel`,
...
+option from [flatpak-override(1)](https://man.voidlinux.org/flatpak-override.1)
+can prove useful.
+
+If sound is not working when Steam is installed using Flatpak,
+[PulseAudio](./media/pulseaudio.md) auto-activation might not be working
+correctly. Make sure PulseAudio is running before launching Steam.
```
ericonr, line 101: I think we can consider moving this to the flatpak item, since other flatpak applications have the same issue. What do you think?

cinerea0:
Faling check is because it couldn't connect to libressl.org. God, travis sucks for connectivity. I ran `check.sh` locally before pushing and it didn't return any errors.

src/config/external-applications.md
cinerea0: I think it looks better in the flatpak section and as such I have moved it there in my most recent push.

ericonr:
Great, thanks!

And yes, we should look into using GH actions.

PR #500, opened 2020-08-29
arm-devices/index: Correct internal links.

flexibeast:
Reported by an IRC user. @ericonr, i presume these are the intended destinations?


ericonr:
oof, yeah! Thanks c:

I should try to enable fragment checking in linkcheck at some point.

PR #507, opened 2020-09-07
kernel: add table for kernel hooks.

ericonr:
@void-linux/pkg-committers opinions?

Not sure about the order, it's currently alphabetical by package name, but could just as well be alphabetical by hook name.


ahesford:
This table is the kind of thing that will fast become outdated. I don't think we should try to enumerate the hooks installed by Void packages.

ericonr:
> This table is the kind of thing that will fast become outdated. 

I'd prefer to do the work in keeping it up to date than lacking basic documentation for hooks. There are few enough hooks that it isn't a huge burden. The other option is including a per package readme when necessary, but those are even more likely to get out of date, imo.

hippi777:
hi folks! :)

i think it is a nice-to-have, so +1 for it from my side :)

PR #511, opened 2020-09-09
firewalls: Mention enabling nftables service.

einnhverr:
Clarifying that before running `sv up nftables` the service has to be enabled.

### DIFF
--- a/src/config/network/firewalls.md
+++ b/src/config/network/firewalls.md
@@ -78,7 +78,8 @@ This adds a core-service which restores the ruleset in `/etc/nftables.conf`.
 ### Applying the rules at runtime
 
 The `nftables` package provides the `nftables` service, which uses rules from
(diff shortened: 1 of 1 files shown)


flexibeast:
Thanks!

Duncaen:
This is wrong anyways, services linked into the runsvdir are enabled by default, `sv up` is not needed, `sv down` flushes the ruleset and keeps it flushed, something like `sv term` or hup or any other signal/command that restarts the "service" is more appropriate here.
 
In general, we have here two sections "Applying rules at boot" and "Applying rules at runtime", which are both ambiguous.
Both of them work at boot and at runtime, they are two alternative methods, with the one "Applying rules at boot" being clearly the better solution because they don't need a fake service and they make sure firewall rules are loaded before network services are up.

PR #514, opened 2020-09-11
[WIP] mdbook-gemini: Add script.

flexibeast:
Depends on https://github.com/void-linux/void-packages/pull/24799.

The current output of this script can be accessed at:

    gemini://republic.circumlunar.space/users/flexibeast/void-handbook/toc.gmi


ericonr:
I'm not sure what happened, but the files besides `toc.gmi` seem to be copies of the files in `src/`

ericonr:
And now they aren't... But the weird part is that lines are apparently using DOS newline instead of unix. Is that expected?

makeworld-the-better-one:
If I can jump in here, yes that is expected. The md2gemini code outputs DOS newlines for greater compatibility. It is likely not really necessary though. I can remove it or add a flag for it, or you guys can post-process it out if it's undesirable.

ericonr:
If the clients deal with it without issue, I don't see why we shouldn't leave as is. Thanks for answering!

ericonr:
@flexibeast `git-merge-pr` also complained locally here about some trailing whitespace in the script. I can try to pick them later.

flexibeast:
@ericonr: Updated PR to remove extraneous whitespace.

ericonr:
I believe it is now working nearly perfectly, and I'd say it's good to merge, with possible future style tweaks.

One of them is the link style:

```
=> ./about-this-handbook.gmi 1: ./about-this-handbook.gmi
=> ../xbps/index.gmi 2: ../xbps/index.gmi
=> https://man.voidlinux.org/void-docs.1 3: https://man.voidlinux.org/void-docs.1
```

I guess I'd prefer to have the link title, instead of the literal path. How do you feel about it?

ericonr:
Oh! It's also missing a note in the README about dependencies. Ideally, the perl script could fail early if it can't find `md2gemini` in `PATH`.

flexibeast:
@ericonr: PR updated:

* bail as early as possible if `md2gemini` is not present;
* use `copy` rather than `paragraph` as argument to `--links` option;
* mention `md2gemini` as dependency in README.

makeworld-the-better-one:
The README mentions md2gemini as a dependency, but I don't see that as a Void pkg.

flexibeast:
@makeworld-the-better-one: It's in the process of being packaged: https://github.com/void-linux/void-packages/pull/25050.

PR #517, opened 2020-10-02
chroot: add info for fallback UEFI location.

ericonr:
There doesn't seem to be a command line option for grub-install to
install there directly.

### DIFF
--- a/src/installation/guides/chroot.md
+++ b/src/installation/guides/chroot.md
@@ -280,11 +280,28 @@ computer's firmware when manually selecting a boot device):
 (chroot) # grub-install --target=x86_64-efi --efi-directory=/boot/efi --bootloader-id="Void"
 ```
 
-If installing onto a removable disk (such as USB), add the option `--removable`
-to the `grub-install` command.
+### Troubleshooting GRUB installation
 
 If EFI variables are not available, add the option `--no-nvram` to the
 `grub-install` command.
+
+#### Installing on removable media or non-compliant UEFI systems
+
+Unfortunately, not all systems have a fully standards compliant UEFI
+implementation. In some cases, it is necessary to "trick" the firmware into
+booting by using the default fallback location for the bootloader instead of a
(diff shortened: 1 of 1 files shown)


kwshi:
Yes, I was just about to suggest this addition to the documentation!  By the way, grub-install _does_ have a command-line option to install to the fallback location directly; it _is_ the `--removable` flag.  Per the [ArchWiki page on GRUB](https://wiki.archlinux.org/index.php/GRUB#Default/fallback_boot_path):

> Some UEFI firmwares require a bootable file at a known location before they will show UEFI NVRAM boot entries. If this is the case, grub-install will claim efibootmgr has added an entry to boot GRUB, however the entry will not show up in the VisualBIOS boot order selector. The solution is to install GRUB at the default/fallback boot path:
>
> ```
> # grub-install --target=x86_64-efi --efi-directory=esp --removable
> ```
>
> Alternatively you can move an already installed GRUB EFI executable to the default/fallback path:
>
> ```
> # mv esp/EFI/grub esp/EFI/BOOT
> # mv esp/EFI/BOOT/grubx64.efi esp/EFI/BOOT/BOOTX64.EFI
> ```

(The solution of manually copying to the fallback location, per your addition to the docs, is the other way suggested by the wiki.)

I suggest modifying the "if installing on a USB" sentence to "if installing on a USB, or on certain systems with non-standard UEFI implementations" and adding a link to the corresponding ArchWiki section.

I spent hours yesterday scratching my head out trying to figure out what was wrong with Void's UEFI installation, because no matter what I did Void would not show up in the boot entries, until I manually ran `grub-install --removable`.  This would be an immensely useful addition to the docs to anyone else encountering the same issue.

kwshi:
Ideally, there should be a change to the Void installer as well to support (optionally) installing with the `--removable` flag/performing this copy-to-fallback-location.  Otherwise, anyone facing this issue will have to manually chroot into their install and run these commands, which is not super ergonomic.

PR #519, opened 2020-10-08
graphics-drivers/intel: mention DDX conflicts

the-maldridge:
I just spent several hours trying to troubleshoot this, so I figure its worth a callout.  It wasn't immediately obvious to me that the DDX drivers and the more specific modesetting drivers don't conflict and manual intervention is needed to put the right set on a machine.

I'm still not entirely sure which classes of intel machines are affected by this, so I added it to troubleshooting rather than the driver selection section above.

### DIFF
--- a/src/config/graphical-session/graphics-drivers/intel.md
+++ b/src/config/graphical-session/graphics-drivers/intel.md
@@ -41,3 +41,8 @@ cmdline](../../kernel.md#cmdline). This problem is expected to happen on the
 Broadwell generation of internal GPUs. If you have another internal GPU and your
 issues are fixed by this kernel option, you should file a bug reporting the
 problem to kernel developers.
+
+For newer Intel chipsets, the [DDX](../xorg.md#ddx) drivers may interfere with
+correct operation. This is characterized by graphical acceleration not working
+and general graphical instability. If this is the case, try removing all
+`xf86-video-*` packages.


ericonr:
Did it fail to work even when manually selecting `modesetting` in Xorg config?

pullmoll:
FWIW I have a new Dell Latitude 3510 laptop with Intel i3-10110U CPU and UHD graphics and it was/is causing me trouble because `vlc`, `mpv`, or video players in general fail to work out of the box with the *default* video output. I'll give your suggestion a try and see if this is what's causing the problems for me too.

pullmoll:
This indeed fixed my issues as well. I removed `xbps-remove -R xorg` and installed just the subset `xbps-install -S xorg-fonts xorg-server xorg-apps xorg-input-drivers` and now everything works just fine. What a nice pitfall!

pullmoll:
Actually no, it didn't fix the problem. It's just that `vlc` did not take the change to the settings when I closed and started it again.
After a reboot the problem is back and the reason is probably this line from `Xorg.0.log`: `(WW) intel(0): Unknown chipset`.. *lol*
Anyway, this is off topic then and I'll see what to do to solve my own issue.

the-maldridge:
@ericonr as editing Xorg.conf is a deprecated configuration mechanism I try to avoid it.  In this case it was due to a non-performant driver being loaded which should simply be absent from the system.

PR #523, opened 2020-10-24
Update wpa_supplicant.md

Piping:
add details on how to use wpa_cli and wpa_supplicant based on my reading experience and real life experiment on this doc.

### DIFF
--- a/src/config/network/wpa_supplicant.md
+++ b/src/config/network/wpa_supplicant.md
@@ -3,7 +3,10 @@
 The `wpa_supplicant` package is installed by default on the base system. It
 includes utilities to configure wireless interfaces and handle wireless security
 protocols. To use wpa_supplicant, you will need to enable [the wpa_supplicant
-service](#the-wpa_supplicant-service).
+service](#the-wpa_supplicant-service). 
+
+`ln -s /etc/sv/wpa_supplicant /var/service`
+`sv start wpa_supplicant`
 
 [wpa_supplicant(8)](https://man.voidlinux.org/wpa_supplicant.8) is a daemon that
 manages wireless interfaces based on
@@ -81,6 +84,11 @@ to specify which network interface to use via the `-i` option, e.g.:
 
 ```
 # wpa_cli -i wlp2s0
+(wpa)$ add_network 0
+(wpa)$ set_network 0 ssid "wifi_name"
+(wpa)$ set_network 0 password abc123
+(wpa)$ enable_entwork 0
+(wpa)$ quit
 ```
 
 Not doing so can result in various `wpa_cli` commands (for example, `scan` and


Duncaen:
At least enabling the service is redundant, users should refer to the linked page if they don't know how services are enabled, otherwise we would have that literally on any page.
`sv start` is also not necessary as enabled services are always started unless they have a down file.

Piping:
@Duncaen does it mean once the soft link happens the runit will run the service immediately? 

As a new user who is referring to the doc during instillation, it is a bit easier to have all information needed in the front, especially for the networking setup. And I found `void-installer` cannot do this for me (wireless-setup) if I don't enable the `wpa_supplicant` service first.

Duncaen:
> @Duncaen does it mean once the soft link happens the runit will run the service immediately?

Yes it will start as soon as the service directory is scanned, which should happen within 5 or 14 seconds not sure,
in this period, sv start won't work because the service is not scanned and can't be controlled with sv anyways.

Regarding the new update, this is exactly why this should be documented in the appropriate file, now we would have to duplicate a lot more text over all pages that enable a service, which doesn't really make it better.

Piping:
@Duncaen Sure. I think the the time delay need to be mentioned . And  using `sv start` to verify the change will be more deterministic.

Duncaen:
Yes but the right place would be https://docs.voidlinux.org/config/services/index.html#enabling-services and not every single page that wants you to enable a service, linking to that page should be sufficient for new users who don't know how to enable services.

Piping:
A little bit of redundancy is not bad at all. Just referring the general enabling services section does not solve the problem of specific service, i.e I have to look up the correct name of service file/directory. Giving the right command up front would be so much easier. Especially referring to my experience when I cannot get Internet access in void installation and starting to search the Internet, find if I am missing wifi firmware etc.

Duncaen:
We've decided that redundancy is just hurting maintainability and reading experience, exactly this example was discussed and is mentioned in the style guide:

> Command code-blocks should not be used to describe routine tasks documented elsewhere in this Handbook. For example, when writing documentation for the foo package, do not provide a command code-block stating that one should install it via xbps-install foo. Similarly, do not provide code blocks describing how to enable the foo service.

https://github.com/void-linux/void-docs/blob/master/CONTRIBUTING.md#general

Piping:
@Duncaen  I don't think it hurts reading experience since it provides whatever information is needed by readers who is looking for the exact command.  But I don't want to spend time on arguing thus removing the ln command.

src/config/network/wpa_supplicant.md
```diff
@@ -3,7 +3,29 @@
 The `wpa_supplicant` package is installed by default on the base system. It
 includes utilities to configure wireless interfaces and handle wireless security
 protocols. To use wpa_supplicant, you will need to enable [the wpa_supplicant
-service](#the-wpa_supplicant-service).
+service](../services/index.md#enabling-services). 
+
+Then you need to run `rfkill` to make sure your radio device is not blocked.
+
+Run `rfkill unblock id|type` if it is shown blocked.
```
ericonr, line 10: Information about `rfkill` should be added to the general network page, because it applies to `wpa_supplicant` and IWD, as well as any other daemon we decide to add.

src/config/network/wpa_supplicant.md
```diff
@@ -3,7 +3,29 @@
...
+
+Sample Setup Commands:
+```
+# wpa_cli -i wlp2s0
+(wpa)$ add_network 0
+(wpa)$ set_network 0 ssid "wifi_name"
+(wpa)$ set_network 0 password abc123
+(wpa)$ enable_entwork 0
+(wpa)$ quit
+```
```
ericonr, line 26: Wow, their man page fails to even mention these `*_network` commands... It would be nice if we could get them added to the upstream man page.

src/config/network/wpa_supplicant.md
```diff
@@ -3,7 +3,29 @@
...
+
+Run `rfkill unblock id|type` if it is shown blocked.
+
+## Using wpa_cli
+
+`wpa_cli` is a tool comes with base system. You can use `wpa_cli` to manage `wpa_supplicant` from the command line, be sure
+to specify which network interface to use via the `-i` option. Not doing so can result in various `wpa_cli` commands (for example, `scan` and
+`scan_results`) not producing the expected output.
+
+Sample Setup Commands:
```
ericonr, line 18: "An example of the procedure to connect to an access point called `wifi_name` can be seen below:"

src/config/network/wpa_supplicant.md
```diff
@@ -3,7 +3,29 @@
...
+```
+# wpa_cli -i wlp2s0
+(wpa)$ add_network 0
+(wpa)$ set_network 0 ssid "wifi_name"
+(wpa)$ set_network 0 password abc123
+(wpa)$ enable_entwork 0
+(wpa)$ quit
+```
+
+## More on wpa_supplicant
```
ericonr, line 28: I don't love this title. I think the page could be reorganized to not require it.

src/config/network/wpa_supplicant.md
```diff
@@ -3,7 +3,29 @@
 The `wpa_supplicant` package is installed by default on the base system. It
 includes utilities to configure wireless interfaces and handle wireless security
 protocols. To use wpa_supplicant, you will need to enable [the wpa_supplicant
-service](#the-wpa_supplicant-service).
+service](../services/index.md#enabling-services).
```
ericonr, line 6: https://docs.voidlinux.org/config/network/wpa_supplicant.html#the-wpa_supplicant-service has a link to the part about enabling services...

src/config/network/wpa_supplicant.md
ericonr: It is already in, but we can perhaps expand it. https://docs.voidlinux.org/config/network/index.html

ericonr (commented):
Please remember to run the `./check` script before pushing the PR, and squash all commits that touch this file (but make a different commit if you touch `network/index.md`

ericonr:
> A little bit of redundancy is not bad at all. Just referring the general enabling services section does not solve the problem of specific service, i.e I have to look up the correct name of service file/directory.

This could be made clearer, but whenever we say "enable `service_name` service", the directory is named `service_name`.

> Especially referring to my experience when I cannot get Internet access in void installation and starting to search the Internet, find if I am missing wifi firmware etc.

New images will hopefully include the void-docs package, allowing you to access the documentation even when offline.

>  And in void-installer shell, there is not auto-completion of directory and filenames.

Is it running `dash`? We can probably add a note to the installation section about switching to `bash` before doing anything complicated.

PR #526, opened 2020-10-30
session management: add seatd

ofiala-a51:
`check.sh` currently reports HTTP 404 for some mirrors, which I believe will be solved by #520 , and for the seatd manpage, which will be fixed by adding seatd to void-packages. 

This is my first attempt at contributing to the docs, so any feedback is welcome.

### DIFF
--- a/src/config/session-management.md
+++ b/src/config/session-management.md
@@ -40,3 +40,15 @@ service, as waiting for a D-Bus activation can lead to issues.
 There is an alternative D-Bus configuration which takes advantage of elogind for
 features such as seat detection. It requires installing the `dbus-elogind`,
 `dbus-elogind-libs` and `dbus-elogind-x11` packages.
+
+## seatd
+
+[seatd(1)](https://man.voidlinux.org/seatd.1) is a minimal seat management
+daemon and an alternative to elogind.
+
+To use it, install the `seatd` package and enable its service. If you want
+non-root users to be able to access the seatd session, add them to the `_seatd`
+group.
+
+Note that, unlike elogind, seatd doesn't doesn't do anything besides managing
+seats.


src/config/session-management.md
```diff
@@ -40,3 +40,13 @@ service, as waiting for a D-Bus activation can lead to issues.
...
 features such as seat detection. It requires installing the `dbus-elogind`,
 `dbus-elogind-libs` and `dbus-elogind-x11` packages.
+
+## seatd
+
+[seatd(1)](https://man.voidlinux.org/seatd.1) is a minimal seat management
+daemon and an alternative to elogind.
+
+To use it, install the `seatd` package and enable its service. If you want
+non-root users to access the seatd session, add them to the `seatd` group.
```
ericonr, line 50: I think groups from packages should begin with an underscore, so it should be `_seatd`... This should be discussed in the PR, of course.

src/config/session-management.md
```diff
@@ -40,3 +40,13 @@ service, as waiting for a D-Bus activation can lead to issues.
...
+
+Note that unlike elogind, seatd doesn't set environmental variables for you.
```
ericonr, line 52: Not only it doesn't set them, it also doesn't handle the runtime directory at all. I think this should be in the first paragraph: "it provides some elogind features, including allowing Wayland compositors to not be setuid." - or something of the sort.

ericonr (commented):
Thank you for taking the initiative, very appreciated :)

src/config/session-management.md
ofiala-a51: Honestly this seemed obvious to me given it is a `minimal seat daemon`, not an elogind clone... which is why it didn't seem necessary to me to state this explicitly. If you feel there is a need to do so, I would alter the last paragraph to something like `Note that unlike elogind, seatd doesn't do anything besides managing seats.`. But again, this should be kinda obvious given the fact that it is described as "minimal".

src/config/session-management.md
ofiala-a51: Well, there are already several packages in the repos that provide groups without underscores, such as `chrony`, `socklog`, `weston` (which apparently doesn't delete its group after being removed from the system), and I am sure others... so if an underscore at the beginning truly is the norm, these should be fixed as well.

src/config/session-management.md
ofiala-a51: Never mind, it will be better to state this explicitly.

src/config/session-management.md
```diff
@@ -40,3 +40,14 @@ service, as waiting for a D-Bus activation can lead to issues.
 There is an alternative D-Bus configuration which takes advantage of elogind for
 features such as seat detection. It requires installing the `dbus-elogind`,
 `dbus-elogind-libs` and `dbus-elogind-x11` packages.
+
+## seatd
+
+[seatd(1)](https://man.voidlinux.org/seatd.1) is a minimal seat management
+daemon and an alternative to elogind.
```
travankor, line 47: technically, it works on top of elogind, too

ofiala-a51: @travankor Care to elaborate on that? I am pretty sure seatd doesn't depend on elogind.

travankor: https://git.sr.ht/~kennylevinsen/seatd#why-does-libseat-support-elogind

ofiala-a51: seatd != libseat (duh). Furthermore, it seems you absolute don't understand what any of those two do.

ericonr: @ofiala-a51 please be politer when making comments. Misunderstandings can happen.

travankor probably thought it would be worth mentioning libseat, and, even if they were confused, there's no reason to be rude about it.

ofiala-a51: Well, it's not a good idea to go "reviewing" a PR if you don't understand the topic. I agree I should have reacted better though. My apologies.

src/config/session-management.md
ofiala-a51: Fixed on both sides (seatd PR & here).

src/config/session-management.md
```diff
@@ -40,3 +40,14 @@ service, as waiting for a D-Bus activation can lead to issues.
...
+
+To use it, install the `seatd` package and enable its service. If you want
+non-root users to access the seatd session, add them to the `_seatd` group.
```
ericonr, line 50:
```suggestion
To use it, install the `seatd` package and enable its service. If you want
non-root users to be able to access the seatd session, add them to the `_seatd` group.
```

src/config/session-management.md
```diff
@@ -40,3 +40,14 @@ service, as waiting for a D-Bus activation can lead to issues.
...
+
+Note that unlike elogind, seatd doesn't doesn't do anything besides managing
+seats.
```
ericonr, line 53: Is `managing seats` the correct phrase for it? Perhaps "hardware access" or similar?

ericonr (commented):
Sorry for the delay!

src/config/session-management.md
ofiala-a51: Agreed, this is better :)

src/config/session-management.md
ofiala-a51: I am fairly sure it is, see for example [systemd's documentation regarding seat management](https://www.freedesktop.org/wiki/Software/systemd/multiseat/).

src/config/session-management.md
```diff
@@ -40,3 +40,15 @@ service, as waiting for a D-Bus activation can lead to issues.
...
+
+[seatd(1)](https://man.voidlinux.org/seatd.1) is a minimal seat management
+daemon and an alternative to elogind.
+
+To use it, install the `seatd` package and enable its service. If you want
+non-root users to be able to access the seatd session, add them to the `_seatd`
+group.
+
+Note that unlike elogind, seatd doesn't doesn't do anything besides managing
+seats.
```
ericonr, line 54:
```suggestion
Note that, unlike elogind, seatd doesn't doesn't do anything besides managing
seats.
```

ericonr:
Re. seat management, would you like to make a follow up PR explaning it briefly? Even if only with links to other documentation.

src/config/session-management.md
ofiala-a51: The former version looks better, that's why I used it. However, I understand that grammar is probably more important than looks :D

Fixed.

ericonr:
There was a tiny correction still needed, thanks!

ofiala-a51:
@ericonr To be honest, I don't think that's a good idea:
1. The Handbook doesn't explain **a lot** of stuff. It contains instructions on how to enable sndiod, yet it doesn't provide any information regarding what sndiod is at all. Also there is no explanation at all regarding Linux audio in the Handbook -- no explanation what a sound server is, etc. This applies to many other parts of the Handbook. And don't get me wrong, this isn't necessarily a bad thing. I had to look up a lot of stuff when installing Void for the first time, but those results gave me way more information than could be reasonably fit into the appropriate Handbook section. And now, when I look at any part of the Handbook, it's mostly straight to the point, I don't have to skip over stuff and decide what's important and what can be ignored. **TL;DR: The Handbook currently encourages people to look up details if they don't know something. Serving them this info (or direct links to the info) directly in the Handbook would prevent them from doing so, which would eventually lead to people knowing less than they do know due to not doing any additional research.**
2. My knowledge of Linux is generally very limited. I do not feel comfortable describing something so broad (and potentially complex) in a document that will be read by a lot of people that know way more than I do.

ericonr:
Honestly, I'd be okay with `multiseat` being a link to https://www.freedesktop.org/wiki/Software/systemd/multiseat/ and only that, but I see your point, though I hadn't thought of it that way :)

> My knowledge of Linux is generally very limited. I do not feel comfortable describing something so broad (and potentially complex) in a document that will be read by a lot of people that know way more than I do.

I and others could help with review and such, but it doesn't really make sense if you're uncomfortable.

For now I will just leave a mental note to revisit this if we get questions about seat management. Thanks again for taking the initiative.

PR #527, opened 2020-11-19
Mention up front the down-side of ROOTFS method

Boruch-Baum:
Mentioning the down-side up front is helpful for user to make an informed decision.


q66:
i don't see how this is not completely redundant/pointless, considering the rootfs tarballs have a date on them (not that it matters, the rootfs is fairly tiny)

ericonr:
I think this matters only for chroot installations specifically, since the installation will be naturally updated if you use the other methods. I'm not sure if I love this phrasing, though.

Regarding the part about `base-system`, it's mentioned closer to the end of article, and I think there could at least be a link to that part. And the explanation should be closer to something like "The ROOTFS tarballs aren't built with the `base-system` package, so they might lack all the components necessary for a full system The `[ROOTFS section](#tag-to-it)` explains the extra steps needeed"

ericonr:
Closing as better wording hasn't been proposed and the configuration section is right below the unpacking one.

PR #530, opened 2020-12-02
[WIP][RFC] Add a mini-guide for debugging packages and programs, to help with better issue reports

ericonr:


src/contributing/debug.md
```diff
@@ -0,0 +1,66 @@
...
+$ DEBUGINFOD_URLS="https://debugingod.s.voidlinux.org" gdb --args <program> [arguments]
+```
+
+Inside GDB, some useful commands are:
+
+- `set logging on`, which creates a `gdb.txt` file which can be shared easily;
+- `backtrace`, which shows the function calls made until the application got to
+   that place;
+- `info threads`, which shows information about threads in the program;
+- `break <function>`, which puts a breakpoint in a function that you may think
```
fosslinux, line 65: I don't think any of this is nessecary.

src/contributing/debug.md
```diff
@@ -0,0 +1,66 @@
...
+
+After this is done, check if the issue persists. It should be noted that issues
+like this can indicate hardware issues in your storage media or RAM.
+
+## Strace the program
+
+If the issue is caused by a program, you can run it under the
+[strace(1)](https://man.voidlinux.org/strace.1) utility to check if it's trying,
+for example, to access files that don't exist, and their programmer didn't
+implement a proper fallback or error message.
```
fosslinux, line 41: Could be shortened.

src/contributing/debug.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Debug the program NEEDS WORK
+
+If a look at `strace` wasn't enough, it may be possible to use the GNU debugger,
+[gdb(1)](https://man.voidlinux.org/gdb.1), to step through the program's
+execution. You can [install debug packages](../xbps/repositories/index.md) or
+use Void's [Debuginfod](https://sourceware.org/elfutils/Debuginfod.html) server.
+GDB is especially useful when an application crashes with `SIGABRT` or `SIGSEGV`
+(see [signal(7)](https://man.voidlinux.org/signal.7), since it will stop
+execution at that point and you can print a backtrace showing all the function
```
fosslinux, line 51: Don't think this is needed either

src/contributing/debug.md
```diff
@@ -0,0 +1,66 @@
...
+
+```
+# xbps-pkgdb -a
+ERROR: p7zip: hash mismatch for /usr/bin/7z.
+ERROR: p7zip: files check FAILED.
+# xbps-install -f p7zip
+```
+
+After this is done, check if the issue persists. It should be noted that issues
+like this can indicate hardware issues in your storage media or RAM.
```
fosslinux, line 34: It might, but it also might not. I feel like this would lead to a lot of people going OH NO MY RAM IS DEAD when they actually powered off the computer in the middle of an upgrade or smth.

src/contributing/debug.md
```diff
@@ -0,0 +1,66 @@
...
+Python programs, for example, complain loudly about missing modules and other
+issues like that. If you're using a compiled binary, and it's missing libraries
+for some reason, your dynamic linker should tell you as well.
+
+## Check for issues in the package database
+
+You can use the `-a` flag for
+[xbps-pkgdb(1)](https://man.voidlinux.org/xbps-pkgdb.1) to run a complete check
+on all systems packages, which can detect any files that may have been altered
+when they shouldn't have been. For any of the files listed in this step, you
```
fosslinux, line 21: files -> packages?

src/contributing/debug.md
```diff
@@ -0,0 +1,66 @@
...
+should attempt to reinstall them with the `-f` flag for
+[xbps-install(1)](https://man.voidlinux.org/xbps-install.1). An example can be
+seen below:
```
fosslinux, line 24: Shorten to `For example:`

src/contributing/debug.md
```diff
@@ -0,0 +1,66 @@
+# Troubleshooting badly behaving apps
+
+Despite the best efforts of Void Linux maintainers, it is possible that you will
+have issues with packages from the official repositories. On such cases, there
+are some steps to follow in order to provide as complete a bug report as
```
fosslinux, line 5: to follow -> you can follow

We dont require this for a bug report

src/contributing/debug.md
```diff
@@ -0,0 +1,66 @@
...
+possible.
+
+Some of these steps can even help you find the cause of the issue yourself!
+
+## Look at error messages attentively THIS NEEDS WORK
+
+Python programs, for example, complain loudly about missing modules and other
```
fosslinux, line 12: Maybe a list of different types of programs? I don't like this format.

src/contributing/debug.md
```diff
@@ -0,0 +1,66 @@
...
+use Void's [Debuginfod](https://sourceware.org/elfutils/Debuginfod.html) server.
+GDB is especially useful when an application crashes with `SIGABRT` or `SIGSEGV`
+(see [signal(7)](https://man.voidlinux.org/signal.7), since it will stop
+execution at that point and you can print a backtrace showing all the function
+calls that lead to that place. An example using the Debuginfod server is show
+below:
+
+```
+$ DEBUGINFOD_URLS="https://debugingod.s.voidlinux.org" gdb --args <program> [arguments]
+```
```
fosslinux, line 57: 1. This is not an example, its more of a 'syntax' i guess
2. I would note the debuginfod server env variable to make it work separately.

src/contributing/index.md
```diff
@@ -2,13 +2,22 @@
 
 There's more to running a distribution than just writing code.
 
-To contribute to the Void packages repository, start by reading the
+If you find an issue with a package and you manage to fix it, you can add your
+solution to the official package repository. New packages and updates are also
+welcome! To contribute to the Void packages repository, start by reading the
```
fosslinux, line 7: Wordy

src/contributing/index.md
```diff
@@ -2,13 +2,22 @@
...
 document in the void-packages GitHub repository.
 
-To contribute to this Handbook, read
-[CONTRIBUTING](https://github.com/void-linux/void-docs/blob/master/CONTRIBUTING.md)
-in the void-docs repository.
+If you are unable to fix a package yourself, make sure to [open an
+issue](https://github.com/void-linux/void-packages/issues/new) in the
+void-packages repository. This makes maintainers aware of the issue, allows
+others to fix it, and can work as a place for anyone to share workarounds in the
+meantime. Remember to check the existing issues for your problem! There are also
```
fosslinux, line 15: This sentence could be shortened to something like 'This gives the problem visibility to expedite fixing' or something like that

src/contributing/index.md
```diff
@@ -2,13 +2,22 @@
...
+some [things you can do to improve your bug report](./debug.md).
```
fosslinux, line 16: improve your bug report could be changed to something along the lines of assist us in finding the probelm

fosslinux (changes requested):

src/contributing/debug.md
ericonr: Perhaps keep only the backtrace part? Because that's what we want to get if someone says "this segfaults when I launch it".

src/contributing/debug.md
ericonr: Good point.

src/contributing/debug.md
ericonr: Heh, indeed

src/contributing/debug.md
ericonr: I'm making a second version.

ericonr:
@fosslinux thanks for the review! I have updated it.

src/contributing/index.md
ericonr: Done.

src/contributing/index.md
ericonr: I moved it around a bit, not 100% sure yet.

src/contributing/index.md
ericonr: Indeed, but I'm not sure I improved it.

PR #531, opened 2020-12-09
config/network-filesystems: remove NFS option intr

paper42:
option `intr` has been ignored since Linux 2.6.25

from `nfs(5)`:
```
       intr / nointr  This option is provided for backward compatibility.  It
                      is ignored after kernel 2.6.25.
```


ericonr:
Heh, thank you.

PR #534, opened 2020-12-19
wayland: add header for seat management.

ericonr:
Also tiny "API" to "library" change.


st3r4g:
+1 for the changes here. I would also add:
> seatd(1) is a minimal seat management daemon and an alternative to elogind **in wlroots compositors**

to https://docs.voidlinux.org/config/session-management.html?highlight=session,and,s#seatd

PR #537, opened 2020-12-24
build: make fd/pandoc optional

CameronNemo:

### DIFF
--- a/res/build.sh
+++ b/res/build.sh
@@ -1,6 +1,7 @@
 #!/bin/sh
-# uses PREFIX from environment
+# uses PREFIX and BUILD_MANPAGES from environment
 : "${PREFIX:=/usr/local}"
+: "${BUILD_MANPAGES:=1}"
 
 set -e
 PATH="$PWD/res:$PATH"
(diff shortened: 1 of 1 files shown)


res/build.sh
```diff
@@ -1,6 +1,7 @@
 #!/bin/sh
 # uses PREFIX from environment
```
ericonr, line 2: Update the comment.

res/build.sh
```diff
@@ -1,6 +1,7 @@
...
 : "${PREFIX:=/usr/local}"
+: "${VDOCS_BUILD_MANPAGES:=1}"
```
ericonr, line 4: maybe just `BUILD_MANPAGES`?

CameronNemo: done

res/build.sh
CameronNemo: to what?

ericonr: `uses PREFIX and BUILD_MANPAGES from environment`

CameronNemo: done

PR #541, opened 2021-01-08
Fix link to favicon in HTML head

paper42:
before:
```
<link rel="shortcut icon" href="">
```
after:
```
<link rel="shortcut icon" href="favicon.png">
```


ericonr:
Thanks!

PR #544, opened 2021-01-19
installation/index: Fix heading levels

msiism:

### DIFF
--- a/src/installation/index.md
+++ b/src/installation/index.md
@@ -35,7 +35,7 @@ from [other mirrors](../xbps/repositories/mirrors/index.md). Previous releases
 can be found under <https://alpha.de.repo.voidlinux.org/live/>, organized by
 date.
 
-### Verifying images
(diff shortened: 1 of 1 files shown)


src/installation/index.md
```diff
@@ -35,7 +35,7 @@ from [other mirrors](../xbps/repositories/mirrors/index.md). Previous releases
 can be found under <https://alpha.de.repo.voidlinux.org/live/>, organized by
 date.
 
-### Verifying images
+## Verifying images
```
ericonr, line 38: Why would you say image verification is not part of downloading the image?

msiism: Well, you first download an image, and then you verify it. It's a separate step. And one that needs way more explanation than downloading. So, I'd say it really deserves to be a separate section. Doing it this way also makes the layout work with fewer heading levels, which, IMO, is appropriate in this case.

PR #545, opened 2021-01-22
software-documentation: add section, embrace manpages

Chocimier:
Main reason for that change is to mention README.voidlinux.


ericonr:
Do you think the man pages section is big enough to need its own doc? I was thinking of everything in the same page, perhaps.

Chocimier:
Yes, manpages page is nicely focused and complete.

src/config/software-documentation/index.md
```diff
@@ -0,0 +1,13 @@
+# Software documentation
```
ericonr, line 1: I'm not in love with this name. Maybe `package documentation`?

src/config/software-documentation/index.md
```diff
@@ -0,0 +1,13 @@
...
+
+Most common format of documentation in Void Linux are [manual pages](./man.md).
```
ericonr, line 3:
```suggestion
The most common format of documentation in Void Linux are [manual pages](./man.md).
```

I think `is` would be more correct, but I don't know for sure.

src/config/software-documentation/index.md
```diff
@@ -0,0 +1,13 @@
...
+
+Many programs are provided along with documentation in other formats, like HTML.
+Usually it is located in `/usr/share/doc/<program>` directory.
```
ericonr, line 6:
```suggestion
Many packages contain documentation in other formats, like HTML.
This documentation can usually be found in a `/usr/share/doc/<package>` directory.
```

src/config/software-documentation/index.md
```diff
@@ -0,0 +1,13 @@
...
+
+Most extensive documentation may be split into separate `*-doc` package, like
+`julia-doc`. This is often case for programming languages, databases and big
+software libraries.
```
ericonr, line 10:
```suggestion
More extensive documentation may be split into separate `*-doc` packages, such as
`julia-doc`. This is often the case for programming languages, databases and big
software libraries.
```

src/config/software-documentation/index.md
```diff
@@ -0,0 +1,13 @@
...
+
+Software that need configuration specific to Void may include
+`/usr/share/doc/<package>/README.voidlinux` document describing it.
```
ericonr, line 13:
```suggestion
Software that requires configuration specific to Void will include
`/usr/share/doc/<package>/README.voidlinux` document describing it.
```

I changed it to `will` as kind of a promise that we *will* document all Void special cases. If you don't agree with the idea or don't think it fits, feel free to remove.

That said, I don't love this paragraph. First is how it's worded, but I haven't yet figured a better way; but also that not all READMEs will be only for void specific information, since they might just include helpful information on the initial setup for some piece of software.

src/config/software-documentation/index.md
Chocimier: Done.

src/config/software-documentation/index.md
Chocimier: > I think `is` would be more correct

Any native speaker here?

src/config/software-documentation/index.md
Chocimier: Done.

src/config/software-documentation/index.md
Chocimier: Done.

src/config/software-documentation/index.md
Chocimier: Rewritten.

src/config/package-documentation/index.md
```diff
@@ -0,0 +1,16 @@
...
+documentation can usually be found in a `/usr/share/doc/<package>` directory.
+
+More extensive documentation may be split into separate `*-doc` packages, such
+as `julia-doc`. This is often the case for programming languages, databases and
+big software libraries.
+
+In addition to documentation provided by upstream authors, packages may also
+contain some provided by Void project. This typically describes initial setup,
+or usage specific to Void. Location of it is a
+`/usr/share/doc/<package>/README.voidlinux`.
```
ericonr, line 16:
```suggestion
In addition to documentation provided by upstream authors, packages may also
contain additional documentation provided by Void Linux (or "the Void project"). This typically describes initial setup,
or usage specific to Void. It can be located in
`/usr/share/doc/<package>/README.voidlinux`.
```

Chocimier: "In addition ... additional" sounds poor.

"Void project" is already a term in docs, do we want to be that verbose here?

What about "It will" rather that "It can"?

src/config/software-documentation/index.md
ahesford: The subject is "the most common format", so "is" would be right here, but it's still a bit awkward. I suggest one of the below alternatives:

> The most common format for documentation is the `[manual page](./man.md)`.

> The most common media for documentation are `[manual pages](./man.md)`.

src/config/package-documentation/index.md
ericonr: > What about "It will" rather that "It can"?

Yes

> "Void project" is already a term in docs, do we want to be that verbose here?

It's ok to use it, but then it needs "the" before it (what I suggested in parentheses).

> "In addition ... additional" sounds poor.

Hmm, true. Maybe "In addition to documentation provided by upstream projects, packages may also
contain other documentation provided" ? (I think switching from "authors" to "upstream" should happen either way).

ericonr: Actually, I don't like "other" either.

src/config/software-documentation/index.md
Chocimier: "Media" suggestion is neat, thanks.

src/config/package-documentation/index.md
Chocimier: reworded

src/config/package-documentation/index.md
```diff
@@ -0,0 +1,16 @@
...
+Many packages contain documentation in other formats, like HTML. This
+documentation can usually be found in a `/usr/share/doc/<package>` directory.
+
+More extensive documentation may be split into separate `*-doc` packages, such
+as `julia-doc`. This is often the case for programming languages, databases and
+big software libraries.
+
+In addition to documentation provided by upstream projects, packages may also
+contain description of initial setup or usage specific to Void, provided by
+distro contributors. It will be located in
```
ericonr, line 15: My only nit is "distro" -> "distribution", which is more widespread in docs. Otherwise LGTM

Chocimier: Done.

PR #546, opened 2021-01-28
mirrors: remove maakpain.

ericonr:
This mirror can no longer be reached.

Removing https://void.webconverger.org has also been suggested (it's about 2 months out of date), but it could still be useful as a source of live images.

Before submitting a pull request, please read [CONTRIBUTING](./CONTRIBUTING.md); pull requests that do not meet the criteria described there will not be merged. Note that this repository's CONTRIBUTING contains information specific to this repository, and is not the same as CONTRIBUTING for the void-packages repository.


ericonr:
It's accessible once more :)

PR #550, opened 2021-02-14
config/sndio: remove mpv section

tibequadorian:
from [void-packages#058af7e](https://github.com/void-linux/void-packages/commit/058af7e2292d2c4947b4d4ac4bf0efb2afcb469a)
>sndio support was removed by mpv#71d218e


ericonr:
We can revert if it comes back, thanks.

PR #554, opened 2021-02-21
installation: sha256.txt -> sha256sum.txt

Gottox:
This PR changes the docs of the installation to reflect the real file names on the server. In the docs the filename that stores shasums was called `sha256.txt`, but if you look on alpha.de.repo the file is called `sha256sum.txt.


the-maldridge:
Good catch.  I noticed that the sig file was sha256sum.sig in infradocs and changed the actual sums file to match.

PR #556, opened 2021-02-28
kernel: reorder and add hardening section

q66:
This is for after https://github.com/void-linux/void-packages/pull/29142.

### DIFF
--- a/src/config/kernel.md
+++ b/src/config/kernel.md
@@ -46,6 +46,60 @@ want to remove the default kernel packages, use
 you will be able to remove the packages with
 [xbps-remove(1)](https://man.voidlinux.org/xbps-remove.1).
 
+## cmdline
+
+The kernel, the initial RAM disk (initrd) and some system programs can be
+configured at boot by kernel command line arguments. The parameters understood
+by the kernel are explained in the [kernel-parameters
+documentation](https://www.kernel.org/doc/html/latest/admin-guide/kernel-parameters.html)
+and by [bootparam(7)](https://man.voidlinux.org/bootparam.7). Parameters
+understood by dracut can be found in
+[dracut.cmdline(7)](https://man.voidlinux.org/dracut.cmdline.7).
+
+Once the system is booted, the current kernel command line parameters can be
+found in the `/proc/cmdline` file. Some system programs can change their
+behavior based on the parameters passed in the command line, which is what
+happens when [booting a different
+runsvdir](./services/index.md#booting-a-different-runsvdir), for example.
+
+There are different ways of setting these parameters, some of which are
+explained below.
+
+### GRUB
+
+Kernel command line arguments can be added through the GRUB bootloader by
+editing `/etc/default/grub`, changing the `GRUB_CMDLINE_LINUX_DEFAULT` variable
+and then running `update-grub`.
+
+### dracut
+
+Dracut can be configured to add additional command line arguments to the kernel
+through a configuration file. The documentation for dracut's configuration files
+can be found in [dracut.conf(5)](https://man.voidlinux.org/dracut.conf.5). To
+apply these changes, it is necessary to [regenerate](#kernel-hooks) the
+initramfs.
+
+## Kernel hardening
+
+Void Linux ships with some kernel security options enabled by default. This was
+originally provided by kernel command line arguments `slub_debug=P page_poison=1`,
+but since kernel series 5.3, these have been replaced with `init_on_alloc` and
+`init_on_free` (see [this commit](https://github.com/torvalds/linux/commit/6471384af)).
+
+Void's kernels come with the `init_on_alloc` option enabled by default where
+available (i.e. `linux5.4` and greater). In most cases you should usually not
+disable it, as it has a fairly minimal impact on performance (within 1%). The
+`init_on_free` option is more expensive (around 5% on average) and needs to be
+enabled manually by passing `init_on_free=1` on the kernel command line. If you
+need to disable `init_on_alloc`, you can do that similarly by passing
+`init_on_alloc=0`.
+
+There is a chance that your existing system still has the old options enabled.
+They still work in newer kernels, but have a performance impact more in line
+with `init_on_free=1`. On older hardware this can be quite noticeable. If you
+are running a kernel series older than 5.4, you can keep them (or add them)
+for extra security at the cost of speed; otherwise you should remove them.
+
 ## Kernel modules
 
 Kernel modules are typically drivers for devices or filesystems.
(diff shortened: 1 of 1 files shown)


src/config/kernel.md
```diff
@@ -46,6 +46,61 @@ want to remove the default kernel packages, use
...
+`init_on_alloc` with `init_on_alloc=0`.
+
+There is a chance that your existing system has the old options enabled. If
+they are enabled, you should remove them, unless using a pre-5.4 kernel; on
+pre-5.4 kernels you can keep them on (or even add them) if you don't mind
+worse performance (it can be significant on older hardware).
+
+Most mainstream distributions choose not to enable the hardening, while some
+now use the new options. Void recommends using a hardened kernel unless you
+have a reason not to.
```
ericonr, line 102: I would simplify this paragraph to say something to the effect of "for the most part, there's no reason to mess with these options". We don't need to mention other distros, IMO, unless you think that it is very relevant that we enable it, compared to those others.

src/config/kernel.md
```diff
@@ -46,6 +46,61 @@ want to remove the default kernel packages, use
...
+can be found in [dracut.conf(5)](https://man.voidlinux.org/dracut.conf.5). To
+apply these changes, it is necessary to [regenerate](#kernel-hooks) the
+initramfs.
+
+## Kernel hardening
+
+Void Linux ships with some kernel hardening enabled by default. This was
+originally provided by kernel command line options `slub_debug=P page_poison=1`,
+but since kernel 5.3 these have been replaced with the options `init_on_alloc`
+and `init_on_free`.
```
ericonr, line 87: I don't think "hardening" is used anywhere in docs. Maybe add a tiny little blurb explaining it or a link to some external resource?

ericonr, line 87:
```suggestion
but since kernel version 5.3 these have been replaced with the `init_on_alloc` and `init_on_free` options.
```

We can also link to the documentation explaining these two options, if it's easy to obtain a permanent link.

src/config/kernel.md
```diff
@@ -46,6 +46,61 @@ want to remove the default kernel packages, use
...
+
+Void's 5.4 and newer kernels come with `init_on_alloc` pre-enabled. You should
+generally always keep that on, as it has minimal impact on performance. The
```
ericonr, line 90:
```suggestion
Void's 5.4 and newer kernels (packages `linux5.4` and greater) come with `init_on_alloc` pre-enabled. Generally speaking,
there is no reason to disable it, as it has minimal impact on performance. The
```

"keep on" seems active, as in the user has to do something to keep it on, instead of just not messing with options.

src/config/kernel.md
```diff
@@ -46,6 +46,61 @@ want to remove the default kernel packages, use
...
+`init_on_free` option is more expensive (~5% performance impact on average)
+and needs to be enabled explicitly with `init_on_free=1`. You can also disable
+`init_on_alloc` with `init_on_alloc=0`.
+
+There is a chance that your existing system has the old options enabled. If
+they are enabled, you should remove them, unless using a pre-5.4 kernel; on
+pre-5.4 kernels you can keep them on (or even add them) if you don't mind
+worse performance (it can be significant on older hardware).
```
ericonr, line 98: The logic seems a bit complicated to follow, I tried to simplify it a bit.

```suggestion
There is a chance that your existing system has the old options enabled. If
they are enabled, and you are using a kernel newer than `linux5.3` (`linux5.4` and on),
you should remove them. For kernels before `linux5.3` (basically the `linux4.x` packages),
 you can keep them on (or even add them) if you don't mind
worse performance (it can be significant on older hardware).
```

src/config/kernel.md
```diff
@@ -46,6 +46,60 @@ want to remove the default kernel packages, use
...
+through a configuration file. The documentation for dracut's configuration files
+can be found in [dracut.conf(5)](https://man.voidlinux.org/dracut.conf.5). To
+apply these changes, it is necessary to [regenerate](#kernel-hooks) the
+initramfs.
+
+## Kernel hardening
+
+Void Linux ships with some kernel security options enabled by default. This was
+originally provided by kernel command line arguments `slub_debug=P page_poison=1`,
+but since kernel series 5.3 these have been replaced with `init_on_alloc` and
```
ericonr, line 86:
```suggestion
but since kernel series 5.3, these have been replaced with `init_on_alloc` and
```

src/config/kernel.md
```diff
@@ -46,6 +46,60 @@ want to remove the default kernel packages, use
...
+`init_on_free` (see [this commit](https://github.com/torvalds/linux/commit/6471384af)).
+
+Void's kernels come with the `init_on_alloc` option enabled by default where
+available (i.e. `linux5.4` and greater). You should usually not disable it, as
+it has a fairly minimal impact on performance (within 1%). The `init_on_free`
+option is more expensive (around 5% on average) and needs to be enabled by
+hand by passing `init_on_free=1` on kernel command line. If you have a good
+reason to disable `init_on_alloc`, you can do that similarly by passing
+`init_on_alloc=0`.
```
ericonr, line 95: I think these comments apply: https://github.com/void-linux/void-linux.github.io/pull/137#discussion_r588628892 https://github.com/void-linux/void-linux.github.io/pull/137#discussion_r588629580 and https://github.com/void-linux/void-linux.github.io/pull/137#discussion_r588628229

src/config/kernel.md
```diff
@@ -46,6 +46,60 @@ want to remove the default kernel packages, use
...
+Dracut can be configured to add additional command line arguments to the kernel
+through a configuration file. The documentation for dracut's configuration files
+can be found in [dracut.conf(5)](https://man.voidlinux.org/dracut.conf.5). To
+apply these changes, it is necessary to [regenerate](#kernel-hooks) the
+initramfs.
+
+## Kernel hardening
+
+Void Linux ships with some kernel security options enabled by default. This was
+originally provided by kernel command line arguments `slub_debug=P page_poison=1`,
```
ericonr, line 85:
```suggestion
originally provided by the kernel command line arguments `slub_debug=P page_poison=1`,
```

src/config/kernel.md
```diff
@@ -46,6 +46,60 @@ want to remove the default kernel packages, use
...
+option is more expensive (around 5% on average) and needs to be enabled by
+hand by passing `init_on_free=1` on kernel command line. If you have a good
+reason to disable `init_on_alloc`, you can do that similarly by passing
+`init_on_alloc=0`.
+
+There is a chance that your existing system still has the old options enabled.
+They still work in newer kernels, but have a performance impact more in line
+with `init_on_free=1`. On older hardware this can be quite noticeable. If you
+are running a kernel series older than 5.4, you can keep them (or add them)
+for extra security at cost of speed; otherwise you should remove them.
```
ericonr, line 101:
```suggestion
for extra security at the cost of speed; otherwise you should remove them.
```

src/config/kernel.md
q66: this is not correct, no "the" is more appropriate here

src/config/kernel.md
q66: yes, this one is legitimate

src/config/kernel.md
q66: either way is fine here

PR #560, opened 2021-03-23
about/index: remove mention of LibreSSL

oynqr:
Since the switch to OpenSSL is done, I don't think LibreSSL should be mentioned as a unique feature.


ericonr:
@oynqr if you are interested, we can add an entry about the OpenSSL switch to the history page as well :)

oynqr:
@ericonr i considered doing just that, but i wasn't sure which date to pick

ericonr:
@oynqr I would use the day `OpenSSL` was merged, you can even link the "new package" commit :)

PR #562, opened 2021-03-26
config/print/index: add section about IPP Everywhere

oynqr:
This adds a short section about IPP Everywhere printers and automatically configuring them to the docs.
Currently, this is only relevant for network printers since https://github.com/void-linux/void-packages/pull/28945 is required to make it work reliably with local printers.


ericonr:
Looking forward to the update :)

PR #566, opened 2021-04-20
Clarifying installation link.

kz6wk9:
Making clearer to the reader that the actual installation command should 
point to the correct repo for the chosen libc implementation, this helps
eliminate the possibility of a wrong system installation.


src/installation/guides/fde.md
```diff
@@ -126,11 +126,15 @@ Before we enter the chroot to finish up configuration, we do the actual install.
 keys](../../xbps/troubleshooting/common-issues.md#verifying-rsa-keys) for the
 packages you are installing.
 
+It should be noted that the next command need to be issued in accordance with
+your chosen libc implementation, if musl libc, the link should be directed to:
+`current/musl`, if GNU libc, it should be directed to: `current`.
```
ericonr, line 131: This isn't complete, aarch64 uses `current/aarch64`

See the `chroot.md` guide, if you could link to it directly from the FDE guide instead of repeating the information here, I think that'd be ideal - ideally this guide would contain only the FDE specific parts, with links to the relevant parts of the chroot one for the rest.

ericonr:
And changing the PR name doesn't fix the commit name, unfortunately ;)

kz6wk9:
> And changing the PR name doesn't fix the commit name, unfortunately ;)

Fixed!!!

And after reading `chroot.md` I found out some spots that could use some linking indeed, so I pointed the packages the user will need alongside with `base-system` and linked the rest of the procedure from `chroot.md`.
I also found to be more readable to separate the `grub` installation from the `base-system`, this way it should be more consistent with the `chroot.md`  procedure where the grub installation is only mentioned after inside the chroot.
Another point is that we would be able to avoid mentioning modifications to `/etc/default/grub` in separate sections of the `fde.md` guide as it is now, but instead do the package installation, configuration, and installation to disk all in one single section.

ericonr:
Please make these changes in separate commits (can be in this same PR, though), since they are mostly unrelated.

kz6wk9:
All done !
Please, let me know if there's anything else that could be improved!

src/installation/guides/fde.md
```diff
@@ -122,27 +122,9 @@ On a UEFI system, the EFI system partition also needs to be mounted.
...
-
-UEFI systems will have a slightly different package selection. The installation
-command for a UEFI system will be as follows.
-
-```
-# xbps-install -Sy -R https://alpha.de.repo.voidlinux.org/current -r /mnt base-system cryptsetup grub-x86_64-efi lvm2
-```
+Follow the instructions on [Base Installation](chroot.md#base-installation) to
+install the `base-system` and, additionally, the other important packages we
+need in this specific case, those are: `lvm2` and `cryptsetup`.
```
ericonr, line 127:
```suggestion
Follow the instructions for a [base installation](chroot.md#base-installation) to
install the `base-system` package together with the packages needed for this guide: `lvm2` and `cryptsetup`.
```

ericonr: This is missing the mention of GRUB UEFI.

src/installation/guides/fde.md
```diff
@@ -233,12 +195,35 @@ a new file at `/etc/dracut.conf.d/10-crypt.conf` with the following line:
 install_items+=" /boot/volume.key /etc/crypttab "
 ```
 
-Next, install the boot loader to the disk.
+Now we will configure our bootloader, see the instructions on [Installing
+GRUB](chroot.md#installing-grub) to install the correct GRUB package for your
+system, but do not install it on the disk just yet (with `grub-install`) as we
+will need to modify it's configuration before doing so.
```
ericonr, line 201: I prefer the old language... "Next, install..." just sounds better than "Now we will..."

src/installation/guides/fde.md
```diff
@@ -233,12 +195,35 @@ a new file at `/etc/dracut.conf.d/10-crypt.conf` with the following line:
...
 ```
 
+Edit the `GRUB_CMDLINE_LINUX_DEFAULT=` line in `/etc/default/grub` and add
+`rd.lvm.vg=voidvm rd.luks.uuid=<UUID>` to it. Make sure the UUID matches the one
+for the `sda1` device found in the output of the
+[blkid(8)](https://man.voidlinux.org/blkid.8) command above.
+
+Next, install the bootloader onto the disk using the correct `grub-install`
+command for your system, as stated on [Installing
+GRUB](chroot.md#installing-grub).
```
ericonr, line 225:
```suggestion
Next, install the bootloader onto the disk using the correct `grub-install`
command for your system, as explained in the ["Installing
GRUB"](chroot.md#installing-grub) section.
```

src/installation/guides/fde.md
kz6wk9: It's because, as I said before, to be more consistent with `chroot.md`, grub could be mentioned only later on as one of the last steps of the installation, and this would help by removing the mentions to modifications on ´/etc/default/grub´ on separate steps of the installation as well, by putting then all together at the end.
This could probably help further improvements of repetition removal on the `fde.md` guide in the future too, since the more consistent the flows of installation are with each other the easier it is to spot duplicated parts.

dkwo (changes requested):
May I suggest to use `cryptsetup -y -v luksFormat /dev/sda1` (with explicit verbose and verify options, as well as removing the paragraph on luks1 and grub, provided it all works now with 2.06) and 
`cryptsetup open --type luks /dev/sda1 voidvm` instead?

PR #573, opened 2021-05-16
chroot: mention aarch64.

mahiuchun:


src/installation/guides/chroot.md
```diff
@@ -168,6 +168,12 @@ Update the package manager and install `base-system`:
 # xbps-remove base-voidstrap
 ```
 
+If on aarch64, install a kernel package separately, for example:
```
ericonr, line 171:
```suggestion
If on aarch64, it will be necessary to install a kernel package separately. For example, the `linux` package can be installed in conjunction with the `base-system` one.
```

And remove the example command. That said, I'm not sure this is under the correct section? It only applies to ROOTFS installs and not the other option (there is a command to install `base-system` further above in this guide).

Maybe it makes sense to add a small aarch64 intro that mentions the need to remember a kernel when doing things with `base-system`, instead of repeating the info in multiple places?

q66: there is no "in conjunction with", base-system on aarch64 does not install a kernel, that is the whole point

ericonr: I meant that in the sense that the command would be `xbps-install base-system linux`. If you can suggest better phrasing for it, I'd appreciate it.

q66: >For example, the `linux` package can be installed in addition to `base-system`.

should be concise and unambiguous

mahiuchun: Combined the suggestions from both of you.

ericonr:
Ok, this should be receiving lint errors, could you please run the `check.sh` script locally?

CI is failing to spot it, but I'm working on fixing that.

mahiuchun:
Do you mean vmdfmt outputs "error: invalid italic or bold formatting"? It is pre-existing and should be fixed by https://github.com/void-linux/void-docs/pull/578

src/installation/guides/chroot.md
```diff
@@ -168,6 +168,9 @@ Update the package manager and install `base-system`:
 # xbps-remove base-voidstrap
 ```
 
+If on aarch64, it will be necessary to install a kernel package separately. For
+example, the `linux` package can be installed in addition to `base-system`.
+
```
ericonr, line 173: I'm not sure this is in the right section? Shouldn't it happen before the steps to install `base-system`?

mahiuchun: From my experience the kernel package could be installed before, at the same time or after `base-system` without affecting the final result much. The `xbps-reconfigure -fa` step in the end should make it work regardless?

I rewrote and moved it to right after "Base Installation". I also added mention of aarch64 in "The XBPS Method".

ericonr:
Thanks!

PR #574, opened 2021-05-18
print/index.md: Added troubleshooting steps for wireless printers

lefuglyduck:
Some wireless printers (for example the Xerox Phaser 6510) will not be able to print with the steps indicated in the Documentation because Avahi's local hostname resolution is not enabled by default. It requires installation of nss-mdns and additional file editing.

### DIFF
--- a/src/config/print/index.md
+++ b/src/config/print/index.md
@@ -99,3 +99,38 @@ The device URI can be found manually by running:
 ```
 # /usr/lib/cups/backend/usb
 ```
+
+### Wireless printer detected but unable to print
+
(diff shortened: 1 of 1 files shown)


jhonfromerov (commented):
It works for me, but without editing

``` # /etc/nsswitch.conf ```

Changing `host` line broke my system.

lefuglyduck:
Weird, I'm following the instructions here: 
https://github.com/avahi/nss-mdns?tab=readme-ov-file#activation
My particular printer does not work without editing that file.

PR #576, opened 2021-05-19
SSD.md

lefuglyduck:
Added instructions on how to TRIM on system start or shutdown using rc.local or rc.shutdown.


lefuglyduck:
I didn't add any bold or italics, I can't figure out why this check failed.

paper42:
I think you should wrap lines to 80 characters, the CI shows you a patch for it.

PR #577, opened 2021-05-19
ssd.md: Implemented fstrim in a more general manner. Added instructions.

lefuglyduck:
Also implemented fstrim more generally. Added instructions on how to TRIM on system start or shutdown using rc.local or rc.shutdown.

### DIFF
--- a/src/config/ssd.md
+++ b/src/config/ssd.md
@@ -13,15 +13,18 @@ solid state drive partition does not show TRIM support, please verify that you
 chose a file system with TRIM support (ext4, Btrfs, F2FS, etc.). Note that F2FS
 requires kernel 4.19 or above to support TRIM.
 
-To run TRIM one-shot, you can run
-[`fstrim(8)`](https://man.voidlinux.org/fstrim.8) manually. For example, if your
-/ directory is on an SSD:
+To run TRIM once, you can run [fstrim(8)](https://man.voidlinux.org/fstrim.8)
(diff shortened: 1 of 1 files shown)


src/config/ssd.md
```diff
@@ -13,15 +13,18 @@ solid state drive partition does not show TRIM support, please verify that you
...
+To run TRIM once, you can run [fstrim(8)](https://man.voidlinux.org/fstrim.8)
+manually. For example, to trim all mounted filesystems mentioned in `/etc/fstab`
+on devices that support the discard operation:
 
 ```
-# fstrim /
+# fstrim --fstab
 ```
 
 To automate running TRIM, use cron or add the `discard` option to `/etc/fstab`.
```
ericonr, line 24: You should mention the new method in this list: "To automate running TRIM, use cron, run `fstrim` at boot or shutdown time, or add the `discard` option to `/etc/fstab`.

src/config/ssd.md
```diff
@@ -13,15 +13,18 @@ solid state drive partition does not show TRIM support, please verify that you
...
+Note: using `discard` is not recommended. The `discard` method might cause the
+system to slow down because it forces the system to apply TRIM instantly on
+every individual file deletion.
```
ericonr, line 27: I think this downside should be mentioned in the `fstab` section. Maybe use "add the appropriate mount option to `/etc/fstab`" above so people don't just use the option and move on without reading everything.

PR #580, opened 2021-05-22
contributing/index.md: Add port requirement for PopCorn

cinerea0:
I just saw the open port requirement for PopCorn on IRC, and since I didn't see it documented anywhere I thought it would be good to add.


ericonr:
Neat, thanks :)

PR #583, opened 2021-06-05
Remove mirror mirrors.hushan.tech.

Wesley-Chan:
Hi.
As of the time I put up a mirror for Void Linux, download speed was miserable and there wasn't any mirror in China. But things have improved since then. Now three universities have made their contribution in hosting Void Linux for users in China. It looks like there's no need of hosting it on my home fiber anymore.

Thanks to Void developers for your great project!


abenson:
Thanks for the mirror when it was needed!

PR #587, opened 2021-07-04
Make explicit that being in the audio group is required for pipewire-pulse.

kjkuan:
In addition to the `XDG_RUNTIME_DIR`, I've found that with a clean base install of voidlinux, if I follow the handbook to add audio support by choosing the pipewire-puse route, I need to also add my user to the `audio` group, otherwise, `pactl info` would show pipewire is being used and everything appears to be correct, but just no sound.

### DIFF
--- a/src/config/media/pipewire.md
+++ b/src/config/media/pipewire.md
@@ -69,4 +69,5 @@ on glibc-based systems:
 
 The Pulseaudio replacement requires the
 [`XDG_RUNTIME_DIR`](../session-management.html#xdg_runtime_dir) environment
-variable to work correctly.
+variable and the user being in the [`audio`](../users-and-groups.md) group to
+work correctly.


ericonr:
That's interesting, I wonder if it needs `elogind` and/or `polkit` otherwise? I'm not in `audio` group and things work ok.

kjkuan:
oh? weird. I'll double check, but I don't have `elogind` nor `polkit` installed.

kjkuan:
I can confirm that if I remove myself from the `audio` group, I get no sound with `pipewire` + `pipewire-pulse`; however, if I add myself back with the `audio` group, I have sound. Also, I don't have `elogind` nor `polkit` installed.

ericonr:
Ok, so I think we should document the requirement with the note that having elogind makes it unnecessary.

st3r4g:
> I need to also add my user to the audio group, otherwise, pactl info would show pipewire is being used and everything appears to be correct, but just no sound.

But isn't this true also for e.g. `alsamixer`? regular users have no permissions to access `/dev/snd/hw*` by default.

ericonr:
Yup, that's very much the case. I recently tried out a session without elogind, and pipewire didn't have devices until I added myself to audio group.

I think this is noted in the page about groups, a more specific call out here might be worth it.

PR #596, opened 2021-08-07
update mirror change instruction.md

d33vliter:
in the instructions of the documentation it appears that I must put this command:

```
# sed -i 's|https://alpha.de.repo.voidlinux.org|<repository>|g' /etc/xbps.d/*-repository-*.conf
```

but it doesn't work(doesn't add changes)

i asked in the telegram group and it turned out that it worked like this:

```
# sed -i 's|<repository>|https://alpha.de.repo.voidlinux.org|g' /etc/xbps.d/*-repository-*.conf
```

leaving the instructions like this:

```
1 # mkdir -p /etc/xbps.d
2 # cp /usr/share/xbps.d/*-repository-*.conf /etc/xbps.d/
3 # sed -i 's|<repository>|https://alpha.de.repo.voidlinux.org|g' /etc/xbps.d/*-repository-*.conf
```

### DIFF
--- a/src/xbps/repositories/mirrors/changing.md
+++ b/src/xbps/repositories/mirrors/changing.md
@@ -11,7 +11,7 @@ To modify mirror URLs cleanly, copy all the repository configuration files to
 ```
 # mkdir -p /etc/xbps.d
 # cp /usr/share/xbps.d/*-repository-*.conf /etc/xbps.d/
-# sed -i 's|https://alpha.de.repo.voidlinux.org|<repository>|g' /etc/xbps.d/*-repository-*.conf
+# sed -i 's|<repository>|https://alpha.us.repo.voidlinux.org|g' /etc/xbps.d/*-repository-*.conf
 ```
 
 After changing the URLs, you must synchronize xbps with the new mirrors:


paper42:
This page assumes you are using https://alpha.de.repo.voidlinux.org and want to switch to a different mirror. I think that in your case, it was the other way around - you switched from a different mirror to https://alpha.de.repo.voidlinux.org, that's why the syntax is the other way around for you.

Could you confirm this is the case? If yes, we could document it in this PR instead of the fix you proposed.

sed syntax:
```
sed -i 's|<oldstring>|<newstring>|g' <files>
```

in our case:
```
sed -i 's|<oldrepo>|<newrepo>|g' <configs>
```

d33vliter:
> This page assumes you are using https://alpha.de.repo.voidlinux.org and want to switch to a different mirror. I think that in your case, it was the other way around - you switched from a different mirror to https://alpha.de.repo.voidlinux.org, that's why the syntax is the other way around for you.
> 
> Could you confirm this is the case? If yes, we could document it in this PR instead of the fix you proposed.
> 
> sed syntax:
> 
> ```
> sed -i 's|<oldstring>|<newstring>|g' <files>
> ```
> 
> in our case:
> 
> ```
> sed -i 's|<oldrepo>|<newrepo>|g' <configs>
> ```

Exactly, that's just it. I'm stupid, the documentation is fine hahaha, thank you very much indeed.

PR #608, opened 2021-10-18
installation: update signature checking information

ericonr:
- we are now using minisign for improved compatibility
- minisign-signed files means 'signify -C' doesn't work anymore, so
  remove that guidance; we can still use 'sha256 -C' if available,
  though
- minisign workaround didn't work, and isn't necessary any more

### DIFF
--- a/src/installation/index.md
+++ b/src/installation/index.md
@@ -67,9 +67,10 @@ on the image to ensure they haven't been tampered with.
 Current images are signed using a signify key that is specific to the release.
 If you're on Void already, you can obtain the keys from the `void-release-keys`
 package, which will be downloaded using your existing XBPS trust relationship
-with your mirror. You will also need a copy of
-[signify(1)](https://man.voidlinux.org/signify.1); on Void this is provided by
-the `outils` package.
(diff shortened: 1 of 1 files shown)


src/installation/index.md
```diff
@@ -81,37 +82,55 @@ than Void Linux:
...
 
-The same warning as above applies. If the verification process isn't successful,
-do not use the file - warn the Void Linux team about it.
+Alternatively, if the `sha256` utility isn't available to you, you can compute
+the SHA256 hash of the file, e.g. using
+[sha256sum(1)](https://man.voidlinux.org/sha256sum.1), and compare it to the
+value contained in `sha256sum.txt`:
+
+```
+$ sha256sum void-live-x86_64-20210930.iso
```
Lokawn, line 128: from the [manpage](https://man.voidlinux.org/sha256sum.1), sha256sum supports BSD style checksum and directly `sha256sum -c --ignore-missing sha256sum.txt` can be used.

ericonr: Oh, that's nice :)
Would you mind opening a PR with that suggestion?

PR #610, opened 2021-11-03
musl.md: replace proot mention with bubblewrap

balejk:
Recently I needed to run a glibc dependent program on musl, preferably without having root privileges, and I had difficulties getting proot to work properly. It would have been useful to know about xbps-uunshare, because that worked out of the box, plus it is usually preinstalled on Void and no additional packages need to be installed.

### DIFF
--- a/src/installation/musl.md
+++ b/src/installation/musl.md
@@ -48,9 +48,32 @@ Use [chroot(1)](https://man.voidlinux.org/chroot.1) to change to the new root,
 then run glibc programs as usual. Once you've finished using it, unmount the
 chroot using [umount(8)](https://man.voidlinux.org/umount.8).
 
-#### PRoot
+#### Bubblewrap
 
-An alternative to the above is [proot(1)](https://man.voidlinux.org/proot.1), a
-user-space implementation of chroot, mount --bind, and binfmt_misc. By
-installing the `proot` package, unprivileged users can utilize a chroot
-environment.
+An alternative for unprivileged users is
+[bwrap(1)](https://man.voidlinux.org/bwrap.1), which uses Linux namespaces. It
+can be installed via the `bubblewrap` package. An example illustrating how to
+use `bwrap` to launch Tor Browser follows.
+
+Create a glibc Void Linux container with the required software:
+
+```
+$ mkdir container
+$ XBPS_TARGET_ARCH=$(xbps-uhelper arch | sed 's/-musl$//') xbps-install -S -r container -R https://repo-default.voidlinux.org/current/ base-voidstrap torbrowser-launcher libXt dejavu-fonts-ttf
+```
+
+(you will have to verify and confirm the repository RSA key before proceeding;
+the `dejavu-fonts-ttf` package is required in order for the launcher to display
+correctly). Uncomment desired locales in `container/etc/default/libc-locales`
+and generate them by issuing:
+
+```
+$ bwrap --bind container / --dev /dev --proc /proc --tmpfs /tmp xbps-reconfigure glibc-locales ca-certificates
+```
+
+which will also configure the `ca-certificates` package required by Tor Browser.
+Launch `torbrowser-launcher`:
+
+```
+$ bwrap --bind container / --dev /dev --proc /proc --tmpfs /tmp --ro-bind ~/.Xauthority ~/.Xauthority --ro-bind /etc/resolv.conf /etc/resolv.conf --ro-bind /tmp/.X11-unix/X${DISPLAY:1} /tmp/.X11-unix/X${DISPLAY:1} --setenv DISPLAY $DISPLAY --unshare-all --share-net torbrowser-launcher
+```


Vaelatern (approved):

Duncaen:
We should not suggest proot at all for that, because it is that cursed and mostly broken and not required on a void system since we have user namespaces enabled.

I don't like to suggest `xbps-uunshare`, its made for xbps-src and not a general namespace/container tool, I would prefer to only suggest bubblewrap or similar programs.

ericonr:
I think the only advantage `proot` brings is that it allows you to fake chroot into a root owned dir. I don't think you can do that with `bwrap` or `xbps-uunshare`, might be wrong though.

balejk:
> I don't like to suggest `xbps-uunshare`, its made for xbps-src and not a general namespace/container tool, I would prefer to only suggest bubblewrap or similar programs.

I wasn't aware of this. bubblewrap seems to work for me, shall I instead of mentioning xbps-uunshare just replace proot with bubblewrap then?

ericonr:
I think that would be preferable! Thanks c:

Vaelatern (approved):

src/installation/musl.md
```diff
@@ -48,9 +48,9 @@ Use [chroot(1)](https://man.voidlinux.org/chroot.1) to change to the new root,
 then run glibc programs as usual. Once you've finished using it, unmount the
 chroot using [umount(8)](https://man.voidlinux.org/umount.8).
 
-#### PRoot
+#### bubblewrap
```
ericonr, line 51:
```suggestion
#### Bubblewrap
```

src/installation/musl.md
```diff
@@ -48,9 +48,9 @@ Use [chroot(1)](https://man.voidlinux.org/chroot.1) to change to the new root,
...
 
-An alternative to the above is [proot(1)](https://man.voidlinux.org/proot.1), a
-user-space implementation of chroot, mount --bind, and binfmt_misc. By
-installing the `proot` package, unprivileged users can utilize a chroot
-environment.
+An alternative for unprivileged users is the
+[bwrap(1)](https://man.voidlinux.org/bwrap) command which uses Linux namespaces
+to run applications in a sandbox container. It can be installed via the
+`bubblewrap` package.
```
ericonr, line 56:
```suggestion
An alternative for unprivileged users is
[bwrap(1)](https://man.voidlinux.org/bwrap.1), which uses Linux namespaces. It can be installed via the
`bubblewrap` package. An example invocation is shown below:
```

Then include an example of how to use it, since it's not always obvious, IMO.

src/installation/musl.md
balejk: From the `bubblewrap` [README](https://github.com/containers/bubblewrap/blob/master/README.md) it was not clear to me whether the official name is stylized with capital or lower "B" and as neither `musl`, nor `glibc chroot` section titles are capitalized, I chose the lowercase variant which seemed more likely based on the README. However, the README uses capital "B" in the title so indeed let's use that.

src/installation/musl.md
balejk: I omitted an example invocation because from [CONTRIBUTING](https://github.com/void-linux/void-docs/blob/master/CONTRIBUTING.md) I inferred that it's preferred that readers study the manual page and decide for themselves which options will best suit their needs. Anyway, if you have any experience with this utility, please propose some full example, as all I needed to do in my case was
```bash
$ bwrap --bind path/to/chroot / command
```
which seems too basic.

ericonr: I think in this case the command is helpful, since what's desired from the tool isn't immediately obvious. I'd suggest `--proc /proc` and `--dev /dev`, at least.

balejk: Actually, I'm currently unable to get a container set up using
```bash
$ XBPS_TARGET_ARCH=x86_64 xbps-install -S -r /tmp/chroot -R https://... base-voidstrap
```
to work properly - I have used
```bash
$ bwrap --bind /tmp/chroot / --uid 0 --gid 0 --dev  /dev --proc /proc --ro-bind /etc/resolv.conf /etc/resolv.conf sh
```
to chroot into it, but when I try to `ping voidlinux.org`, I get `ping: socket: Operation not permitted`. What am I missing?

balejk: Nevermind, apparently it's because of capabilities.

ericonr:
Ping?

balejk:
Sorry, I'm just busy lately, but I didn't forget.

My last effort was to use `bubblewrap` to launch `torbrowser-launcher` on musl, but I haven't been successful. So I'd first like to be able to do that and then use that as an example. Any hints are appreciated.

ericonr:
Ok, just checking!

Can you share errors you got, if any? Given that torbrowser will probably use namespaces on its own, there might be some confusion.

balejk:
Currently I'm getting
```python
Tor Browser Launcher
By Micah Lee, licensed under MIT
version 0.3.5
https://github.com/micahflee/torbrowser-launcher
Traceback (most recent call last):
  File "/bin/torbrowser-launcher", line 30, in <module>
    torbrowser_launcher.main()
  File "/usr/lib/python3.10/site-packages/torbrowser_launcher/__init__.py", line 76, in main
    common = Common(tor_browser_launcher_version)
  File "/usr/lib/python3.10/site-packages/torbrowser_launcher/common.py", line 71, in __init__
    self.init_gnupg()
  File "/usr/lib/python3.10/site-packages/torbrowser_launcher/common.py", line 282, in init_gnupg
    self.import_keys()
  File "/usr/lib/python3.10/site-packages/torbrowser_launcher/common.py", line 367, in import_keys
    imported = self.import_key_and_check_status(key)
  File "/usr/lib/python3.10/site-packages/torbrowser_launcher/common.py", line 337, in import_key_and_check_status
    with gpg.Context() as c:
  File "/usr/lib/python3.10/site-packages/gpg/core.py", line 206, in __init__
    self.protocol = protocol
  File "/usr/lib/python3.10/site-packages/gpg/core.py", line 161, in __setattr__
    super(GpgmeWrapper, self).__setattr__(key, value)
  File "/usr/lib/python3.10/site-packages/gpg/core.py", line 885, in protocol
    errorcheck(gpgme.gpgme_engine_check_version(value))
  File "/usr/lib/python3.10/site-packages/gpg/errors.py", line 120, in errorcheck
    raise GPGMEError(retval, extradata)
gpg.errors.GPGMEError: GPGME: Invalid crypto engine
```
`gnupg`, `gpgme` and `libgpgme` are all installed in the container.

balejk:
@ericonr I believe this is ready for another review.

balejk:
@classabbyamp I see you are working on some changes rendering this pull request rather obsolete, can I close it then?

classabbyamp:
up to you, I haven't been able to test what i wrote fully yet

PR #612, opened 2021-11-04
wayland: add SDL environment variable.

mahiuchun:


ericonr:
What did you test with this? I've had quite a few broken applications when using it, but I tested this a year ago at least.

q66:
works fine for me (with xonotic and a bunch of other SDL shooters)

nowadays you even get a window decoration :p

q66:
but also it's not required per se, because it will go through xwayland and still work fine without (unless you disable that)

PR #615, opened 2021-11-14
Update index.md

subsetpark:
Printers can be discovered, but won't accept jobs, unless nss is installed as well.


ericonr:
Can you reword your commit title, please? It should be `config/print: add nss dependency.`

Line wrapping the commit body would also be appreciated.

That said, this is an unfortunate situation, since it means musl users can't print :p

Vaelatern:
698a6e9f2af90499e1c3df43789a63b74f13f0e5

Issue #620, opened 2021-11-26
Full Disk Encryption: Misleading note on GRUB support for LUKS2


ryanm0:
Currently, the void-docs page for full-disk encryption states:
> [Cryptsetup](https://man.voidlinux.org/cryptsetup.8) defaults to LUKS2, yet GRUB releases before 2.06 only had support for LUKS1. Therefore, it might make sense to force LUKS1 if you wish to achieve better compatibility.

GRUB 2.06 did add support for LUKS2 encryption, however cryptsetup's default key derivation function for LUKS2 (Argon2) does not have support as of GRUB 2.06 (see [the cryptomount section of the GRUB documentation](https://www.gnu.org/software/grub/manual/grub/grub.html#cryptomount)). The current wording seems to imply that LUKS2 will work by default, but you'd (theoretically[^1]) need to manually select PBKDF2 instead of Argon2 (with `--pbkdf pbkdf2`) if using LUKS2 with GRUB. 

[^1]: I say theoretically here because the [Arch Wiki's entry on GRUB](https://wiki.archlinux.org/title/GRUB#LUKS2) claims that, in addition to GRUB not supporting Argon2, grub-install also doesn't support creating images for LUKS2. I haven't tested to see if this is actually the case even when selecting PBKDF2 (I gave up and just went with LUKS1 after LUKS2 with Argon2 didn't work).

ahesford:
"GRUB releases before 2.06 only had support for LUKS1" does not imply that "LUKS 2 will work by default" on later releases. However, if you believe rewording th text would improve clarity, please PR some revisions.

camoz:
It does not imply "LUKS 2 will work by default on later releases", but it heavily suggests it IMO. The PR looks good to me, FWIW, but I'm not a void dev.

PR #621, opened 2021-11-26
guides/fde: clarify note on LUKS2 support in GRUB

ryanm0:
See issue #620. Let me know if anything should be revised.

### DIFF
--- a/src/installation/guides/fde.md
+++ b/src/installation/guides/fde.md
@@ -45,9 +45,10 @@ Device      Start       End   Sectors  Size Type
 
 ## Encrypted volume configuration
 
-[Cryptsetup](https://man.voidlinux.org/cryptsetup.8) defaults to LUKS2, yet GRUB
-releases before 2.06 only had support for LUKS1. Therefore, it might make sense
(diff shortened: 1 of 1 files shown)


src/installation/guides/fde.md
```diff
@@ -45,9 +45,10 @@ Device      Start       End   Sectors  Size Type
 
 ## Encrypted volume configuration
 
-[Cryptsetup](https://man.voidlinux.org/cryptsetup.8) defaults to LUKS2, yet GRUB
-releases before 2.06 only had support for LUKS1. Therefore, it might make sense
-to force LUKS1 if you wish to achieve better compatibility.
+GRUB 2.06 added partial support for LUKS2, however LUKS2 will not work with GRUB
+by default. See the [Arch Wiki](https://wiki.archlinux.org/title/GRUB#LUKS2) for
```
paper42, line 49: We should mention that only PBKDF2 is supported and it won't work with the default Argon2i instead of linking to a whole wiki page. It's not that complicated.

mhmdanas:
Now that https://github.com/void-linux/void-docs/pull/705 has been merged, this should be closed.

PR #622, opened 2021-11-26
guides/fde: warn about XTS

curiousjunior:
if you don't understand the importance of encryption with authentication, please learn some basics about cryptography, I recommend reading [crypto 101](https://www.crypto101.io/) book

### DIFF
--- a/src/installation/guides/fde.md
+++ b/src/installation/guides/fde.md
@@ -3,6 +3,13 @@
 **Warning**: Your drive's block device and other information may be different,
 so make sure it is correct.
 
+## Caution
+
+Filesystem-level encryption with AEAD is more reliable than using
+[XTS](https://sockpuppet.org/blog/2014/04/30/you-dont-want-xts/),
+you need to adapt the guide from Alpine Linux wiki for setting up
+[encrypted root on ZFS](https://wiki.alpinelinux.org/wiki/Root_on_ZFS_with_native_encryption)
+
 ## Partitioning
 
 Boot a live image and login.


src/installation/guides/fde.md
```diff
@@ -3,6 +3,13 @@
 **Warning**: Your drive's block device and other information may be different,
 so make sure it is correct.
 
+## Caution
+
+Filesystem-level encryption with AEAD is more reliable than using
+[XTS](https://sockpuppet.org/blog/2014/04/30/you-dont-want-xts/),
```
ericonr, line 9: The article you linked mentions:

> Be wary of applications that claim to use it for anything other than disk encryption.

We are using it for disk encryption.

> If you’re an end-user looking for crypto advice: use Truecrypt, use Filevault, use dm-crypt.

LUKS uses `dm-crypt` as its backend, so we are also doing things correctly here.

src/installation/guides/fde.md
```diff
@@ -3,6 +3,13 @@
...
+you need to adapt the guide from Alpine Linux wiki for setting up
+[encrypted root on ZFS](https://wiki.alpinelinux.org/wiki/Root_on_ZFS_with_native_encryption)
```
ericonr, line 11: I'd rather not point to ZFS on Alpine when https://github.com/zbm-dev/zfsbootmenu/ exists, is a Void first solution, and works superbly well.

abenson:
I agree with @ericonr's assessments.  I don't think these changes warrant their merging.

paper42:
While this PR was already rejected on technical grounds, I would like to point out that the tone of this PR is not acceptable. Do better next time.

curiousjunior:
Dear Void Community,

Please be careful with XTS, nowadays we have options to do authenticated filesystem-level encryption that is superior in terms of both speed and reliability, so recommendation to use `dm-crypt` and friends is an outdated practice, cryptography constantly evolves and there are many popular algorithms and systems that are vulnerable due to increasing amount of kinds of attacks, so they have modern replacements that address known issues.

Please forgive me for pointing to Alpine wiki, I should've carefully written Void-specific documentation on setting up ZFS on root with native encryption instead.

Sincerely, Leo

Vaelatern:
Can you please clarify how the source you shared agrees with your recommendation above that we avoid `dm-crypt` in general?

PR #625, opened 2021-11-29
media: add note about audio group.

ericonr:
elogind will use posix attributes to give logged in users access to
devices tagged as uaccess (such as audio and joystick input devices).
Without it, it's necessary to be in the audio group to access them.

Closes #587

@st3r4g thoughts?


st3r4g:
Yes I think this is correct and is good to mention.

PR #626, opened 2021-11-29
session-management: mention dbus-update-activation-environment

ericonr:
Necessary, for example, to allow gnome-keyring-daemon to launch prompts
for user interaction, since the prompts are launched by the dbus daemon.


ericonr:
@steinex iirc we talked about this when I mentioned my service that had a `sleep 4` in the middle :p 

Does it look ok?

steinex:
I think it's not actionable enough - imo audience should be told to run it as part of their (sway|X11 wm) config so that it's called after their Wayland/X session is up.

Then for sway, you also want `XDG_CURRENT_DESKTOP=sway` (but I forgot why and i'm not sure if is important to mention / in scope, iirc it was for portal-wlr).

ericonr:
Thanks, I will try to include some guidance on when to run it.

Portal-wlr does need the current desktop var, I will try to sneak in something about "other applicable variables" (though for my usage I just launch dbus-run-session with that one set already).

PR #628, opened 2021-12-03
Update checkout

tiziodcaio:
Is a dumb update... I only suggest it no worries... 😉

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -13,5 +13,5 @@ jobs:
           xbps-install -Syu || xbps-install -yu xbps
           xbps-install -yu
           xbps-install -y mdbook-linkcheck vmdfmt git
-      - uses: actions/checkout@v1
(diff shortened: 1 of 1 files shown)


abenson:
Change the commit message to be a little more explanatory.

ericonr:
I'm letting it run, but pretty sure v2 doesn't work because it uses glibc linked nodejs.

ericonr:
```
  /usr/bin/docker exec  3e41505112e32180bbbf07d26432ab8c194f0d3ad9599e8d0bb55b04725e586d sh -c "cat /etc/*release | grep ^ID"
OCI runtime exec failed: exec failed: container_linux.go:380: starting container process caused: no such file or directory: unknown
```

or it fails even early D:

tiziodcaio:
> I'm letting it run, but pretty sure v2 doesn't work because it uses glibc linked nodejs.

Ok sure, I will close the pull if is it

PR #632, opened 2021-12-22
src/config/services/index.md: Improve wording

camoz:
AIUI, this bullet point refers to the clean "environment" of a process, and not to the concept of process state (e.g. running/failed/degraded/...).


camoz:
Please review carefully, as I'm quite new to `runit`.

Also note that after my changes it says "[the service] will be started with the same environment **variables**". I'm not 100% sure this is correct. But if it is, this wording is more precise IMO.

ericonr:
Looks good, but please fix the commit message to be the same as the PR title.

`git commit --amend` should allow you to fix the commit message, and then just do `git push -f`

camoz:
Done.

PR #636, opened 2021-12-24
remove https://mirror.erickochen.nl/voidlinux/ mirror, which is down

bobertlo:
This link was dead a couple weeks ago, and it is dead again (still?) now.


abenson:
@eirikrcoquere Any chance of this coming back?

eirikrcoquere:
@abenson the mirror will not return.

PR #638, opened 2021-12-26
users-and-groups.md: network group is still used

camoz:
network group is still used according to https://github.com/void-linux/void-packages/pull/34703#issuecomment-1001186315


Duncaen:
I think this might require different wording, "Access to input devices" is different, as this groups provides access to the actual input device files, while with the network group it some network managers use it as access control to control the network manager.

camoz:
I'd be happy to change it, I'm all for precise and descriptive wording. :)

But I don't have a good description, I also lack understanding of how exactly the group is used. So I'll just do nothing until someone proposes something better or it gets merged...

bobertlo:
Would "access to network management interfaces" be more accurate maybe? Should it be qualified, like 'some' network management interfaces?

ethamck:
Closed by https://github.com/void-linux/void-docs/pull/794

(@classabbyamp)

PR #640, opened 2021-12-27
src/xbps/index.md: Mention `xpkg` to list installed packages

camoz:
Reasons for mentioning this:

- using `xpkg` is much simpler: one command without arguments vs. 3 commands piped together
- xpkg is much faster: 0.13 s versus 3.28 s (on a C2D laptop with SSD from 2011)

### DIFF
--- a/src/xbps/index.md
+++ b/src/xbps/index.md
@@ -120,7 +120,14 @@ This requires `xbps-query` to download parts of every package to find the file.
 `xlocate`, however, queries a locally cached index of all files, so no network
 access is required.
 
-To get a list of all installed packages, without their version:
+To get a list of all installed packages without their version, use `xpkg` from
+the `xtools` package:
+
+```
+$ xpkg
+```
+
+Alternatively, use:
 
 ```
 $ xbps-query -l | awk '{ print $2 }' | xargs -n1 xbps-uhelper getpkgname


camoz:
I forgot to mention, I have 255 packages installed on this test system, and it's in a VM.

```
[root@void ~]# time xpkg > /dev/null

real	0m0.130s
user	0m0.080s
sys	0m0.046s
[root@void ~]# time xbps-query -l | awk '{ print $2 }' | xargs -n1 xbps-uhelper getpkgname > /dev/null

real	0m3.280s
user	0m1.637s
sys	0m1.583s
```

camoz:
Hm, maybe it would be better to mention `xpkg` for a package list w/o version, and plain `xbps-query -l` for a package list with version.

abenson:
`xpkg` isn't part of the base system and is provided by a separate set of tools.

camoz:
Yes, but there are many places in the docs that refer to using a tool from `xtools`. Even on the same page it's mentioned already two times.

FWIW I'm in favor of my second idea:

> Hm, maybe it would be better to mention xpkg for a package list w/o version, and plain xbps-query -l for a package list with version.

bobertlo:
> `xpkg` isn't part of the base system and is provided by a separate set of tools.

I agree with this partly, but I think that it might be worth offering a simple command from xtools as an alternative to a long shell command people are going to just copy and paste.

camoz:
I think ideally xbps would support printing the package names natively, like pacman (and probably many other package managers):

```
# pacman -Q
a52dec 0.7.4-11
abcde 2.9.3-5
...
# pacman -Qq
a52dec
abcde
...
```

Even the first version would be enough, one could a do more simple `| cut -d' ' -f1` then.

paper42:
I would be in favor of mentioning this, but I think we should reverse the order - first the long command using only packages from base, then a shorter alternative - the xpkg script from xtools.

PR #643, opened 2021-12-27
src/xbps/troubleshooting/static.md: Correct location of example mirror

camoz:


camoz:
AFAICS that mirror is in finland now, not germany.

camoz:
Incorporated into https://github.com/void-linux/void-docs/pull/644.

PR #661, opened 2022-03-02
config/media/pipewire.md: mention system services

paper42:
A lot of people just start the services when they don't have to without checking the warnings in the services and in README.voidlinux and run into unnecessary issues. This PR adds one more place with a warning.


Vaelatern:
fails fmt lint, can be fmted

PR #663, opened 2022-03-20
config/users-and-groups.md: add plugdev to groups

0x5c:
~~This depends on void-linux/void-packages#36238 being merged.~~

This documents the existence of the new default group.

### DIFF
--- a/src/config/users-and-groups.md
+++ b/src/config/users-and-groups.md
@@ -77,6 +77,7 @@ Void Linux defines a number of groups by default.
 | `network`  | Unused - present for historical reasons.                      |
 | `kvm`      | Ability to use KVM for virtual machines, e.g. via QEMU.       |
 | `input`    | Access to input devices: `/dev/mouse*`, `/dev/event*`.        |
+| `plugdev`  | Access to pluggable devices.                                  |
 | `nogroup`  | System daemons that don't need to own any files.              |
 | `users`    | Ordinary users.                                               |
 | `xbuilder` | To use xbps-uchroot(1) with `xbps-src`.                       |


tornaria:
It seems important to add `plugdev` to this table.

The proposed description "Access to pluggable devices." is not completely obvious to me. Maybe we can try to be more specific? For example, mention that this group controls mtp access (at least via libmtp). I'm not sure what other "pluggable devices" are.

In debian the description is: "plugdev: Allows members to mount (only with the options nodev and nosuid, for security reasons) and umount removable devices through pmount. "  Does it make sense to do the same?

0x5c:
I don't think what `plugdev` does on void quite matches up with what it did on debian before they made it irrelevant by having logind in the base install. You mention `pmount`, but as far as I can tell there's no mention of `plugdev` in our `pmount` package nor its manpage, and I doubt it would actually do anything about `plugdev` unless we configure it to do so. Apart from `libmtp` (and as far as I can tell), the other packages that use it are unrelated to mounting stuff; all they have in common is generic access to usb devices (things like `libsigrok`, `openocd`, some razer mouse thing, etc).

On Debian it seems it was focused mainly on mounting stuff, but here we're mostly seeing the result of other projects deciding to use that group for more devices (rightly so I think); they were describing was it was supposed to mainly be, while we're after the fact describing what people have decided they would use it for. 

I don't think we can limit the description to "mounting stuff", but you're damn right the description is lacking lol
I'm not sure what would be best, adding some examples and/or finding better wording.

0x5c:
There's also some packages too that currently (and wrongly) patch out plugdev in udev rules in favour of `uaccess`, which breaks those packages for everyone who does not have elogind. I plan on making a PR to change the patches to add `uaccess` instead of replacing plugdev with it. Those packages are `airspy` & `hackrf` (two SDRs), `libnfc` (library for NFC readers/writers), and `openobex` (library for binary transfers on IR ports and bluetooth).

Security keys (yubikeys and such) also normally use plugdev through `linfido2`, but on void there's a policy of having udev rules for those be patched to use `users` instead. It doesn't break those packages since that group is also a default group, so I don't feel like challenging that policy.

0x5c:
There's probably more packages that use plugdev or allow usage of it through udev rules, but it's not trivial to search through package contents that way.

0x5c:
Did grep over all the udev rules in the x64-glibc repos, only one not mentioned already is `fingerprint-gui` for access to fingerprint devices. So that's not much more, at least for the udev rules. Grepping over the entire repo would be completely unreasonable without direct access to the it.

So it seems to be a mix of mtp devices and niche devices

PR #667, opened 2022-04-02
Update theme, add light/dark mode switch

classabbyamp:
This PR builds off of @ericonr's WIP PR #581, and adds UI elements for printing and editing void-docs, as well as a
button to toggle between light and dark modes.

Examples of the updated version is live at https://void-book.qrm.cat

closes #581

### DIFF
--- a/book.toml
+++ b/book.toml
@@ -6,6 +6,9 @@ title = "Void Linux Handbook"
 
 [output.html]
 theme = "src/theme"
+default-theme = "void-light"
+preferred-dark-theme = "void-dark"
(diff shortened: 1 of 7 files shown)


ericonr:
This looks mostly ok to me :) Thanks for pushing it ahead!

The print button sends me to https://void-book.qrm.cat/print.html and opens the print dialog for the About page, so there seems to be a bug there.

classabbyamp:
Hm, that might be a limitation of the mdbook printing capability. I'll take a look into it

classabbyamp:
can confirm. for example, [this mdbook](https://doc.rust-lang.org/rustdoc/lints.html) will go back to the first page when you press the print button.

PR #671, opened 2022-04-23
xbps/repositories/custom.md: document creating a repository

classabbyamp:
- xbps/repositories/signing.md was merged into custom.md with minor modifications
  - signing.md is now redirected to custom.md
- incorporates @camoz's change to show using 4096-bit RSA keys for signing

closes #641


Duncaen:
I don't really see the usefulness of documenting `xbps-rindex -a` or the `XBPS_TARGET_ARCH` environment variable.
Usually people should just be using xbps-src to create the repositories and then sign them.

classabbyamp:
> Usually people should just be using xbps-src to create the repositories and then sign them

That doesn't help when someone builds locally with xbps-src and uploads the built package to a server to be served. When I was setting up a repository for personal use, I had to search through a bunch of scripts to figure out how to do it without xbps-src's help.

Duncaen:
I don't really like mixing "custom repository" with the signing stuff, usually people should be using xbps-src and then just sign the resulting repository index.

classabbyamp:
So should the signing info be put back as a separate page?

Duncaen:
I think same page is ok, but a separate section for signing and "custom repositories" and maybe instead of calling the `xbps-rindex` stuff "custom" call it something like "manually maintaining repository".

camoz:
Can we remove the `-des3`? Not sure what it does in the command, but DES3 has been broken for some years, and the `-des3` in the command was added a long time ago even before DES3 was broken.

classabbyamp:
I've amended based on feedback.

PR #675, opened 2022-05-18
res/void-docs: add evince to fallback pdf viewers

paper42:
evince is installed on the GNOME isos, but xdg-open isn't, so
void-docs -p fails.

cc @ericonr


ericonr:
Sorry for missing this, thanks!

PR #680, opened 2022-05-29
Add troubleshooting for hardware video decoding on chromium based browsers

Animeshz:


Animeshz:
Lmk if note section is useful or redundant.

PR #681, opened 2022-05-29
Update default mirror URL in sed command

sburris0:
While changing my mirrors, I noticed that the sed command replaces a URL (`https://alpha.de.repo.voidlinux.org`) different than the one present on my system (`https://repo-default.voidlinux.org`). This PR updates the command.


Vaelatern:
Please change commit message to `xbps/mirrors: update default mirror`

PR #683, opened 2022-06-02
fix CI for git 2.35.2+, fix outstanding CI failures

classabbyamp:
tested [here](https://github.com/void-linux/void-docs/actions/runs/2430553725)

### DIFF
--- a/book.toml
+++ b/book.toml
@@ -20,4 +20,4 @@ optional = true
 [output.linkcheck]
 optional = true
 follow-web-links = true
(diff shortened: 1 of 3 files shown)


classabbyamp:
check failure is expected because it was not run on f18496e74e412e176c18bd6cd29ffd5f8079b1ba, which contained a formatting error.

The other failure is a transient issue with one of the mirrors in China. Perhaps it should be added to the linkcheck ignore list.

PR #684, opened 2022-06-23
Add Switzerland mirror to mirror list

millhaus:
add swiss mirror at mirror.puzzle.ch


paper42:
Commit message should be `mirrors: <current commit body>`

PR #685, opened 2022-06-23
mirrors: add swiss mirror at mirror.puzzle.ch

millhaus:


classabbyamp:
please remove 8ec196db4ea8ee19f85a17b9edf55a8cd7e42b8f, it's unrelated to this change and breaks lint.

classabbyamp:
and squash so just the first commit exists, thanks

PR #687, opened 2022-07-03
pipewire: mention xdg autostart files

paper42:
needs [void-packages#37822](https://github.com/void-linux/void-packages/pull/37822) to be merged first

### DIFF
--- a/src/config/media/pipewire.md
+++ b/src/config/media/pipewire.md
@@ -14,6 +14,18 @@ environment or [startx](../graphical-session/xorg.md#startx). The `pipewire`
 package provides `pipewire` and `pipewire-pulse` system services, but they are
 not recommended for a typical setup.
 
+The `pipewire` package ships [Desktop
+Entry](https://specifications.freedesktop.org/desktop-entry-spec/latest/) files
+for `pipewire` and `pipewire-pulse` in `/usr/share/applications`. If your
+environment supports the [Desktop Application Autostart
+Specification](https://specifications.freedesktop.org/autostart-spec/autostart-spec-latest.html),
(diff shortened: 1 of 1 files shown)


src/config/media/pipewire.md
```diff
@@ -14,6 +14,15 @@ environment or [startx](../graphical-session/xorg.md#startx). The `pipewire`
 package provides `pipewire` and `pipewire-pulse` system services, but they are
 not recommended for a typical setup.
 
+The `pipewire` package ships XDG Autostart files for `pipewire` and
+`pipewire-pulse` in `/usr/share/doc/pipewire`. If your desktop environment
+supports the XDG Autostart specification, you can enable pipewire by symlinking
+the files from documentation to the XDG autostart directory:
+
+```
+#: ln -s /usr/share/pipewire/pipewire.desktop /etc/xdg/autostart/pipewire.desktop
```
classabbyamp, line 23:
```suggestion
# ln -s /usr/share/pipewire/pipewire.desktop /etc/xdg/autostart/pipewire.desktop
```
is the usual style I think

paper42: thanks, fixed

st3r4g:
Maybe also mention it can be done per user with:
```
$ ln -s /usr/share/applications/pipewire.desktop ~/.config/autostart
```

paper42:
> Maybe also mention it can be done per user with:
> 
> ```
> $ ln -s /usr/share/applications/pipewire.desktop ~/.config/autostart
> ```

I can't think of a common usecase where enabling autostart per-user would make sense. If there is one, we can mention it, but people can probably figure it out if they want something less usual.

PR #693, opened 2022-08-02
installation/guides/chroot: improve the fstab section

r-ricci:
From mount(8):

> **rw**
>     Mount the filesystem read-write.

> **noatime**
>    Do not update inode access times on this filesystem (e.g. for faster access on the news spool to speed up news servers). This works for all inode types (directories too), so it implies nodiratime.

Swap is not a filesytem and it doesn't have inodes, so these options are useless.

Since this is just an example, it's ok if it is opinionated, equal to the defaults or not suitable for all use cases, but it should not be wrong or irrelevant.


classabbyamp:
the installer does this: https://github.com/void-linux/void-mklive/blob/master/installer.sh.in#L1042

classabbyamp:
this section no longer exists with #826

PR #694, opened 2022-08-04
elogind: mention polkit for loginctl {poweroff,reboot}

motorto:
Refer that elogind might depend on polkit for some use cases. (poweroff/reboot). 

https://github.com/void-linux/void-packages/pull/38347


classabbyamp:
commit message should be `config/power-management: ...` i think

motorto:
It doesn't seem to be the problem to be honest don't know what I am
doing wrong, the `./check` script doesn't detect nothing wrong.

classabbyamp:
run `vmdfmt -w src/config/power-management.md`, it'll reflow the text

src/config/power-management.md
```diff
@@ -17,7 +17,9 @@ and enabled. Either disable `acpid` when enabling `elogind`, or configure
 `elogind` to `ignore` ACPI events in
 [logind.conf(5)](https://man.voidlinux.org/logind.conf.5). There are several
 configuration options, all starting with the keyword `Handle`, that should be
-set to `ignore` to avoid interfering with `acpid`.
+set to `ignore` to avoid interfering with `acpid`. In order to be able to use
+`loginctl {poweroff,reboot}`as a normal user it is needed to install the polkit
+package.
```
mhmdanas, line 22: I think this is better added in a separate paragraph:

```suggestion
set to `ignore` to avoid interfering with `acpid`.

In order to be able to use `loginctl poweroff` and `loginctl reboot` as a normal user, install the `polkit` package.
```

mhmdanas: Also, I think I recall that `polkit` is actually required to run something like Sway at all?

motorto: Thanks, sorry for the delay totally forgotten this.

> Also, I think I recall that `polkit` is actually required to run something like Sway at all?


I recall that also, not 100%, but I think that without polkit package installed and service running, it will not launch. (doesn't require a polkit client).

src/config/power-management.md
```diff
@@ -19,6 +19,9 @@ and enabled. Either disable `acpid` when enabling `elogind`, or configure
 configuration options, all starting with the keyword `Handle`, that should be
 set to `ignore` to avoid interfering with `acpid`.
 
+In order to be able to use `loginctl {poweroff,reboot}`as a normal user it is
+needed to install the `polkit` package.
```
mhmdanas, line 23:
```suggestion
In order to be able to use `loginctl poweroff` and `loginctl reboot` as a
normal user it is needed to install the `polkit` package.
```

mhmdanas (changes requested):

src/config/power-management.md
```diff
@@ -19,6 +19,9 @@ and enabled. Either disable `acpid` when enabling `elogind`, or configure
 configuration options, all starting with the keyword `Handle`, that should be
 set to `ignore` to avoid interfering with `acpid`.
 
+In order to be able to use `loginctl poweroff` and `loginctl reboot`as a
+normal user it is needed to install the `polkit` package.
```
classabbyamp, line 23: maybe

> To run `loginctl poweroff` and `loginctl reboot` without root privileges, `polkit` must be installed.

motorto: Sounds okay to me, that changed.

PR #697, opened 2022-08-14
user-service: parametrize $USER using dir name

Goorzhel:
I took a page from agetty's book:

```
~ ❯ cat /etc/sv/runsvdir-ag/run
#!/bin/sh

export USER="${PWD##*-}"
export HOME="/home/$USER"

groups="$(id -Gn "$USER" | tr ' ' ':')"
svdir="$HOME/.sv"

exec chpst -u "$USER:$groups" runsvdir "$svdir"
~ ❯ pgrep -af "runsvdir-$USER"
15673 runsv runsvdir-ag
```
<details>
<summary>Less interesting</summary>

```
~ ❯ ls -lR .sv
.sv:
total 4
drwxr-xr-x 3 ag users 4096 Aug 11 17:06 uhhhh

.sv/uhhhh:
total 8
-rwxr--r-- 1 ag users   36 Aug 11 17:05 run
drwx------ 2 ag users 4096 Aug 11 17:09 supervise

.sv/uhhhh/supervise:
total 12
prw------- 1 ag users  0 Aug 11 17:06 control
-rw------- 1 ag users  0 Aug 11 17:06 lock
prw------- 1 ag users  0 Aug 11 17:06 ok
-rw-r--r-- 1 ag users  6 Aug 11 17:09 pid
-rw-r--r-- 1 ag users  4 Aug 11 17:09 stat
-rw-r--r-- 1 ag users 20 Aug 11 17:09 status
~ ❯ cat .sv/uhhhh/run
#!/bin/sh

echo "uhhhh" && sleep 60
```
</details>

### DIFF
--- a/src/config/services/user-services.md
+++ b/src/config/services/user-services.md
@@ -13,8 +13,8 @@ with the following `run` script, which should be executable:
 ```
 #!/bin/sh
 
-export USER="<username>"
-export HOME="/home/<username>"
+export USER="${PWD#*-}"
+export HOME="/home/$USER"
 
 groups="$(id -Gn "$USER" | tr ' ' ':')"
 svdir="$HOME/service"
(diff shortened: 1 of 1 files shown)


Vaelatern (changes requested):
Seems like a reasonable change here. Especially since it's user-installed.

But:

```
$ A="a-b-c"
$ echo "${A##*-}"
c
$ echo "${A#*-}"
b-c
```

Please do 1 `#` not 2.

Goorzhel:
I should've remembered usernames can contain hyphens too.

Duncaen:
The agetty stuff should definitively not be copied, its a lot more trouble than its worth it.

My opinion on this is that we either ship example scripts in the docs that are basic and easy to understand and hard to get wrong so users can just copy them without messing up things. The alternative is to actually ship user services, but even then I would suggest against the shared run script setup.

Goorzhel:
... I _also_ should've known somebody would've had this idea before I did.

PR #711, opened 2022-10-02
mirrors: remove mirror.aarnet.edu.au

paper42:
no sync for 481 days


paper42:
actually it's not that old, but it's not recent either, linux is still at 5.18 (at least 4 days old), but linux5.19 is at 5.19.10 which is the latest version (11 days old)

PR #717, opened 2022-11-10
config/network: add groups for ConnMan and IWD

cinerea0:
I saw this in IRC and didn't want it to get lost. This commit adds information on the groups users need to belong to in order to operate ConnMan and IWD.


paper42:
Should we also mention that the user has to be in the wheel group to control wpa_supplicant? How does this work when polkit is enabled? Are the groups still required? If not, this should be documented.

cinerea0:
> Should we also mention that the user has to be in the wheel group to control wpa_supplicant?

Sure, I've added wording to wpa_supplicant's page that indicates this in the latest push.

> How does this work when polkit is enabled? Are the groups still required?

As far as I'm aware polkit only interacts with NetworkManager. I'm not 100% sure about that.

src/config/network/connman.md
```diff
@@ -15,6 +15,8 @@ Finally, enable the `connmand` service.
 
 ## Configuring ConnMan
 
+Users of ConnMan must belong to the `network` group.
```
classabbyamp, line 18: can also use polkit, but i'm not sure if it works without the network group

src/config/network/wpa_supplicant.md
```diff
@@ -17,6 +17,10 @@ pre-shared keys for use in configuration files.
 [wpa_cli(8)](https://man.voidlinux.org/wpa_cli.8) provides a CLI for managing
 the `wpa_supplicant` daemon.
 
+You must be a member of the group assigned as the `ctrl_interface_group` in the
+configuration file to interact with `wpa_supplicant`. On Void, this defaults to
+`wheel`.
```
classabbyamp, line 22:
```suggestion
configuration file to interact with `wpa_supplicant`. This defaults to `wheel`.
```

PR #718, opened 2022-11-12
add perl-File-Which to list of required packages

kevcrumb:
Resolves error:

    [INFO] (mdbook::renderer): Invoking the "gemini" renderer
    Can't locate File/Which.pm in @INC (you may need to install the File::Which module) (@INC contains: /usr/lib/perl5/site_perl /usr/share/perl5/site_perl /usr/lib/perl5/vendor_
    perl /usr/share/perl5/vendor_perl /usr/lib/perl5/core_perl /usr/share/perl5/core_perl) at void-docs/res/mdbook-gemini line 6.
    BEGIN failed--compilation aborted at void-docs/res/mdbook-gemini line 6.


paper42:
Squash the two commits to one and the commit message should be something like `README: add perl-File-Which to required packages`

PR #721, opened 2022-11-22
Fix minor mistake in FDE installation guide

mhmdanas:


classabbyamp:
next time, commit message should be: `installation/guides/fde: description of changes`

mhmdanas:
@classabbyamp my bad, I missed the part about commit messages in the contributing guidelines.

PR #722, opened 2022-11-23
ci: split scripts, add commit lint

classabbyamp:
- res/ci: split CI steps into separate scripts
- res/ci/commit-lint.sh: add commit lint script
- .github/workflows/ci.yaml: split into multiple steps

### DIFF
--- a/res/ci/commit-lint.sh
+++ b/res/ci/commit-lint.sh
@@ -1,0 +1,46 @@
+#!/bin/sh
+
+die() {
+	printf '%s\n' "$*" >&2
+	exit 1
+}
+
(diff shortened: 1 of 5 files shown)


res/ci/commit-lint.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/sh
```
the-maldridge, line 1: Please refactor this component to be like `-packages`.

the-maldridge (changes requested):

classabbyamp:
confirmed working with treeless checkout action: https://github.com/void-linux/void-docs/actions/runs/4031890940

PR #726, opened 2022-11-29
texi2pdf and texi2html usage on Void with tlmgr

giuliods:
A texinfo file can be exported in pdf and html using the two programs texi2pdf and texi2html. These programs require a file called texinfo.tex to compile first in latex and then in the actual format. This file is expected to be included with the texinfo installation or to be a dependency resolved by the package manager, Void uses tlmgr which doesn't automatically install the required package and probably this is why the error occur.

I added a short paragraph explaining the problem and making a fast solution to be used with tlmgr.

### DIFF
--- a/src/config/texlive.md
+++ b/src/config/texlive.md
@@ -99,9 +99,9 @@ For a full description, check:
 
 ## Texinfo PDF and HTML compilation with TeX Live
 Texinfo is the documentation system used by GNU and other projects and can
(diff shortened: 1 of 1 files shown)


giuliods:
Sorry, I had some troubles with commits, I'll not open a new request because i found out a problem and i have to research a little bit more. Next request will stay open

PR #731, opened 2022-12-08
config/php.md: new PHP docs to help users with the `php` metapackage transition

TinfoilSubmarine:
Hopefully this will help some users with the rocky but necessary `php` metapackage transition: https://github.com/void-linux/void-packages/pull/38536

### DIFF
--- a/src/config/php.md
+++ b/src/config/php.md
@@ -1,0 +1,35 @@
+# PHP
+
+There are two ways to install PHP packages with XBPS:
+
+1. Using the versioned packages (recommended).
+2. Using the meta-packages.
+
+## Versioned PHP Packages
+
+It is generally recommended to use versioned PHP packages (e.g. `php8.1`,
+`php8.1-apcu`, etc.) for most use cases as this ensures a consistent environment
+on updates with minimal or no intervention required.
+
+## PHP Meta-packages
+
+In Void, the `php` package is a meta-package that points to the latest upstream
+PHP version. This convention is followed by all packages prefixed with `php-`,
+such as `php-fpm`, as well as `xdebug` and `composer`. See [the `php`
+template](https://github.com/void-linux/void-packages/blob/master/srcpkgs/php/template)
+for a complete list. It is recommended to only use these meta-packages for
+development purposes.
+
+When using a PHP meta-package, be warned that updating may require manual
+intervention if a new major PHP version has been added to the repository. As a
+part of the version change, the configuration location will change to reflect
+the new version. For example, upgrading from 8.0 to 8.1 would result in the
+configuration path changing from `/etc/php8.0` to `/etc/php8.1`. Any
+customizations that have been made need to be manually applied to the new
+configuration directory.
+
+`php-fpm` updates require special care since they include a runit service. In
+this case, ensure that the new runit service is started and that applications
+using the previous version of `php-fpm` can access the new `php-fpm` instance.
+In particular, make sure any applications accessing the FPM instance have the
+correct TCP/unix socket address.
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -69,6 +69,7 @@
       - [libvirt](./config/containers-and-vms/libvirt.md)
       - [LXC](./config/containers-and-vms/lxc.md)
    - [OpenPGP](./config/openpgp.md)
+   - [PHP](./config/php.md)
 - [XBPS Package Manager](./xbps/index.md)
    - [Advanced Usage](./xbps/advanced-usage.md)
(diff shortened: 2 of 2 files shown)


paper42:
At the top, I would suggest to use the versioned packages for most usecases in order to not break stuff and only use the php metapaclage for development.

TinfoilSubmarine:
I reworked the wording a little and tried to make sure it was clear that the bottom 2 paragraphs only apply if you opt to use meta-packages instead of the versioned ones.

TinfoilSubmarine:
This is docs for https://github.com/void-linux/void-packages/pull/38536.

I have a template for 8.2 that builds and runs but haven't had time to test it. I also hoped we could get both of these merged before adding 8.2 to repo.

paper42:
That's unfortunately blocked right now on finding the best way to handle alternatives, see more information in the PR.

paper42:
I would have two sections in the php documentation
- one explaining versioned php packages
- one explaining what the metapackage is and what using it could cause

Right now, we are mixing together both versioned packages and the metapackage which can easily confuse people.

paper42:
> I think you guys should check what sury does (Debian/Ubuntu packager for PHP). Using it for production for several years. I think `php` points to the latest always. We can install multiple php versions at the same time and switch as well. `php8.2-*` `php7.4-*` and so on. Maybe its better to drop php7.x since it is EOL from upstream. [php.net/supported-versions.php](https://www.php.net/supported-versions.php) Announce in blog and remove after a month?

We have been doing something like that since 8,0, this is about alternatives and repurposing the old php (7) package for a metapackage that follows the latest php version.

paper42 (approved):

src/config/php.md
```diff
@@ -0,0 +1,35 @@
+# PHP
+
+There are two ways to install PHP packages with XBPS:
+
+1. Using the versioned (otherwise known as pinned) packages (reccomended).
```
mhmdanas, line 5:
```suggestion
1. Using the versioned (otherwise known as pinned) packages (recommended).
```

src/config/php.md
```diff
@@ -0,0 +1,35 @@
...
+2. Using the metapackages.
+
+## Pinned PHP Packages
+
+It is generally recommended to use versioned php packages (e.g. `php8.1`,
+`php8.1-apcu`, etc.) for most use cases as this ensures a consistent environment
+on updates with minimal or no intervention required.
+
+## PHP Metapackages
```
mhmdanas, line 14:
```suggestion
## PHP Meta-packages
```

src/config/php.md
```diff
@@ -0,0 +1,35 @@
...
+
+In Void, the `php` package is a meta-package that points to the latest upstream
+PHP version. This convention is followed by all packages prefixed with `php-`,
+such as `php-fpm`, as well as `xdebug` and `composer`. See [the php
+template](https://github.com/void-linux/void-packages/blob/master/srcpkgs/php/template)
+for a complete list. It is recommended to only use these meta-packages for
+development purposes.
+
+When using a PHP meta-package, be warned that updating may require manual
+intervention if a new major PHP version has been included in the repository. As
```
mhmdanas, line 24: Change "included" to "added" maybe?

src/config/php.md
```diff
@@ -0,0 +1,35 @@
...
+a part of the version change, the configuration location will change to reflect
+the new version. For example, upgrading from 8.0 to 8.1 would result in the
+config path changing from `/etc/php8.0` to `/etc/php8.1`. Any configurations
+that have been customized need to be manually applied to the new configuration
+directory.
```
mhmdanas, line 29: I think this reads a little better, but feel free to ignore it: 
```suggestion
configuration path changing from `/etc/php8.0` to `/etc/php8.1`. Any
customizations that have been made need to be manually applied to the new
configuration directory.
```

src/config/php.md
```diff
@@ -0,0 +1,35 @@
+# PHP
+
+There are two ways to install PHP packages with XBPS:
+
+1. Using the versioned (otherwise known as pinned) packages (reccomended).
+2. Using the metapackages.
+
+## Pinned PHP Packages
```
mhmdanas, line 8:
```suggestion
## Versioned/Pinned PHP Packages
```

paper42: I like Versioned PHP packages more, pinned implies something like holding or repolocking it in the package manager to me.

mhmdanas: To be honest, that's also what I thought when I saw "pinned" being used here. It just seems out of place.

paper42:
@TinfoilSubmarine this should be good to go, only please fix the commit lint

PR #732, opened 2022-12-17
config/network/index: Also change the link to supervise symlink

vincele:
In the section talking about interface-specific dhcpcd, the run script is modified with sed to have the proper interface name, but there is also the supervise symlink, pointing into /run/runit.

The target is "supervise.dhcpcd-eth0", which kind of works as-is, until you also have an actual eth0 interface that is handled by a dhcpcd supervised by runit.

So to avoid having this misleading symlink, change it to a proper target: "supervise.dhcpcd-enp3s0" in the sample code.

Signed-off-by: Vincent Legoll <vincent.legoll@gmail.com>

### DIFF
--- a/src/config/network/index.md
+++ b/src/config/network/index.md
@@ -38,6 +38,7 @@ $ ip link show
 2: enp3s0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast state UP mode DEFAULT group default qlen 1000
         link/ether ff:ff:ff:ff:ff:ff brd ff:ff:ff:f
 # cp -R /etc/sv/dhcpcd-eth0 /etc/sv/dhcpcd-enp3s0
+# ln -nsf /run/runit/supervise.dhcpcd-enp3s0 /etc/sv/dhcpcd-enp3s0/supervise"
 # sed -i 's/eth0/enp3s0/' /etc/sv/dhcpcd-enp3s0/run
 # ln -s /etc/sv/dhcpcd-enp3s0 /var/service/
 ```


vincele:
Talking about the following supervise symlink which is copied  to /etc/sv/dhcpcd-enp3s0
<pre>
# ls -lAh /etc/sv/dhcpcd-eth0
total 4.0K
-rwxr-xr-x 1 root root 65 Apr  6  2022 run
lrwxrwxrwx 1 root root 32 Apr  6  2022 supervise -> /run/runit/supervise.dhcpcd-eth0
</pre>

vincele:
I've not seen the interface clash happen, I just saw the symlink referencing an interface that is not existing here.

Vaelatern:
I think this is the wrong way to handle this. Probably should just make sure the supervise symlink is created by runit correctly in the first place.

classabbyamp:
if the `dhcpcd-eth0` service has never been run, I believe it shouldn't have any supervise symlink. runit should create the correct supervise link itself when it starts the service.

vincele:
> I think this is the wrong way to handle this. Probably should just make sure the supervise symlink is created by runit correctly in the first place.

I tried to remove the symlink, the supervise directory is created in /etc/runit/runsvdir/default/dhcpcd-enp37s0. Not what we want.

I tried to create the symlink in the run script (even if I knew that's dumb...), the directory is created before running that script (obviously) so it's not working either.

Is there a way to make runit do it itself ?

classabbyamp:
> the supervise directory is created in /etc/runit/runsvdir/default/dhcpcd-enp37s0

I think you'll find that `/var/service` is a chain of symlinks leading to `/etc/runit/runsvdir/default`

vincele:
OK, looks like I don't know what I'm talking about, disregard that PR.
Sorry for the noise.

PR #735, opened 2023-01-16
About zfs.md links

xiarenfan:
The link leads to a 404, and modified the link


ahesford:
Thanks! We did some restructuring and overlooked the links here.

PR #702, opened 2022-09-05
pipewire: add media session notes

CameronNemo:
Based on recent changes to the pipewire package, there is an easy way to
switch to wireplumber or disable p-m-s autostart entirely.

https://github.com/void-linux/void-packages/pull/38521

PR #738, opened 2023-01-26
config/media/pipewire.md: document use of wireplumber

ahesford:
When we actually remove `pipewire-media-session` and don't override `context.exec` in the default `pipewire.conf`, the configuration part will need to be rewritten.

### DIFF
--- a/src/config/media/pipewire.md
+++ b/src/config/media/pipewire.md
@@ -11,10 +11,7 @@ $ pipewire
 
 When pipewire works as expected, use the autostarting mechanism of your desktop
 environment or [startx](../graphical-session/xorg.md#startx). The `pipewire`
-package provides `pipewire` and `pipewire-pulse` system services, but they are
-not recommended for a typical setup.
-
-The `pipewire` package ships [Desktop
+package ships [Desktop
 Entry](https://specifications.freedesktop.org/desktop-entry-spec/latest/) files
 for `pipewire` and `pipewire-pulse` in `/usr/share/applications`. If your
 environment supports the [Desktop Application Autostart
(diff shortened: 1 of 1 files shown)


Vaelatern:
Could we maybe also have a brief description of "Why"? I don't know why these things matter and I'm curious.

src/config/media/pipewire.md
```diff
@@ -26,6 +23,66 @@ directory:
...
+```
+$ : "${XDG_CONFIG_HOME:=${HOME}/.config}"
+$ mkdir -p "${XDG_CONFIG_HOME}/pipewire"
+$ sed '/path.*=.*pipewire-media-session/s/{/#{/' \
+    /usr/share/pipewire/pipewire.conf > "${XDG_CONFIG_HOME}/pipewire/pipewire.conf"
+
+```
+
+Now, configure `wireplumber` to start in the same fashion as the `pipewire`
+executable. The `wireplumber` package includes a Desktop Entry file that can be
```
oreo639, line 61: It is the opposite. wireplumber must be launched after pipewire otherwise wireplumber will exit.

src/config/media/pipewire.md
```diff
@@ -26,6 +23,66 @@ directory:
...
+linked to the Desktop Application Autostart directory:
+
+```
+# ln -s /usr/share/applications/wireplumber.desktop /etc/xdg/autostart/wireplumber.desktop
+```
+
+When launching `pipewire` and `wireplumber` from a `.xinitrc` script, it may be
+necessary to launch `wireplumber` *before* `pipewire` to prevent the latter from
```
oreo639, line 69: It is the opposite, wireplumber must be launched after pipewire otherwise wireplumber will exit.

ahesford: Thanks. I launch `wireplumber` from `pipewire.conf` and wasn't sure about the ordering. This has been fixed.

ahesford:
@Vaelatern I've added a new introductory sentence to this section. Does that provide sufficient context?

Vaelatern:
Yes, thank you, that helps!

oreo639:
Also, I don't think there is a way to guarantee that the wireplumber.desktop scrip runs after pipewire.

The previous PR changed the pipewire docs to reccomend using `/etc/pipewire/pipewire.conf.d/` and `~/.config/pipewire/pipewire.conf.d/` rather than changing the `/etc/pipewire/pipewire.conf`:
https://github.com/void-linux/void-docs/pull/702

ahesford:
We should not ship configs that enable pipewire by default when the default is to not do that. It will eventually land us in the same mess we are now trying to resolve.

I'm currently using two `/etc/pipewire/pipewire.conf.d` snippets to launch `pipewire-pulse` and `wireplumber` in `context.exec` and the concatenation works fine. We can't recommend this solution until we drop the `pipewire-media-session` from our default configs because launching two session managers leads to failure. Until then, it doesn't make sense to tell people to copy the system config and edit that, just so they can later add configuration snippets alongside the overridden config.

src/config/media/pipewire.md
```diff
@@ -26,6 +23,66 @@ directory:
...
+start `wireplumber`, which may make process management simpler.
+
+> In its default configuration, `wireplumber` requires an active [D-Bus
+> session](../session-management.md#d-bus). When using Desktop Application
+> Autostart symlinks, it should be sufficient to wrap the launch of a window
+> manager or desktop environment in `dbus-run-session`. When launching
+> `pipewire` and `wireplumber` from a `.xinitrc` script, it may be necessary to
+> create a new shell script suitable for passing to `dbus-run-session` that will
+> launch `pipewire`, `wireplumber` and the window manager or desktop environment
+> in a common D-Bus session.
```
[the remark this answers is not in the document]
ahesford: How about I strike the "when launching from .xinitrc" paragraph (to avoid endorsing that as a solution) and change the note to

> In its default configuration, `wireplumber` requires an active [D-Bus session](../session-management.md#d-bus). When using Desktop Application Autostart symlinks, it should be sufficient to wrap the launch of a window manager or desktop environment in `dbus-run-session`. Users of window managers that do not respect the autostart mechanism should consult upstream documentation regarding the process for automatically starting programs at launch.

At some point, we have to be willing to tell people that this isn't Ubuntu and they might need to do some learning on their own.

cinerea0:
If/when use of `pipewire.conf.d` directories is documented in the handbook, would it make sense for it to say the order in which the configs are sourced?

ahesford:
> If/when use of `pipewire.conf.d` directories is documented in the handbook, would it make sense for it to say the order in which the configs are sourced?

No. That information is already described in [pipewire.conf(5)](https://man.voidlinux.org/pipewire.conf.5) so there is no need to reproduce it here.

ahesford:
I rewrote some things to prefer custom configuration edits that will make `pipewire` launch `wireplumber`. It also notes the `wireplumber.desktop` file and the ordering issues surrounding its use, but should give people enough hints to find their way around alternatives if they don't want to make the config change.

src/config/media/pipewire.md
```diff
@@ -26,6 +23,94 @@ directory:
 # ln -s /usr/share/applications/pipewire.desktop /etc/xdg/autostart/pipewire.desktop
 ```
 
+## Session Management
+
+In PipeWire, a session manager assumes responsibility for interconnecting media
+sources and sinks as well as enforcing routing policy. Without a session
+manager, PipeWire will not function. The reference
+[`pipewire-media-session`](https://gitlab.freedesktop.org/pipewire/media-session)
+package was originally provided and, in Void, configured to run by default to
```
paper42, line 32: I *think* this sounds better?
```suggestion
package was originally provided and, in Void and configured to run by default to
```

paper42 (approved):

src/config/media/pipewire.md
ahesford: I reworded to

```suggestion
package was originally provided in the Void `pipewire` package and configured to run by default to
```

PR #744, opened 2023-04-07
config/date-time: warn that rc.conf TIMEZONE should be removed when linking

classabbyamp:


classabbyamp:
ugh, it seems i've accidentally pushed not to my fork

PR #746, opened 2023-04-15
installation/live-images/index: describe a11y features and KCL params

classabbyamp:
for void-linux/void-mklive#295, void-linux/void-mklive#348, void-linux/void-mklive#349


classabbyamp:
note: should check on instructions for isolinux

PR #749, opened 2023-04-30
Add a section to static.md for when the /current repository isn't found.

amogusussy:
Yesterday, I accidently uninstalled the `xbps` package, and I went to the docs, followed the instructions, but when the `xbps` package was uninstalled, it also remove the /current repository, so I couldn't reinstall the `xbps` package.

This just gives instructions on how to install the `xbps` in the case that the /current repository is remove.

### DIFF
--- a/src/xbps/troubleshooting/static.md
+++ b/src/xbps/troubleshooting/static.md
@@ -33,3 +33,11 @@ that can no longer boot, it is recommended to chroot in using a Void
 installation medium and use the static tools from there, as it is unlikely that
 even a shell will work correctly on those systems. When using static XBPS with a
 glibc installation, the environment variable `XBPS_ARCH` needs to be set.
+
+## Repositories not found
+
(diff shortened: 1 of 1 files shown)


paper42:
I don't think this is necessary, the error message about missing repositories is clear, -R is documented in the manpage and people shouldn't even be able to remove xbps if they don't also remove the whole base system metapackage.

NACK from me

classabbyamp:
agree, this is already documented with the manpage

amogusussy:
@paper42 
There is no error message about missing repositories. It just tells you that whatever package you're trying to install doesn't exist, or if you're updating, it just updates whatever repos you do have, and gives no indication that the normal repository isn't installed.

While, yes, it is documented in the man page, it's in no way obvious that the issue lays with not having a repo installed, so it isn't obvious that you'd need to use the -R flag.

PR #759, opened 2023-05-29
pipewire, pulse: small cleanups

classabbyamp:
- config/media/pulseaudio: remove mention of pulse system service
    - if it's not recommended, why mention it? requires void-linux/void-packages#44141
- config/media/pipewire: symlink example config snippets
    - instead of writing them manually (requires void-linux/void-packages#44144)


ahesford (approved):
This is definitely better than making people duplicate the content of the example configs.

PR #763, opened 2023-07-10
config/media/sndio: remove

Duncaen:
Having this documentation makes it seem like its "supported" and users install or want to use it for some reason while its pretty incomplete and only supported by a handful of programs for which some require patches that I don't think I would maintain if they ever break.


classabbyamp (approved):

Vaelatern:
My only request is a followup comment on the primary multimedia page explaining that sndio exists, but is not supported nor recommended for use on Void. That way people do not feel gaslit, since we used to have a page on it.

Vaelatern (approved):

classabbyamp:
> My only request is a followup comment

added

PR #765, opened 2023-07-20
config/users-and-groups: add OpenDoas section.

0x5c:


Duncaen:
I don't think this is void specific enough to be worth including.

ethamck:
I'd like to see better `doas` documentation merged, but the essence of this already exists under https://docs.voidlinux.org/xbps/advanced-usage.html#ignoring-packages with `doas` provided as an example. Regardless, (having found the old wiki's article on doas before this) I do believe that referencing this section under sudo would be helpful for anyone seeking doas on Void.

Would simply linking to this heading be a welcome addition to the docs? Anyone looking for doas can more than likely [do their own research](https://wiki.archlinux.org/title/Doas) on how to configure it. The `ignorepkg` part is the Void-specific bit, so I'm not sure any more info is necessary.

0x5c:
That seems reasonable, I'll update the PR and we'll see if there's more interest then

PR #780, opened 2023-11-02
Update kde.md: include note about VTs and wayland

SamWilsn:


classabbyamp:
this might fit better on the wayland page, as SDDM is not exclusively for KDE

src/config/graphical-session/kde.md
```diff
@@ -19,6 +19,19 @@ If you wish to start an X-based session from the console, use
...
 session, run `startplasma-wayland` directly.
 
+## SDDM
+
+Note that if you are configuring SDDM to [run under wayland][sddm-wayland],
+you will need to ensure the first virtual terminal is not consumed by
+`agetty`:
+
+```
+rm /var/service/agetty-tty1
```
classabbyamp, line 29: this is not the recommended way to stop agetty services, as this will be restored by an update to runit-void. instead, `touch /etv/sv/agetty-tty1/down`

src/config/graphical-session/kde.md
```diff
@@ -32,3 +45,5 @@ package. If you want video thumbnails, the `ffmpegthumbs` package is also
 necessary. Enable previews in "Control" -> "Configure Dolphin" -> "General" ->
 "Previews" by checking the corresponding boxes. File previews will be shown for
 the selected file types after clicking "Preview" in Dolphin's toolbar.
+
+[sddm-wayland]: https://wiki.archlinux.org/title/SDDM#Running_under_Wayland
```
classabbyamp, line 49: I think our style is to use inline links

SamWilsn:
How's that? First time setting up void successfully, so I'm still not super familiar with runit.

classabbyamp:
the existence of a file called `down` in a service means runit will not start it, even if enabled (symlinked)

> If the file service/down exists, runsv does not start ./run immediately. The control interface (see below) can be used to start the service and to give other commands to runsv.

https://man.voidlinux.org/runsv.8

classabbyamp:
I think it makes more sense to frame it something like this, so we don't rely on archwiki:

> To run SDDM under Wayland, install `weston`, set `DisplayServer` to `wayland` in [sddm.conf(5)](https://man.voidlinux.org/sddm.conf.5), and disable agetty-tty1 ...

also, no need to mention xorg because that's not relevant in the wayland page

classabbyamp:
last thing, commit message should be something like `config/graphical-session/wayland: explain running SDDM on wayland`

SamWilsn:
Are you folks interested in merging, or should I close this?

PR #781, opened 2023-11-04
config/media/index: doesn't need audio group with seatd

Sertonix:
Any seat manager allows accessing of audio device without the `audio` group. Not only elogind.


classabbyamp:
this isn't true, I use seatd + pipewire and I need to be in the audio group

Sertonix:
Hm, let me check again.

Sertonix:
Ok, sorry. My mistake

classabbyamp:
the mechanism for not needing the audio group is uaccess, which only elogind currently implements

PR #783, opened 2023-11-25
wayland.md: Added Qtile to the list of Wayland compositors

heuer:
Added Qtile (qtile-wayland) to the list of Wayland compositors


classabbyamp:
commit message should be `config/graphical-session/wayland: add qtile to list of wayland compositors` please

PR #785, opened 2023-12-08
user-services.md: add session dbus and ssh example

tjheeta:
Added examples for setting up a session dbus and ssh-agent via user-services. References to setting up a session dbus were made in pipewire.md and session-management.md .

### DIFF
--- a/src/config/media/pipewire.md
+++ b/src/config/media/pipewire.md
@@ -13,7 +13,8 @@ bus](../session-management.md#d-bus). If your desktop environment, window
 manager, or Wayland compositor is configured to provide this, no further
 configuration should be required. If not, the desktop environment, window
 manager, or Wayland compositor may need to be launched with
-[`dbus-run-session(1)`](https://man.voidlinux.org/dbus-run-session.1).
+[`dbus-run-session(1)`](https://man.voidlinux.org/dbus-run-session.1) or a [dbus
+user-service](../services/user-services.md#d-bus-session) should be setup.
 
 PipeWire also requires the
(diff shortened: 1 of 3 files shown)


tjheeta:
I spent a few hours struggling with dbus until someone explained in irc how dbus was supposed to work. For instance, I was launching applications with dbus-launch and dbus-run-session to get them to connect to user dbus. 

With that in mind, I was considering adding the following blurb to session-management.md: 
>... in Linux. D-Bus can provide a system bus and/or a session bus, the
latter being specific to a user session. 

> The system bus is provided by `/etc/sv/dbus` and listens on `/run/dbus/system_bus_socket`. The session bus does not have a set location, but provides `DBUS_SESSION_BUS_ADDRESS` for any applications to find the listening socket. Note that for the session dbus to work as designed there can only be one instance per user account and `DBUS_SESSION_BUS_ADDRESS` must be in the environment of all dbus-using programs of that user pointing to the listening socket.

> To  provide ...

ahesford (commented):
I don't think we should be encouraging users to manage D-Bus as a user service when we don't have an officially supported mechanism for managing user services. While we note that the user-specific `runsvdir` is viable, it also has some caveats (for example, the D-Bus service that you provide has no means to reliably check if the service is functional). `dbus-run-session` is the most reliable method to ensure that the bus is functional when needed, and that's especially important for inexperienced users who are more likely to naively follow instructions in the documentation.

Furthermore, I don't think providing two arbitrary examples is particularly helpful. It might be better instead to note that user services are constructed in the same fashion as system services and advise users to consult those services for examples. People designing custom services will need to manage them, and it is better to foster a deeper understanding of things they will write than encouraging people to copy a simple example.

PR #791, opened 2024-02-18
installation/guides/fde: Correct GRUB bug link

MeganerdNL:
The link to the GRUB bug (59409) is pointing and referring to a duplicate of the real bug (55093).
This PR corrects that error.


abenson:
That would not be the correct bug. `luks2` support has been added, what's missing is `argon2i` support.

MeganerdNL:
> That would not be the correct bug. `luks2` support has been added, what's missing is `argon2i` support.

No the title of the bug is wrong because they only added patial LUKS2 support (due to libncrypt, but that's another story). You can read in the comments that the submitter is requesting that they change the title to "Add _full_ LUKS2 support". They did not close the report because of this. But I see the confusion as there is indeed a newer PR.

PR #792, opened 2024-02-23
session-management.md: Proposed way of handling $XDG_RUNTIME_DIR

### LINKED ISSUE #635: session-management.md: Proper handling of `XDG_RUNTIME_DIR`
The docs have a section called [XDG_RUNTIME_DIR](https://docs.voidlinux.org/config/session-management.html#xdg_runtime_dir) which advises users that are not in a `elogind` environment (which takes care of `XDG_RUNTIME_DIR` automatically) to manually manage it through the shell:

> Alternatively, manually set the environment variable through the shell. Make sure to create a dedicated user directory and set its permissions to 700. A good default location is /run/user/$(id -u).

This is not trivial to do while staying compliant to the [XDG spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html). Here's the relevant part:

>  $XDG_RUNTIME_DIR defines the base directory relative to which user-specific non-essential runtime files and other file objects (such as sockets, named pipes, ...) should be stored. The directory MUST be owned by the user, and he MUST be the only one having read and write access to it. Its Unix access mode MUST be 0700.
> 
> The lifetime of the directory MUST be bound to the user being logged in. It MUST be created when the user first logs in and if the user fully logs out the directory MUST be removed. If the user logs in more than once he should get pointed to the same directory, and it is mandatory that the directory continues to exist from his first login to his last logout on the system, and not removed in between. Files in the directory MUST not survive reboot or a full logout/login cycle.
> 
> The directory MUST be on a local file system and not shared with any other system. The directory MUST by fully-featured by the standards of the operating system. More specifically, on Unix-like operating systems AF_UNIX sockets, symbolic links, hard links, proper permissions, file locking, sparse files, memory mapping, file change notifications, a reliable hard link count must be supported, and no restrictions on the file name character set should be imposed. Files in this directory MAY be subjected to periodic clean-up. To ensure that your files are not removed, they should have their access time timestamp modified at least once every 6 hours of monotonic time or the 'sticky' bit should be set on the file.
> 
> If $XDG_RUNTIME_DIR is not set applications should fall back to a replacement directory with similar capabilities and print a warning message. Applications should use this directory for communication and synchronization purposes and should not place larger files in it, since it might reside in runtime memory and cannot necessarily be swapped out to disk. 

On IRC it was recently proposed to instead manage `XDG_RUNTIME_DIR` via [pam_rundir](https://github.com/void-linux/void-packages/tree/master/srcpkgs/pam_rundir) in environments without `elogind`.

It seems there is also another alternative already packaged: [rundird](https://github.com/void-linux/void-packages/tree/master/srcpkgs/rundird)

bermudi:
Manually setting `$XDG_RUNTIME_DIR` is not trivial to do while staying compliant to the XDG spec as noted in issue #635. This proposed way handles `$XDG_RUNTIME_DIR` properly without having to install `elogind` and should suggested as its alternative

fixes #635

### DIFF
--- a/src/config/session-management.md
+++ b/src/config/session-management.md
@@ -62,6 +62,8 @@ user-specific runtime files.
 Install [elogind](#elogind) as your session manager to automatically set up
 `XDG_RUNTIME_DIR`.
 
-Alternatively, manually set the environment variable through the shell. Make
-sure to create a dedicated user directory and set its permissions to `700`. A
-good default location is `/run/user/$(id -u)`.
+Alternatively, install [pam_rundir(8)](https://man.voidlinux.org/pam_rundir) and
+add it to `/etc/pam.d/login` to automatically create a $XDG_RUNTIME_DIR at login. 
+Manually setting the environment variable can be a bit cumbersome specially if you 
+need $XDG_RUNTIME_DIR at the default location of `/run/user/$(id -u)` for apps that 
+have it hardcoded. 


ahesford:
I reject your assertion that managing the environment variable is cumbersome, and the alternative you propose is both known to have consistency issues and hasn't been touched in several years.

The lifetime requirement on XDG runtime directories is without merit, especially because the spec itself declares that applications must properly handle situations where the runtime directory contains stale content.

PR #793, opened 2024-03-01
Simplification of the Full Disk Encryption installation. Split into two guides for BIOS und UEFI systems. 

Andy-ghub:
Simplification of the Full Disk Encryption installation.
* Less steps.
* Unification of the initrd systems dracut and mkinitcpio.
* With encrypted swap.

 Split in two guides for BIOS und UEFI systems. 
* More details for both systems.
* No more exceptions.

Tested with 32 Bit und 64 Bit.


abenson:
> Less steps.

Debatable.

> With encrypted swap.

It had encrypted swap before, but now its much less flexible with the removal of LVM.

> Split in two guides for BIOS und UEFI systems.

This will introduce a lot of redundancy in the guide, since there were only a few steps that differ between BIOS and UEFI systems.

Lastly, please make sure you follow [the contribution guide](https://github.com/void-linux/void-docs/blob/master/CONTRIBUTING.md#submitting-changes) and ensure your commits follow the guidelines.

ahesford:
As I explained in IRC, the removal of LVM reduces flexibility and actually *increases* complexity and fragility (you put two LUKS volumes on separate partitions of the same disk, which complicates key management for no benefit). These changes are not advisable in general-purpose instructions.

PR #801, opened 2024-06-28
Update chroot.md with network requirement hint for xchroot

kardwen:
When following the [Installation via chroot](https://docs.voidlinux.org/installation/guides/chroot.html#installation-via-chroot-x86x86_64aarch64) guide, I encountered the small issue that ``xchroot`` would return the message ``special device /etc/resolv.conf does not exist.`` I thought that maybe a hint indicating to ensure that a network connection is configured might clear this up.

### DIFF
--- a/src/installation/guides/chroot.md
+++ b/src/installation/guides/chroot.md
@@ -142,8 +142,9 @@ methods.
 ### Entering the Chroot
 
 [xchroot(1)](https://man.voidlinux.org/xchroot.1) (from `xtools`) can be used to
-set up and enter the chroot. Alternatively, this can be [done
-manually](../../config/containers-and-vms/chroot.md#manual-method).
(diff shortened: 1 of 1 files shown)


classabbyamp:
would probably be better to add this note earlier, as more than just xchroot in that guide would need network

also separately: it might be good to make xchroot not fail hard if resolv.conf doesn't work

kardwen:
I actually used ``void-installer`` and then skipped to <https://docs.voidlinux.org/installation/guides/chroot.html#configuration> to install [Limine](https://github.com/limine-bootloader/limine) as my bootloader because it does not require an additional partition like GRUB does for GPT partitioned disks on BIOS. Because I rebooted into the live environment for some reason and forgot to reestablish my network connection, I ran into the described issue.

PR #804, opened 2024-07-14
Update about-this-handbook.md

msiyer:
1. Starting positively is important: describing the purpose of the handbook is more important. 
2. Sentence describing what not to expect from the handbook has been moved to a better place. This can even be turned into a callout.
3. Consistency: replaced the word document with handbook.
4. Grammar edits: a) a comma must precede "and" that joins two independent clauses; b) article usage.

### DIFF
--- a/src/about/about-this-handbook.md
+++ b/src/about/about-this-handbook.md
@@ -1,14 +1,14 @@
 # About This Handbook
 
-This handbook is not an extensive guide on how to use and configure common Linux
-software. The purpose of this document is to explain how to install, configure,
-and maintain Void Linux systems, and to highlight the differences between common
+The purpose of this handbook is to explain how to install, configure,
+and maintain Void Linux systems and to highlight the differences between common
 Linux distributions and Void.
 
 To search for a particular term within the Handbook, select the 'magnifying
 glass' icon, or press 's'.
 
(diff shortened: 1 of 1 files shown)


Vaelatern:
Counter argument: Describing the negative purpose of the handbook, "This is NOT," is also important. If it goes anywhere later than the first sentence, fewer people will read it.

msiyer:
> Counter argument: Describing the negative purpose of the handbook, "This is NOT," is also important. If it goes anywhere later than the first sentence, fewer people will read it.

It is definitely important. This is the reason I proposed turning "This is NOT" into a "warning" callout.

msiyer:
@Vaelatern do you think the changes add value? should i add the callout for "This is NOT", make all checks successful, commit again?

tranzystorekk:
These changes feel mostly stylistic, without any substantial addition or correction. Unless there's some additional value I'm missing, it doesn't seem worth to include them.

ahesford:
Adding "an" in the "Example Commands" section changes the meaning of the sentence and is not correct.

Given the number of people who approach the documentation (and the IRC channel) expecting remedial Linux tutorials, emphasis that this guide is not intended to provide such content deserves top billing.

PR #806, opened 2024-07-18
Update pipewire.md

MartinGresens:
You need to start a session manager for pipewire to get access to the sound cards. So i add wireplumber to Launching Automatically.

### DIFF
--- a/src/config/media/pipewire.md
+++ b/src/config/media/pipewire.md
@@ -125,7 +125,7 @@ Server Name: PulseAudio (on PipeWire 0.3.82)
 
 ### Launching Automatically
 
-Once `pipewire` works as expected, it can be configured to launch when starting
+Once `pipewire` works as expected, `pipewire` and `wireplumber` can be configured to launch when starting
 a graphical session. There are several ways this can be achieved:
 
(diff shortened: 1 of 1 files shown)


classabbyamp:
the instructions before the Starting Automatically section say to set it up so launching pipewire launchers wireplumber.

classabbyamp:
notably:

https://github.com/void-linux/void-docs/blob/0d0395f7fa4f7c746fcb522d3e9c58cc9439862c/src/config/media/pipewire.md?plain=1#L53-L56

MartinGresens:
Yes but this is only for the configuration not for the autostart.
At user level you need
```
ln -s /usr/share/applications/wireplumber ~/.config/autostart
```
and systemwide
```
ln -s /usr/share/applications/wireplumber /etc/xdg/autostart
```
You shoud put it in the "[Session Management](https://docs.voidlinux.org/config/media/pipewire.html#session-management)" seccion or [Launching Automatically](https://docs.voidlinux.org/config/media/pipewire.html#launching-automatically)

classabbyamp:
no you don't, because pipewire was configured to start wireplumber. if you do this, you will end up with two wireplumbers.

Issue #803, opened 2024-06-29
Improve CSS highlighting of selected entry


kardwen:
I repeatedly had troubles to find the current active entry in the navigation tree view on the left. In the following example "Intel" is highlighted and I overlooked it frequently when scrolling down on my 15,6 Inch FHD+ IPS display.

Maybe adding the highlight color to the background would already help. Another idea would be to put the sidebar in its own scrollable container that automatically displays the according section in the viewport.

![grafik](https://github.com/void-linux/void-docs/assets/107949893/b004e037-63fd-4beb-b555-5067b1fb27f8)

PR #809, opened 2024-08-03
theme/css/general.css: emphasize active sidebar item

ethamck:
May (help) close https://github.com/void-linux/void-docs/issues/803, appends a bullet character (U+2022, •) to the end of a selected item in the sidebar for visibility.

For screen readers, this should render as "selected" instead.

Example

(I would've tried to color the background differently but I'm not a CSS wizard and this seemed more practical.)

### DIFF
--- a/src/theme/css/general.css
+++ b/src/theme/css/general.css
@@ -202,7 +202,14 @@ svg {
 	text-decoration: none;
 }
 #sidebar a.active {
-	color: var(--sidebar-active);
+	display: block;
+	padding: 0.5em;
+	width: 100%;
+	background-color: var(--sidebar-active);
+	color: var(--bg);
+}
+#sidebar a.active::after {
+	content: "" / " selected";
 }
 
 #sidebar-toggle {


src/theme/css/general.css
```diff
@@ -204,6 +204,9 @@ svg {
 #sidebar a.active {
 	color: var(--sidebar-active);
 }
+#sidebar a.active::after {
+	content: ' •' / ' selected';
+}
```
classabbyamp, line 209: how about this?

```suggestion
#sidebar a.active::after {
 	--size: 12px;
	content: "" / "selected";
	display: inline-block;
	width: var(--size);
	height: var(--size);
	-moz-border-radius: calc(var(--size) / 2);
	-webkit-border-radius: calc(var(--size) / 2);
	border-radius: calc(var(--size) / 2);
	background: var(--sidebar-active);
	margin-left: 0.5em;
}
```
- a bit bigger
- won't get messed up by font changes
- more centered

image

ethamck (commented):
I figured out one way to change background color rather than messing with `::after`s (it doesn't go as wide as I would've liked but I'm fairly certain it's impossible without rewriting a lot more than this)

dark theme

Not sure if the CSS rule is on the right line now that the selector's changed (or if it matters).

On the light theme, it may be difficult to read the text atop it,

light theme

but the Void green on white theme background doesn't technically meet WCAG contrast guidelines even without darkening it (4.44 original background vs 3.38 after dimming out of 4.50)

Should the dimming color change depending on theme? It's also possible to just change the alpha from 25% to a more adequate value for both themes (something like 10%).

classabbyamp:
looks good generally, a few things:
- maybe in the light theme, the background could be void green while the text is white or near-white? (kinda like inverse video)
- `has()` is a little iffy [support-wise](https://caniuse.com/css-has) but i think it's reasonable
- I think we should still have this, it's probably a good thing for screen readers:
```css
#sidebar a.active::after {
  content: "" / " selected";
}
```

ethamck:
I do prefer this approach, actually.

Light theme preview

Dark theme preview

It adheres to contrast guidelines when using `color: var(--bg)` and I rewrote it to not use `:has()` (rather styling the active `a` to take up the full width of `li`).

I also played with font weight and text color but concluded that this configuration is best.

classabbyamp (approved):

Issue #802, opened 2024-06-29
Add hint for font configuration under Wayland section


kardwen:
I installed River while following the article on [Wayland](https://docs.voidlinux.org/config/graphical-session/wayland.html) and had the issue that I could not start any terminal emulator with the keybinding in my init script (I tried out alacritty and food). This issue arised because I had not configured a font, I think it would be useful to add a reference to [Fonts](https://docs.voidlinux.org/config/graphical-session/fonts.html) under Installation or Configuration in this article.

classabbyamp:
done in #810

PR #810, opened 2024-08-03
config/graphical-session: hint to font installation for compositors

ethamck:
https://github.com/void-linux/void-docs/issues/802

Links to `fonts.md` under `wayland.md` and adds a hint to `fonts.md` that fonts are usually packaged with the `font-` prefix.

### DIFF
--- a/src/config/graphical-session/fonts.md
+++ b/src/config/graphical-session/fonts.md
@@ -1,5 +1,16 @@
 # Fonts
 
+A number of fonts and font collections are [available from
+XBPS](../../xbps/index.md#finding-files-and-packages). `dejavu-fonts-ttf` or
+`xorg-fonts` are a good baseline if you're unsure of what to pick.
+`noto-fonts-ttf` contains fonts for many languages and scripts. `noto-fonts-cjk`
+extends this with fonts for Chinese, Japanese, and Korean, and
+`noto-fonts-emoji` provides emojis. `nerd-fonts` provides a number of fonts with
+special characters like custom icons included.
+
+Fonts not available from XBPS can be manually installed to either
+`/usr/share/fonts` (system-wide) or `~/.local/share/fonts` (per-user).
+
 To customize font display in your graphical session, you can use configurations
 provided in `/usr/share/fontconfig/conf.avail/`. To do so, create a symlink to
 the relevant `.conf` file in `/etc/fonts/conf.d/`, then use
--- a/src/config/graphical-session/wayland.md
+++ b/src/config/graphical-session/wayland.md
@@ -30,6 +30,9 @@ Void Linux currently packages the following Wayland compositors:
 - labwc: a window-stacking compositor, inspired by Openbox
 - Qtile: a dynamic tiling Wayland compositor (via qtile-wayland)
 
+Some compositors do not depend on any [fonts](./fonts.md), which can cause many
+applications to not work. Install a font package to fix this.
+
 ### Video drivers
 
 Both GNOME and KDE Plasma have EGLStreams backends for Wayland, which means they


src/config/graphical-session/fonts.md
```diff
@@ -1,5 +1,8 @@
 # Fonts
 
+[xbps-query(1)](https://man.voidlinux.org/xbps-query.1) can be used to list
```
ethamck, line 3: I wanted to link to https://docs.voidlinux.org/xbps/index.html#finding-files-and-packages here but I found many more links that referenced man pages so went with that instead. Going to this page would present the user with an immediate code block to do exactly what they'd be looking for at this point, so I don't know that I prefer the man page. (Should I change it?)

classabbyamp: link the docs page

src/config/graphical-session/wayland.md
```diff
@@ -30,6 +30,9 @@ Void Linux currently packages the following Wayland compositors:
 - labwc: a window-stacking compositor, inspired by Openbox
 - Qtile: a dynamic tiling Wayland compositor (via qtile-wayland)
 
+Some compositors don't depend on any [fonts](./fonts.md), so you have to install
+them manually for most apps to work.
```
classabbyamp, line 34:
```suggestion
them manually for most applications to work.
```

src/config/graphical-session/fonts.md
```diff
@@ -1,5 +1,8 @@
 # Fonts
 
+[xbps-query(1)](../../xbps/index.md#finding-files-and-packages) can be used to
+list font-related packages. Many fonts are packaged with the `font-` prefix.
```
classabbyamp, line 4:
```suggestion
A number of fonts and font collections are [available from
XBPS](../../xbps/index.md#finding-files-and-packages). If you are unsure which
fonts to install, `dejavu-fonts-ttf` and `xorg-fonts` are good basic options.
`noto-fonts-ttf` provides fonts that contain characters for many languages and
scripts. `noto-fonts-cjk` provides fonts for Chinese, Japanese, and Korean
characters. `noto-fonts-emoji` provides emojis. `nerd-fonts` provides a number
of fonts with special characters added.

Fonts not available from XBPS can be installed manually to `/usr/share/fonts`
(system-wide) or `~/.local/share/fonts` (per-user).
```

src/config/graphical-session/wayland.md
```diff
@@ -30,6 +30,9 @@ Void Linux currently packages the following Wayland compositors:
 - labwc: a window-stacking compositor, inspired by Openbox
 - Qtile: a dynamic tiling Wayland compositor (via qtile-wayland)
 
+Some compositors don't depend on any [fonts](./fonts.md), so you have to install
+them manually for most applications to work.
```
classabbyamp, line 34:
```suggestion
Some compositors do not depend on any [fonts](./fonts.md), which can cause many
applications to not work. Install a font package to fix this issue.
```

src/config/graphical-session/fonts.md
ethamck: Is (from line 4 to 9)

```diff
+ XBPS](../../xbps/index.md#finding-files-and-packages). `dejavu-fonts-ttf` or `xorg-fonts`
+ are a good baseline if you're unsure of what to pick.
+ `noto-fonts-ttf` contains fonts for many languages and scripts.
+ `noto-fonts-cjk` extends this with fonts for Chinese, Japanese, and Korean, and
+ `noto-fonts-emoji` provides emojis.
+ `nerd-fonts` provides a number of fonts with special characters like custom icons included.
+ Many other font packages begin with `font-`.
```

(I'm fine dropping the last line if it's unnecessary)

adequate or would you prefer that suggestion verbatim? It's pedantic but I find this wording more varied and a bit easier to understand. (Written without ensuring correct column wrapping.)

classabbyamp: my thinking is it's reasonable for the user to understand that `xbps-query - Rs font` would be how you find fonts, so we don't have to write it out.

classabbyamp (approved):

PR #815, opened 2024-09-27
config/containers-and-vms/lxc: document incus

duskmoss:
Incus is the current community tool for easy LXC administration. LXD has been moved to canonical, relicensed (Apache 2.0 to AGPL3) and contributors are asked to sign a CLA. 

LXD package in repos is out of date, but incus is up to date. I thought we should have docs for this package. 

For now I've left the LXD section in as ahesford recommended on IRC.

### DIFF
--- a/src/config/containers-and-vms/lxc.md
+++ b/src/config/containers-and-vms/lxc.md
@@ -1,11 +1,16 @@
 # LXC
 
-The [Linux Containers project](https://linuxcontainers.org/) includes three
-subprojects: [LXC](https://linuxcontainers.org/lxc/introduction/),
-[LXD](https://linuxcontainers.org/lxd/introduction/) and
-[LXCFS](https://linuxcontainers.org/lxcfs/introduction/). The project also
-included the CGManager project, which has been deprecated in favor of the CGroup
-namespace in recent kernels.
+The [Linux Containers project](https://linuxcontainers.org/) includes four
+subprojects: [Incus](https://linuxcontainers.org/incus/introduction/),
+[LXC](https://linuxcontainers.org/lxc/introduction/),
+[LXCFS](https://linuxcontainers.org/lxcfs/introduction/) and
+[distrobuilder](https://linuxcontainers.org/distrobuilder/introduction/).
+
+The project also formerly included the CGManager and
+[LXD](https://canonical.com/lxd) projects. CGManager has been deprecated in
+favor of the CGroup namespace in recent kernels. LXD has become a Canonical
+project. Incus was forked from LXD to be a community driven alternative, and is
+led and maintained by many of the original creators.
 
 ## Configuring LXC
 
@@ -114,6 +119,47 @@ access to other containers. To isolate containers from each other, alter the
 create each container. Trying to fix permissions on a container created with the
 wrong map is possible, but inconvenient.
 
+## Incus
+
+Incus provides an alternative interface to LXC's `lxc-*` utilities. However, it
+does not require the configuration described in [the previous section](#lxc).
+
+In `/etc/rc.conf`, set the `CGROUP_MODE` variable to `unified`. Install the
+`incus` package, and [enable](../services/index.md#enabling-services) the
+`incus` service.
+
+Some parts of Incus require optional dependencies, see
+[README.voidlinux](../package-documentation/index.md).
+
+Add users who should have full control over Incus to the `_incus-admin` group.
+
+Optionally, some users can be given limited access to Incus as described
+[here](https://linuxcontainers.org/incus/docs/main/howto/projects_confine/#confine-projects-to-specific-incus-users).
+Add these users to the `_incus` group, and enable the `incus-user` service.
+
+Note that `incus-user` will initialize the default Incus profile when it is
+started. To avoid default configuration initialize Incus for yourself with
+`incus admin init` before enabling `incus-user`.
+
+> Warning: `incus-user` will also replace the networking config of the default
+> profile if it appears invalid. runit does not have socket activation so this
+> happens when when the machine boots instead of when a user calls `incus`. If
+> the default profile uses a [bridge
+> interface](../network/index.md#bridge-interfaces), `incus-user` may replace it
+> on boot.
+
+To migrate existing LXD setups to Incus, use the `lxd-to-incus` tool from the
+`incus-tools` package as described
+[here](https://linuxcontainers.org/incus/docs/main/howto/server_migrate_lxd/).
+
+To migrate existing LXC containers to Incus, use the `lxc-to-incus` script from
+the `incus-tools` package as described
+[here](https://linuxcontainers.org/incus/docs/main/howto/migrate_from_lxc/).
+
+Some Incus features have additional dependencies. To run virtual machines,
+install `qemu` and `edk2-ovmf`. To run OCI containers, install `skopeo` and
+`umoci`.
+
 ## LXD
 
 LXD provides an alternative interface to LXC's `lxc-*` utilities. However, it
@@ -125,4 +171,4 @@ the `lxd` service.
 LXD users must belong to the `lxd` group.
 
 Use the `lxc` command to manage instances, as described
-[here](https://linuxcontainers.org/lxd/getting-started-cli/#lxd-client).
+[here](https://documentation.ubuntu.com/lxd/en/latest/howto/).


src/config/containers-and-vms/lxc.md
```diff
@@ -114,6 +119,40 @@ access to other containers. To isolate containers from each other, alter the
 create each container. Trying to fix permissions on a container created with the
 wrong map is possible, but inconvenient.
 
+## Incus
+
+Incus provides an alternative interface to LXC's `lxc-*` utilities. However, it
+does not require the configuration described in [the previous section](#lxc).
+
+Edit `/etc/rc.conf` and set the `CGROUP_MODE` variable to `unified`. Install the
```
classabbyamp, line 127:
```suggestion
In `/etc/rc.conf`, set the `CGROUP_MODE` variable to `unified`. Install the
```

src/config/containers-and-vms/lxc.md
```diff
@@ -114,6 +119,40 @@ access to other containers. To isolate containers from each other, alter the
...
+`incus` package, and [enable](../services/index.md#enabling-services) the
+`incus` service.
+
+Incus users must be belong to the `_incus-admin` group.
```
classabbyamp, line 131: what are "incus users"? users that can perform admin actions in incus?

src/config/containers-and-vms/lxc.md
```diff
@@ -114,6 +119,40 @@ access to other containers. To isolate containers from each other, alter the
...
+
+Optionally enable some users to have limited acces to incus as described
```
classabbyamp, line 133:
```suggestion
Optionally, some users can be given limited access to Incus as described
```

src/config/containers-and-vms/lxc.md
```diff
@@ -114,6 +119,40 @@ access to other containers. To isolate containers from each other, alter the
...
+[here](https://linuxcontainers.org/incus/docs/main/howto/projects_confine/#confine-projects-to-specific-incus-users).
+Add users to the `_incus` group, and enable the `incus-user` service
```
classabbyamp, line 135:
```suggestion
Add users to the `_incus` group, and enable the `incus-user` service.
```

src/config/containers-and-vms/lxc.md
```diff
@@ -114,6 +119,40 @@ access to other containers. To isolate containers from each other, alter the
...
+Note that `incus-user` will initialize the default incus profile when it is
+started. To avoid default configuration initialize incus for youself with `incus
+admin init` before enabling `incus-user`.
+
+**WARNING:** `incus-user` will also replace the networking config of the default
+profile if it appears invalid. runit does not have socket activation so this
+happens when when the machine boots instead of when a user attempts to access
+it. If the default profile uses a [bridge
+interface](../network/index.md#bridge-interfaces) `incus-user` may replace it on
+boot.
```
classabbyamp, line 146:
```suggestion
> Warning: `incus-user` will also replace the networking config of the default
> profile if it appears invalid. runit does not have socket activation so this
> happens when when the machine boots instead of when a user attempts to access
> it. If the default profile uses a [bridge
> interface](../network/index.md#bridge-interfaces) `incus-user` may replace it on
> boot.
```

src/config/containers-and-vms/lxc.md
```diff
@@ -114,6 +119,40 @@ access to other containers. To isolate containers from each other, alter the
...
+
+To migrate existing LXD setups to incus use the 'lxd-to-incus' binary from the
```
classabbyamp, line 148:
```suggestion
To migrate existing LXD setups to Incus, use the `lxd-to-incus` tool from the
```

src/config/containers-and-vms/lxc.md
```diff
@@ -114,6 +119,40 @@ access to other containers. To isolate containers from each other, alter the
...
+`incus-tools` package as described
+[here](https://linuxcontainers.org/incus/docs/main/howto/server_migrate_lxd/).
+
+To migrate existing LXC containers to incus use the 'lxc-to-incus' binary from
```
classabbyamp, line 152:
```suggestion
To migrate existing LXC containers to Incus use the `lxc-to-incus` script from
```

classabbyamp (commented):
mostly copyediting review as I don't know much about incus

src/config/containers-and-vms/lxc.md
0x5c: Upstream docs do say that the `_incus-admin` group should only be given to users who should be able to do system-wide incus operations without root. The normal recommended way to use incus as a normal user (other than setup of the daemon) is with the plain `_incus` group.

src/config/containers-and-vms/lxc.md
```diff
@@ -114,6 +119,40 @@ access to other containers. To isolate containers from each other, alter the
...
+`incus` service.
+
+Incus users must be belong to the `_incus-admin` group.
+
+Optionally enable some users to have limited acces to incus as described
+[here](https://linuxcontainers.org/incus/docs/main/howto/projects_confine/#confine-projects-to-specific-incus-users).
+Add users to the `_incus` group, and enable the `incus-user` service
+
+Note that `incus-user` will initialize the default incus profile when it is
+started. To avoid default configuration initialize incus for youself with `incus
```
0x5c, line 138:
```suggestion
started. To avoid default configuration initialize incus for yourself with `incus
```

0x5c (commented):
> Note that `incus-user` will initialize the default incus profile when it is started. To avoid default configuration initialize incus for youself with `incus admin init` before enabling `incus-user`.

I spent multiple days trying to figure out what was going on with incus automatically creating a new zfs pool, thanks for saving me from going mad :D

src/config/containers-and-vms/lxc.md
duskmoss: @classabbyamp yes, that's what I meant by incus users. all actions are admin action in normal operation as it's roughly a hypervisor. Open to rephrasing.

@0x5c I don't think that's a fair characterization The normal way is to have access to everything in the incus admin. adding to just the incus group gets a user access to just a personalized project space as described here, as an optional isolation strategy. 

https://linuxcontainers.org/incus/docs/main/explanation/projects/#projects-confined

Having an admin is essential, having limited access is optional, is my view I wrote into the docs. 


or riffed on briefly here 

https://linuxcontainers.org/incus/docs/main/tutorial/first_steps/

    "incus allows basic user access, no configuration and all actions restricted to a per-user project."

   " incus-admin allows full control over Incus."

0x5c: The [upstream documentation page for install instructions](https://linuxcontainers.org/incus/docs/main/installing/) has this, which points out that only root access is actually needed to perform system-wide operations
> The root user and all members of the incus-admin group can interact with the local daemon.

It also includes
> :fire: Important
> Local access to Incus through the Unix socket always grants full access to Incus. This includes the ability to attach file system paths or devices to any instance as well as tweak the security features on any instance.
> Therefore, you should only give such access to users who you’d trust with root access to your system.

Additionally, the [About Security](https://linuxcontainers.org/incus/docs/main/explanation/security/) page states
> - Restrict access to the Incus daemon and the remote API.
> - Do not use privileged containers unless required. If you use privileged containers, put appropriate security measures in place. See the [LXC security page](https://linuxcontainers.org/lxc/security/) for more information.

The above mentioned LXC Security page has this to say about privileged and unprivileged containers
> The former can be thought as old-style containers, they're not safe at all and should only be used in environments where unprivileged containers aren't available and where you would trust your container's user with root access to the host.

Critically, `_incus-admin` is equivalent to unrestricted passwordless sudo access.

While I agree that the `_incus-admin` group should be mentioned, saying that users "must" be in that group does not sound like a reasonable *default* recommendation, especially when prefixing admin commands with `sudo`/`doas` would do the same without the problem of granting passwordless root access.
In other words, while the Void documentation should not be there to criticise security choices a user could deem safe for their setup, the documentation shouldn't be making the assumption of what's "super safe enough" in what it recommends, especially when upstream is quite clear about the risks.

To me the proper way to handle this is to indicate that root access in needed to perform incus administrative operations, with a mention of `_incus-admin` as an alternative along with its caveats. Further details may possibly be more appropriate for the upstream documentation, which can be linked to.

classabbyamp: that makes sense to me

0x5c: Sound good to me

0x5c:
There's two remaining bits of documentation in the `README.voidlinux` file that if added here would render it completely unnecessary.

> The root user should also be assigned a range of subordinate user and group IDs to be mapped in
containers.
> For example, the command
> ```
> usermod --add-subuids 1000000-1065535 \
> 	--add-subgids 1000000-1065535 root
> ```
> will assign a range of 65,536 IDs for users and groups.

The sub{u,g}id stuff is mentioned vaguely (with different values and the manual method) in the LXC section, but there's no indication that one would need to follow the instruction in the LXC section, nor good indication at to what it means in practice Maybe this could be refactored into a section that both the LXC and Incus sections link to.


> Optional dependencies:
> 
> To run virtual machines, install `qemu` and `edk2-ovmf`.
> To run OCI containers, install `spokeo` and `umoci`.

duskmoss:
> There's two remaining bits of documentation in the `README.voidlinux` file that if added here would render it completely unnecessary.

I didn't include these because it's actually not neccesary, these are setup in void by default. I did make a pr to remove from the readme but maintainer wanted to keep.

0x5c:
Interesting, the only range that was setup without my intervention was for my user, not root. What part of void would be setting that up automatically?

duskmoss:
https://github.com/void-linux/void-packages/blob/master/srcpkgs/base-files/files/subgid

0x5c:
Weirdly, that was added 5 years ago and my install is not that old, I guess something to investigate separately from this issue

duskmoss:
I'll try and confirm at some point, but eitherway, it's in the readme. 

I think optional dependencies should stay in the readme, it's much easier to update them when the package is updated than keeping the documentation in sync. And it's the standard place for void, like with the steam package. 

I'll get the edits talked about here worked on either today or tomorrow.

src/config/containers-and-vms/lxc.md
```diff
@@ -114,6 +119,40 @@ access to other containers. To isolate containers from each other, alter the
...
+## Incus
+
+Incus provides an alternative interface to LXC's `lxc-*` utilities. However, it
+does not require the configuration described in [the previous section](#lxc).
+
+In `/etc/rc.conf`, set the `CGROUP_MODE` variable to `unified`. Install the
+`incus` package, and [enable](../services/index.md#enabling-services) the
+`incus` service.
+
+Add users who should have full control over Incus to the '_incus-admin' group.
```
classabbyamp, line 131:
```suggestion
Add users who should have full control over Incus to the `_incus-admin` group.
```

src/config/containers-and-vms/lxc.md
```diff
@@ -114,6 +119,40 @@ access to other containers. To isolate containers from each other, alter the
...
+
+Optionally, some users can be given limited access to Incus as described
+[here](https://linuxcontainers.org/incus/docs/main/howto/projects_confine/#confine-projects-to-specific-incus-users).
+Add users to the `_incus` group, and enable the `incus-user` service.
```
classabbyamp, line 135:
```suggestion
Add these users to the `_incus` group, and enable the `incus-user` service.
```

src/config/containers-and-vms/lxc.md
```diff
@@ -114,6 +119,40 @@ access to other containers. To isolate containers from each other, alter the
...
+
+Note that `incus-user` will initialize the default Incus profile when it is
+started. To avoid default configuration initialize Incus for yourself with
+`incus admin init` before enabling `incus-user`.
+
+> Warning: `incus-user` will also replace the networking config of the default
+> profile if it appears invalid. runit does not have socket activation so this
+> happens when when the machine boots instead of when a user calls `incus`. If
+> the default profile uses a [bridge
+> interface](../network/index.md#bridge-interfaces) `incus-user` may replace it
```
classabbyamp, line 145:
```suggestion
> interface](../network/index.md#bridge-interfaces) that `incus-user` may replace
```

src/config/containers-and-vms/lxc.md
```diff
@@ -114,6 +119,40 @@ access to other containers. To isolate containers from each other, alter the
...
+> on boot.
+
+To migrate existing LXD setups to Incus, use the `lxd-to-incus` tool from the
+`incus-tools` package as described
+[here](https://linuxcontainers.org/incus/docs/main/howto/server_migrate_lxd/).
+
+To migrate existing LXC containers to Incus use the `lxc-to-incus` script from
```
classabbyamp, line 152:
```suggestion
To migrate existing LXC containers to Incus, use the `lxc-to-incus` script from
```

src/config/containers-and-vms/lxc.md
```diff
@@ -114,6 +119,40 @@ access to other containers. To isolate containers from each other, alter the
...
 wrong map is possible, but inconvenient.
 
+## Incus
+
+Incus provides an alternative interface to LXC's `lxc-*` utilities. However, it
+does not require the configuration described in [the previous section](#lxc).
+
+In `/etc/rc.conf`, set the `CGROUP_MODE` variable to `unified`. Install the
+`incus` package, and [enable](../services/index.md#enabling-services) the
+`incus` service.
```
classabbyamp, line 129: maybe we could mention looking in README.voidlinux for optional dependencies? like `Some parts of Incus require optional dependencies, see [README.voidlinux](../package-documentation/index.md).` (link might not be 100% right)

duskmoss: Sounds good

src/config/containers-and-vms/lxc.md
duskmoss: I don't think this reads correctly.  "If there is a a bridge that incus may replace on boot" isn't a complete sentence.

classabbyamp: oops, got messed up by link/line breaks. maybe a comma after the link instead:
```suggestion
> interface](../network/index.md#bridge-interfaces), `incus-user` may replace it
```

classabbyamp:
> I think optional dependencies should stay in the readme, it's much easier to update them when the package is updated than keeping the documentation in sync.

a pr to void-docs updating that kind of thing can be merged quickly, but it is also valuable to have them in README.voidlinux for history.

Can we put them in both places?

duskmoss:
Sure

classabbyamp:
and btw, commit message should be `config/containers-and-vms/lxc: add incus`

duskmoss:
> and btw, commit message should be `config/containers-and-vms/lxc: document incus` or something like that

Thanks! Maybe the contribution docs should be more clear about what you want for <filename> because path, and file name with  extension removed is not something I would have gotten from <filename>

PR #817, opened 2024-10-23
Fix typo: system -> service

eightbitraptor:
This PR corrects what I think is a typo in the services documentation.

To enable a service we link it directly into the default runsvdir if _the service_ is not currently running. If the _system_ is not currently running, then enabling services becomes a bit harder.

### DIFF
--- a/src/config/services/index.md
+++ b/src/config/services/index.md
@@ -125,7 +125,7 @@ directory in `/var/service/`:
 # ln -s /etc/sv/<service> /var/service/
 ```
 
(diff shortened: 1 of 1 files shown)


classabbyamp:
system is correct. if the void install is not booted (e.g. via chroot), /var/service won't exist (as it is a symlink created during boot)

eightbitraptor:
@classabbyamp Thanks for the clarification. Apologies for the confusion.

Issue #821, opened 2024-11-24
Locales and Translations section of the handbook should mention it's glibc-specific


grtcdr:
I was perusing the handbook yesterday and noticed that it was suggested in https://docs.voidlinux.org/config/locales.html to use the `locale` command to list currently enabled locales.

I was confused when that command wasn't available and suspected it was glibc-specific. I spoke with a few individuals on the `#voidlinux` and user "aedinius" mentioned that `musl` does not support locales which confirmed my suspicion. User "aninternettroll" suggested that the page should probably clarify that it is specific to `musl`. I also agree. What do you think?

PR #822, opened 2024-11-24
Mention that Locales and Translations section is specific to glibc

grtcdr:
This pull request closes issue #821.

### DIFF
--- a/src/config/locales.md
+++ b/src/config/locales.md
@@ -1,6 +1,11 @@
 # Locales and Translations
 
-For a list of currently enabled locales, run
+`glibc` supports setting the system locale whereas `musl` does not; both support
+setting the language for applications.
+
+## Listing locales
+
+For a list of currently enabled locales, run:
 
 ```
 $ locale -a
@@ -12,7 +17,7 @@ To enable a certain locale, un-comment or add the relevant lines in
 `/etc/default/libc-locales` and [force-reconfigure](../xbps/index.md) the
 `glibc-locales` package.
 
-## Setting the system locale
+## Setting the system language
 
 Set `LANG=xxxx` in `/etc/locale.conf`.
 


abenson:
I'd explain that `musl` doesn't support locales.

grtcdr:
I'll do that right away.

grtcdr:
Would you consider https://wiki.gnome.org/Initiatives(2f)Wayland(2f)Applications.html a reasonable replacement for the currently broken GNOME link?

Here's an archived version of that URL for comparison: https://web.archive.org/web/20240620195313/https://wiki.gnome.org/Initiatives/Wayland/GTK%2B

classabbyamp:
setting `LANG` should still work on musl for applications, iirc

grtcdr:
Does that environment variable matter though? https://wiki.musl-libc.org/environment-variables doesn't list it.

grtcdr:
@classabbyamp: I phrased things differently to take into account the fact that `musl` does indeed support `LANG`. It would be great if the musl documentation had anchors for the different headings because then we could reference that section.

classabbyamp:
musl itself doesn't support `LANG`, but most applications will use `LANG` to set the language, even in the absence of system locale support (like on musl)

classabbyamp:
`/etc/locale.conf` is just sourced by a `profile.d` snippet (`/etc/profile.d/locale.sh`) so it should apply regardless

grtcdr:
I understand, I said it wasn't required because it apparently isn't. That statement does not negate yours, though.

src/config/locales.md
```diff
@@ -14,7 +16,8 @@ To enable a certain locale, un-comment or add the relevant lines in
 
 ## Setting the system locale
```
classabbyamp, line 17: this heading is inaccurate, it's for setting the system *language*

src/config/locales.md
```diff
@@ -1,6 +1,8 @@
 # Locales and Translations
 
-For a list of currently enabled locales, run
+You can use the following command, which is specific to `glibc` installations of
+Void (given that `musl` does not support locales), to see a list of currently
+enabled locales:
```
classabbyamp, line 5: personally, I think this is better as an introductory sentence/paragraph, so something like this

```suggestion
glibc supports setting the system locale, while musl does not. Both support setting the language for applications.

## Listing locales

For a list of currently enabled locales, run:
```

src/config/locales.md
grtcdr: Fixed.

src/config/locales.md
grtcdr: I like that too, fixed.

src/config/locales.md
```diff
@@ -12,9 +17,10 @@ To enable a certain locale, un-comment or add the relevant lines in
 `/etc/default/libc-locales` and [force-reconfigure](../xbps/index.md) the
 `glibc-locales` package.
 
-## Setting the system locale
+## Setting the system language
 
-Set `LANG=xxxx` in `/etc/locale.conf`.
+Set `LANG=xxxx` in `/etc/locale.conf`, this applies to `glibc` and `musl`
+although it's not required by the latter.
```
classabbyamp, line 23: it is required for glibc, if you want the language to change for most applications

grtcdr: I never said it wasn't though, did you mean to say musl instead of glibc?

classabbyamp: i read latter as former, but `LANG` is required for setting the system language on both libcs

grtcdr: I do that too sometimes, I'll remove that part altogether.

grtcdr: Okay, done.

src/config/locales.md
```diff
@@ -1,6 +1,11 @@
 # Locales and Translations
 
-For a list of currently enabled locales, run
+`glibc` supports setting the system locale whereas `musl` does not, both support
```
classabbyamp, line 3:
```suggestion
`glibc` supports setting the system locale whereas `musl` does not; both support
```
nit

classabbyamp (commented):
please squash your commits into 1 with all these changes, then this should be good to go

grtcdr:
Okay, all set :)

PR #827, opened 2024-12-23
add apple silicon section

dkwo:
Moved from the readme as discussed.

To do: Once this is merged, add a reference to it in asahi wiki/alt distros.

### DIFF
--- a/src/installation/guides/arm-devices/apple-silicon.md
+++ b/src/installation/guides/arm-devices/apple-silicon.md
@@ -1,0 +1,50 @@
+# Apple Silicon
+
+Void's Apple Silicon support is based on Asahi Linux. See their
+[website](https://asahilinux.org) and
+[documentation](https://github.com/AsahiLinux/docs/wiki) for more information.
+
+## Installation
+
+Before installing, use the Asahi Linux install script to install "UEFI
+environment only" from macOS:
+
+```
+macos $ curl https://alx.sh > alx.sh
+macos $ sh ./alx.sh
+```
+
+Then, [create a Live USB](../../live-images/prep.md) using an [Apple Silicon
+Void Linux ISO](https://voidlinux.org/download/#arm%20platforms). U-Boot
+(installed by the Asahi installer) should show the external USB as a boot
+option. If it does not, run these commands in the U-Boot prompt to boot:
+
+```
+U-Boot> setenv boot_targets "usb"
+U-Boot> setenv bootmeths "efi"
+U-Boot> boot
+```
+
+To install, follow the [chroot install guide](../chroot.md), using the "XBPS
+method", observing the following modifications:
+
+For the base installation, install `base-system`, `asahi-base`, and
+`asahi-scripts`. These packages provide important configurations and install the
+necessary dependencies. When running `grub-install`, add the `--removable` flag.
+
+To use another bootloader with `m1n1`, ensure it installs the bootloader EFI
+executable at `EFI\BOOT\BOOTAA64.EFI` within the EFI system partition. `m1n1`
+can also be configured to boot a kernel and initramfs directly, without a
+bootloader. To do this, change the `PAYLOAD` in `/etc/default/m1n1-kernel-hook`
+to `kernel`.
+
+## Audio
+
+The `asahi-audio` package is required for audio. Ensure the speakersafetyd
+service is [enabled](../../../config/services/index.md#enabling-services), and
+set up [pipewire and wireplumber](../../../config/media/pipewire.md).
+
+## Firmware
+
+Firmware can be updated with `asahi-fwupdate` from `asahi-scripts`. It is
+recommended to do so whenever the `asahi-firmware` package is updated.
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -15,6 +15,8 @@
       - [Full Disk Encryption](./installation/guides/fde.md)
       - [Root on ZFS](./installation/guides/zfs.md)
       - [ARM Devices](./installation/guides/arm-devices/index.md)
+         - [Apple Silicon
+            (Asahi)](./installation/guides/arm-devices/apple-silicon.md)
          - [Lenovo Thinkpad
             X13s](./installation/guides/arm-devices/thinkpad-x13s.md)
          - [Pinebook Pro](./installation/guides/arm-devices/pinebook-pro.md)
--- a/src/installation/guides/arm-devices/index.md
+++ b/src/installation/guides/arm-devices/index.md
@@ -18,6 +18,7 @@ specific to such devices.
 
 Platform-specific documentation is available for:
 
+- [Apple Silicon](./apple-silicon.md)
 - [Lenovo ThinkPad X13s](./thinkpad-x13s.md)
 - [Pinebook Pro](./pinebook-pro.md)
 - [Raspberry Pi](./raspberry-pi.md)


dkwo:
cc @classabbyamp

src/installation/guides/arm-devices/platforms.md
```diff
@@ -110,3 +110,29 @@ default](https://github.com/raspberrypi/linux/commit/28aec65bb1743c9bfa53b036999
...
+are required, unless one changes initramfs or payload for m1n1.
+
+Void ships a kernel hook for m1n1, modeled upon Asahi upstream. This allows to
+either just use m1n1, or use it in combination with uboot and grub (default) by
+setting the `PAYLOAD` variable (requires `asahi-uboot`, `grub-arm64-efi`). We
+ship tinyramfs config and hooks (to use, add `asahi` to `hooks`). The default is
+dracut, its hooks are provided by `asahi-scripts` together with mkinitcpio ones.
+
+If using audio, you must install `asahi-audio`, ensure the speakersafetyd
+service is enabled, and utilize pipewire and wireplumber.
```
classabbyamp, line 135:
```suggestion
service is [enabled](/config/services/index.md#enabling-services), and set up [pipewire and wireplumber](/config/media/pipewire.md).
```

src/installation/guides/arm-devices/platforms.md
```diff
@@ -110,3 +110,29 @@ default](https://github.com/raspberrypi/linux/commit/28aec65bb1743c9bfa53b036999
 This breaks workloads which use containers. Therefore, if you want to use
 containers on your Raspberry Pi, you need to enable memory cgroups by adding
 `cgroup_enable=memory` to `/boot/cmdline.txt`.
+
+## Apple Silicon
+
+For some time, the kernel, mesa and uboot packages will have Asahi Linux as
```
classabbyamp, line 116:
```suggestion
For some time, the kernel, mesa, and uboot packages will have Asahi Linux as
```

src/installation/guides/arm-devices/platforms.md
```diff
@@ -110,3 +110,29 @@ default](https://github.com/raspberrypi/linux/commit/28aec65bb1743c9bfa53b036999
...
+upstream. Their [documentation](https://github.com/AsahiLinux/docs/wiki) is an
+excellent source. Here we discuss Void Linux specific aspects only.
+
+Use Asahi Linux install scripts with `UEFI environment only`, then proceed with
+usual Void installation. The default is EFI partition at `/boot/efi`, other
+choices can be made by editing m1n1 kernel hook config.
+
+At the very least, install `base-system`, `asahi-base` and an initramfs
```
classabbyamp, line 124:
```suggestion
At the very least, install `base-system`, `asahi-base`, and an initramfs
```

src/installation/guides/arm-devices/platforms.md
```diff
@@ -110,3 +110,29 @@ default](https://github.com/raspberrypi/linux/commit/28aec65bb1743c9bfa53b036999
...
+generator. By default, also `asahi-uboot`, `grub-arm64-efi` and `asahi-scripts`
+are required, unless one changes initramfs or payload for m1n1.
```
classabbyamp, line 126:
```suggestion
generator. By default, `asahi-uboot`, `grub-arm64-efi`, and `asahi-scripts`
are also required, unless one changes initramfs or payload for m1n1.
```

src/installation/guides/arm-devices/platforms.md
```diff
@@ -110,3 +110,29 @@ default](https://github.com/raspberrypi/linux/commit/28aec65bb1743c9bfa53b036999
...
+
+Void ships a kernel hook for m1n1, modeled upon Asahi upstream. This allows to
```
classabbyamp, line 128:
```suggestion
Void ships a kernel hook for m1n1, modeled on Asahi upstream. This allows you to
```

src/installation/guides/arm-devices/platforms.md
```diff
@@ -110,3 +110,29 @@ default](https://github.com/raspberrypi/linux/commit/28aec65bb1743c9bfa53b036999
...
+either just use m1n1, or use it in combination with uboot and grub (default) by
+setting the `PAYLOAD` variable (requires `asahi-uboot`, `grub-arm64-efi`). We
```
classabbyamp, line 130:
```suggestion
setting the `PAYLOAD` variable (requires `asahi-uboot` and `grub-arm64-efi`). We
```

src/installation/guides/arm-devices/platforms.md
```diff
@@ -110,3 +110,37 @@ default](https://github.com/raspberrypi/linux/commit/28aec65bb1743c9bfa53b036999
...
+upon Asahi upstream. This allows you to either just use m1n1, or use it in
+combination with uboot and grub by setting the `PAYLOAD` variable (requires
+`asahi-uboot` and `grub-arm64-efi`).
+
+If using audio, you must install `asahi-audio`, ensure the speakersafetyd
+service is [enabled](/config/services/index.md#enabling-services), and set up
+[pipewire and wireplumber](/config/media/pipewire.md).
+
+Firmware can be updated with `asahi-fwupdate` from `asahi-scripts`. It is
+recommended to do so whenever the asahi-firmware package is updated.
```
Calandracas606, line 146: This is not true.

`asahi-firmware` does not contain any firmware, all the firmware is extracted from MacOS. These scripts are used either in the initramfs, or for extracting the firmware from MacOS by the installer script

dkwo: Once again: the initramfs copies the files, but asahi-firmware does essentially what the install script does. It was recommended to me on irc than when asahi-fwupdate package is updated, this could signal that they are bringing in new firmware, hence the suggestion to run the command.

Calandracas606: I don't think that it's possible to bring in new firmware, but if this is really the case, shouldn't it be run as a post install hook instead of making the user do it manually?

dkwo: Maybe my wording in unclear. I just meant that whenever that package is updated, the user may want to run that script. Running automatically by trigger may be also an option.

dkwo: See e.g. https://github.com/AsahiLinux/docs/wiki/Open-OS-Ecosystem-on-Apple-Silicon-Macs#packaged-blobs (as well as some irc conversation I cannot find now): there is definitely a possibility that new blobs are going to be used, and when this is the case, it will be implemented through an update of asahi-installer/asahi-firmware pkg. So maybe not all updates will require running fwupdate, but it is still advisable to do so.

dkwo: This is the ref to irc: https://oftc.irclog.whitequark.org/asahi/2023-04-11#32057960;

> the used OS would be responsible for providing asahi-fwextract and running it on upgrades 

I am happy to either recommend it (as is now) or use a trigger.

classabbyamp:
rewritten to match void's documentation style

src/installation/guides/arm-devices/apple-silicon.md
```diff
@@ -0,0 +1,42 @@
...
+U-Boot (installed by the Asahi installer) should show the external USB as a boot
+option. If it does not, run these commands in the U-Boot prompt to boot:
+
+```
+setenv boot_targets "usb"
+setenv bootmeths "efi"
+boot
+```
+
+will instruct it to do so. Then proceed with the usual Void installation.
```
dkwo, line 22: maybe delete 'will instruct to do so' ?

src/installation/guides/arm-devices/apple-silicon.md
```diff
@@ -0,0 +1,42 @@
...
+
+To install, follow the [chroot install guide](../chroot.md), using the "XBPS
+method", observing the following modifications:
+
+For the base installation, install `base-system`, `asahi-base`, and
+`asahi-scripts`. This package provides important configurations and installs the
```
dkwo, line 28: Maybe these packages provide ... and install? Since it is both asahi-base and -scripts.

src/installation/guides/arm-devices/apple-silicon.md
```diff
@@ -0,0 +1,42 @@
...
+necessary dependencies. When running `grub-install`, the `--removable` flag may
+be required. If using another bootloader, you may want to change the m1n1
+`PAYLOAD` in `/etc/m1n1.conf`.
```
dkwo, line 31: That should be `/etc/default/m1n1-kernel-hook`.
You could also add something like: 

> Edit `/etc/m1n1.conf` to change options for the m1n1 bootloader.

src/installation/guides/arm-devices/platforms.md
classabbyamp: pretty sure jannau meant *macOS* upgrades, not upgrades of a package on the linux side, if the script extracts fw from macOS

dkwo: If you read the lines a bit above and below the linked one, it is already explained that they do _not_ mean upgrades of macos. The rationale is that they change/upgrade `asahi-installer` (which is where we get `asahi-firmware`) whenever they discover they need to provide additional firmware in `/boot/vendorfw/firmware.{cpio,tar.gz}`.

classabbyamp:
in what cases would you want to use the kernel payload? even if not using grub, it seems that you'd typically want to use u-boot with whichever bootloader/uki

also, is `--removable` for grub required or recommended?

Calandracas606:
> also, is `--removable` for grub required or recommended?

Required. There is no access to efivars, and m1n1 only knows how to execute the hardcoded `$ESP\EFI\BOOT\BOOTAA64.EFI`

dkwo:
m1n1 can take kernel+initramfs+dtbs as payload, that's the meaning of `payload=kernel`: this does not use grub, nor uboot, and does not require any hardcoded `$ESP\EFI\BOOT\BOOTAA64.EFI` or similar. @Calandracas606 Not sure why you think that. (dtbs are taken from kernel btw, not uboot). m1n1 is the bootloader, in some sense.

As for grub, I only meant that `--removable` works for me. Did not mean it is required. I'm not an experienced grub user, but by simply looking at https://github.com/AsahiLinux/asahi-scripts/blob/main/update-grub in asahi-scripts it seems clear that one can configure grub properly, without using `--removable` (at least it's not mentioned there).

dkwo:
Quick question: currently on these machines
```
$ xuname 
Void 6.12.10+2-asahi_2 aarch64 Unknown uptodate rFFFF
```
Would it make sense to use something like
```
$ cat /proc/device-tree/model 
Apple MacBook Air (M1, 2020)
```
for CPU?

classabbyamp:
bring it up in https://github.com/leahneukirchen/xtools/pull/227 i guess

classabbyamp:
> (at least it's not mentioned there).

https://github.com/AsahiLinux/asahi-scripts/blob/main/update-grub#L28
https://github.com/AsahiLinux/asahi-scripts/blob/main/update-grub#L77

this is equivalent to `grub-install --removable`

dkwo:
Thanks for clarifying that they are equivalent to `--removable`.
The current text looks good to me.

PR #828, opened 2024-12-29
Sort kernels by version

ka-xa-ta:


src/config/kernel.md
```diff
@@ -7,7 +7,7 @@ named `linux<x>.<y>`: for example, `linux4.19`. You can query for all available
 kernel series by running:
 
 ```
-$ xbps-query --regex -Rs '^linux[0-9.]+-[0-9._]+'
+$ xbps-query --regex -Rs '^linux[0-9.]+-[0-9._]+' | sort -Vrk2 -t" "
```
classabbyamp, line 10: just `-Vk2` should be enough

PR #830, opened 2025-01-12
release for 2025Q1 images

classabbyamp:


classabbyamp:
merge into master when promoting images, so docs.voidlinux.org is up-to-date

PR #831, opened 2025-01-27
Weston no longer has xwayland subpkg.

Calandracas606:
Weston no longer has xwayland subpkg; It has been rolled into the main package.


classabbyamp:
ab5360d

PR #841, opened 2025-05-04
Update mdbook

### LINKED ISSUE #840: Building documentation using recent versions of mdBook is slightly broken
This has been discussed in `#voidlinux`. Here is the original screen capture I posted there which showcases some of the problems:

https://github.com/user-attachments/assets/b90cbd47-aa59-4d7f-8afd-28b9919b9780

Trying to build void-docs with a recent version of mdBook (including v0.4.43 which is the version currently packaged in void-packages) will lead to the following quirks not present in https://docs.voidlinux.org/:

1. Sidebar's links become broken (see https://github.com/rust-lang/mdBook/issues/2685) (this is probably the most severe problem affecting the usability of the documentation)
2. The selected chapter is not highlighted like in https://docs.voidlinux.org/:

   sidebar with a selected chapter
3. There are probably more issues (more minor or specific ones)

Local builds **and** the `void-docs` official package are affected.

# Cause
This is caused by void-docs' custom theme which completely overrides mdBook's default sidebar logic. Void's substitute code relies on [mdBook's toc helper](https://rust-lang.github.io/mdBook/format/theme/index-hbs.html#1-toc), which endured a breaking change in newer versions of mdBook: https://github.com/rust-lang/mdBook/issues/2685

# Relevant mdBook versions
void-docs' theme resides in `src/theme/`. Most of it was overhauled by 7281cedcdd99294c93d71f4ef33953bea22cc819. The commit description indicates that these files were made by taking mdBook's default theme files and modifying them to Void's needs.

The void-packages packaged version of mdBook was v0.4.14 at the time of making that commit, so mdBook v0.4.14 default theme files were most likely used as base for void-docs theme.

The [Void infrastructure](https://github.com/void-linux/void-infrastructure/blob/032b72320089cb7ab7ea657d5b2d58cc7a9efafc/services/nomad/apps/void-docs.nomad) generates https://docs.voidlinux.org/ from the [`ghcr.io/void-linux/infra-mdbook:20231023R1`](https://github.com/void-linux/void-infrastructure/pkgs/container/infra-mdbook/140322358?tag=20231023R1) container (which is created from [here](https://github.com/void-linux/void-infrastructure/blob/032b72320089cb7ab7ea657d5b2d58cc7a9efafc/services/pkg/docker-bake.hcl#L57-L60) using [these files](https://github.com/void-linux/void-infrastructure/tree/032b72320089cb7ab7ea657d5b2d58cc7a9efafc/services/pkg/mdbook)). The `ghcr.io/void-linux/infra-mdbook:20231023R1` container includes mdBook v0.4.35, which doesn't suffer from the problem described in https://github.com/rust-lang/mdBook/issues/2685

# Void's changes to mdBook's default theme
Here is a cleaned up diff (I had to reindent void-docs' theme) between mdBook v0.4.14 default theme files and void-docs' current theme files: [void-docs-theme-changes.patch.txt](https://github.com/user-attachments/files/20027290/void-docs-theme-changes.patch.txt)

# Possible solutions
The most straight forward solution to the sidebar problem would be to somehow convince the relative links generated in [here](https://github.com/void-linux/void-docs/blob/11d333092eb6badb2d3d3296a67d48518081bdd6/src/theme/index.hbs#L147-L149) by [`{{#toc}}{{/toc}}`](https://rust-lang.github.io/mdBook/format/theme/index-hbs.html#1-toc) to be relative to `path_to_root` or to something similar. I personally do not know how to do that or whether it is possible.

Another solution would be to switch to mdBook's default sidebar logic. This would lead to less custom theme logic in void-docs. This would result in the following behavior changes:

1. The sidebar would become scrollable, it would no longer be static.

   Most of the other features mentioned here can be relatively easily removed, but preserving the static nature of the sidebar would require heavy modification to mdBook's default sidebar logic (I assume, I haven't tried implementing it). In recent enough versions of mdBook, the sidebar is handled by

   ```hbs
   <!DOCTYPE HTML>
   <html lang="{{ language }}" class="{{ default_theme }} sidebar-visible" dir="{{ text_direction }}">
       <head>
           <!-- ... -->
   
           <!-- Start loading toc.js asap -->
           <script src="{{ resource "toc.js" }}"></script>
       </head>
       <body>
       <div id="body-container">
           <!-- ... -->
   
           <nav id="sidebar" class="sidebar" aria-label="Table of contents">
               <!-- populated by js -->
               <mdbook-sidebar-scrollbox class="sidebar-scrollbox"></mdbook-sidebar-scrollbox>
               <noscript>
                   <iframe class="sidebar-iframe-outer" src="{{ path_to_root }}toc.html"></iframe>
               </noscript>
               <div id="sidebar-resize-handle" class="sidebar-resize-handle">
                   <div class="sidebar-resize-indicator"></div>
               </div>
           </nav>
   
           <!-- ... -->
       </div>
       </body>
   </html>
   ```

   A JS version is managed by `toc.js` and a non-JS one is built by mdBook and outputted into `toc.html`.

   The `toc.js` file would likely have to be modified to preserve the original UI. This file is not [one of the files overrideable by mdBook theme](https://rust-lang.github.io/mdBook/format/theme/index.html), so a custom one would have to be created, modified and shipped alongside the (unused) `toc.js`.
2. Collapse/uncollapse animations would be added.

   These are easy to patch out.
3. The selected chapter would also have to be remodified.
4. The UI on mobile devices would presumably also be affected (I haven't tested it)

The non-JS `toc.html` file mentioned above might be usable as a replacement to `{{#toc}}{{/toc}}` if void-docs wishes to steer away from mdBook's default sidebar logic.

# Things to consider after solving this issue
The [`ghcr.io/void-linux/infra-mdbook`](https://github.com/void-linux/void-infrastructure/pkgs/container/infra-mdbook) container will have to be updated.

Once that's done, [this](https://github.com/void-linux/void-infrastructure/blob/032b72320089cb7ab7ea657d5b2d58cc7a9efafc/services/nomad/apps/void-docs.nomad#L26) line will have to be updated to point to the new container. I am not closely familiar with Void's infrastructure, more steps may be necessary.

The `void-docs` package should be revbumped once this issue gets resolved.

meator:
closes #840

I don't have much experience with HTML/CSS, feel free to take over this PR.

The first part of this PR "rebases" the theme files from base mdBook v0.4.14 to mdBook v0.4.43 (current version of `mdBook` package). This was done for several reasons:

- Adding newer and newer things on the old base can have negative effects.
- I split this update into two commits. The first commit overwrites void-docs theme files with the default ones and the second one applied void-docs specific changes. The second commit nicely shows how void-docs configuration differs from the default one (this comparison is much more difficult to make with the current code, but I did it, see "Void's changes to mdBook's default theme" in https://github.com/void-linux/void-docs/issues/840).

The next step is to actually fix the issues in #840. You can see my attempts in my "UNFINISHED" commit.

# Things to consider when solving this issue
The [`ghcr.io/void-linux/infra-mdbook`](https://github.com/void-linux/void-infrastructure/pkgs/container/infra-mdbook) container will have to be updated.

The `void-docs` package should be revbumped once this issue gets resolved.


meator:
I have dealt with similar issues before (1. a utility provides a default "template" file 2. a project overwrites the template file and uses the modified one 3. updates to the utility make changes to the template file which do not propagate to the project using it, because the project completely overwrites it). This is usually desirable behavior, but it means that the project will not get any new features introduced in the updates to the template and it is prone to breakage when backwards-incompatible changes get introduced (like https://github.com/rust-lang/mdBook/issues/2685).

I usually try to solve them by not completely overwriting the entire thing, but trying to overwrite just the part that needs to be changed. New modifications to the "template" file made by updates to the utility (here mdBook) get propagated to the project that way.

There are two ways I know of to achieve this:
1. Use options such as `additional-css` which _append_ to the template instead of overwriting them.

   This solution isn't applicable everywhere.
2. During configuration, generate a default template directory, apply a patch stored in the repo modifying the necessary files and then use this patched theme.

   This approach has several advantages:
   1. It works for things which do not have a way to _append_ configuration (like `additional-css` above).
   2. The patch file stored in the repo clearly documents in what ways does the local configuration deviate from the default one (no need to manually obtain mdBook v0.4.14, generate its default theme and diff it to void-docs one like I had to do).
   3. If the patch doesn't apply (because an update to mdBook changed the patched data), the docs will not build. This is a good thing, if an update to mdBook made breaking changes, the patch should be reassessed and fixed. The current theme would continue "working" in that case.

   This approach has several disadvantages:
   1. It is slightly more complicated to implement, it adds a build time dependency on `patch` or `git am`.
   2. If a breaking change is made to a section of the theme data which is not modified by the patch, the patch application will still succeed, but the change could negatively affect void-docs.

What do you think about this approach?

meator:
If the approach described above gets implemented, commits following my "Reapply void-docs theme changes" commit (including the commit itself) could become the base of the patches which would be applied during build. This is another reason why I split the "rebase" into two commits.

You may find that I'm overthinking this (which may be true), but my (limited) experience shows that tools like `mdbook` tend to introduce a lot of breaking changes in their updates especially when they are heavily configured (which is the case here). This isn't the last time issues like these will occur. And even if mdBook succeeds in being backwards compatible, getting new features implemented in new versions of mdBook would be nice.

PR #844, opened 2025-06-29
Mention target flag in chroot install

axtloss:
It may be useful for users to know what options there are for the --target flag. Someone in the IRC channel had troubles with GRUB because they specified aarch64 instead of arm64-efi, which can be avoided by listing the targets in the docs directly.


ahesford:
This doesn't really clarify the matter. People will still have to read the manual and divine what you mean by "specifying a target". Either make the explanation more explicit, or make it more generic (*e.g.* "You may need to set an appropriate target architecture.")

Issue #754, opened 2023-05-16
Add a section pointing to package development


vincele:
*Before opening a request, please be sure to check [what content is suitable for the Handbook](https://docs.voidlinux.org/about/about-this-handbook.html).*

I've read it and still think it would be nice to have.

We prioritise requests for information specific to Void, and try to avoid duplicating information already available elsewhere (e.g. on the Arch wiki).

This is void-specific.

It would be nice to have some pointers related to packaging.

I looked at:
* voidlinux.org's site header
* Documentation section "4. XBPS Package Manager"
* Documentation section "5. Contributing"

The site or doc could have some links pointing to:
* https://github.com/void-linux/void-packages
* https://github.com/void-linux/void-packages/blob/master/README.md
* https://github.com/void-linux/void-packages/blob/master/Manual.md

WDYT ?
If this is acceptable, I can craft a PR.

sam6816:
Good idea. My connecting point would be "Virtual Packages" (in 4.1 Advanced Usage) and **xbps-create** (not listed). I am not at all against a whole page. But the handbook should at least mention/link this. This issue seems related to (old) #616 which I try to revive.

paper42:
I don't think this is a good idea, we have MANUAL.md and CONTRIBUTING.md in the void-packages repository and this would be just duplicating that information. (also, xbps-create shouldn't really be used outside of xbps-src)

sam6816:
It is not a duplication when you just mention/link it for completeness sake. I thought xbps-create could be used to kind of extend the virtualpkg= functionality. Another open issue 631 has problems with the example itself (virtualpkg=linux:linux5.6). Anyway you look at it, there are loose ends: I can go `man xbps-create` and happily try something.    

"There is also `xbps-create` et al. to foo the bar. For further info go to void-packages repository."

That is all it takes. Otherwise I feel like "Function defined but not called".

My main point is that it is all about xbps-commands and the design of 4. XBPS Package Manager part.

PR #847, opened 2025-07-18
Add `Packaging` subsection

vincele:
When I started building packages for void, I searched for documentation. I initially did not find it, despite having skimmed the chapter containing it. I then proposed to add something to the doc in #754.

After revisiting the issue and searching more thoroughly (someone gave an actual pointer to the right location) I think this can be made a bit more visible by creating a subsection specifically for it. 

I don't think the (small) additionnal space is lost, but if it can help myopic readers, why not.


vincele:
And then there's this other related PR: #639.

PR #850, opened 2025-07-29
config/network-filesystems.md: document NFSv4-only server setup

slymattz:
### Reference
https://github.com/void-linux/void-packages/pull/56488

### check.sh
It ran without errors.

### DIFF
--- a/src/config/network-filesystems.md
+++ b/src/config/network-filesystems.md
@@ -4,12 +4,12 @@
 
 ### Mounting an NFS Share
 
-To mount an NFS share, start by installing the `nfs-utils` and `sv-netmount`
-packages.
-
-Before mounting an NFS share, [enable](./services/index.md#enabling-services)
-the `statd`, `rpcbind`, and `netmount` services. If the server supports `nfs4`,
-the `statd` service isn't necessary.
+To mount an NFS share, install the `nfs-utils` package. If desired, the
+`sv-netmount` package provides a simple service that will automatically mount
+network filesystems at boot. Clients and servers using NFSv3 or older protocols
+require that the `rpcbind` and `statd` service be
+[enabled](./services/index.md#enabling-services). Clients and servers using
+NFSv4 exclusively do not require these services.
 
 To mount an NFS share:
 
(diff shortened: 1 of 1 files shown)


src/config/network-filesystems.md
```diff
@@ -8,8 +8,9 @@ To mount an NFS share, start by installing the `nfs-utils` and `sv-netmount`
 packages.
 
 Before mounting an NFS share, [enable](./services/index.md#enabling-services)
-the `statd`, `rpcbind`, and `netmount` services. If the server supports `nfs4`,
-the `statd` service isn't necessary.
+the `statd`, `rpcbind`, and `netmount` services. Both `statd` and `rpcbind`
+services are only needed for NFSv3, so if you mount shares on your clients like
+that, keep them. For NFSv4 shares, `netmount` is enough.
```
ahesford, line 13: This whole section should be rewritten. Suggest:

> To mount an NFS share, install the `nfs-utils` package. If desired, the `sv-netmount` package provides a simple service that will automatically mount network filesystems at boot. Clients and servers using NFSv3 or older protocols require that the `rpcbind` and `statd` services be [enabled](./services/index.md#enabling-services). Clients and servers using NFSv4 exclusively do not require these services.

src/config/network-filesystems.md
```diff
@@ -69,3 +70,39 @@ server status:
...
+
+As the last step, set `NFSV4_ONLY` to a non-empty value in
+`/etc/sv/nfs-server/conf`
+
+```
+NFSV4_ONLY=true
+```
+
+Now you can [enable](./services/index.md#enabling-services) `nfs-server`
+service.
```
ahesford, line 108: This section should not exist. The last sentence of the preceding section should just say

> You can use [nfs.conf(5)](https://man.voidlinux.org/nfs.conf.5) to configure your server. In particular, to disable legacy protocol versions and support only NFSv4, add the following section:
>
> ```
> [nfsd]
> vers3=n
> vers4=y
> vers4.1=y
> vers4.2=y
> ```
>
> You can verify the configured list of supported versions by inspecting the contents of the file `/proc/fs/nfsd/versions`.

Any mention of the `/etc/sv/nfs-server/conf` file should be omitted until a final decision about how to best to check (or not) for the `statd` server in `nfs-server` has been rendered.

ahesford (changes requested):

PR #851, opened 2025-07-31
[rfc] mention which architectures are built natively, cross, or just supported as build profiles

dkwo:
I think this info can be useful.

### DIFF
--- a/src/installation/index.md
+++ b/src/installation/index.md
@@ -17,7 +17,12 @@ following minimums for most installations:
 
 Note that xfce image installations require more resources.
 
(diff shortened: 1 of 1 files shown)


classabbyamp:
mentioning build profiles and all that is too technical for the "installation" section. documenting that is more for void-packages.

I have included some more information about available architectures in this pr:

https://github.com/void-linux/void-docs/pull/880/changes#diff-35b9da77fdadaf105549a4ec7b2e10f897c96fd3cd71faa8bd08021c4dd0d9fcR21-R24

PR #854, opened 2025-09-02
config/session-management.md: dumb_runtime_dir and pam_rundir

DonPiotr:
Use of dumb_runtime_dir or pam_rundir with seatd as alternative to manual approach for XDG_RUNTIME_DIR.

It is needed. If you don't like my form, please put it in some form you like. ;-)

Thx

### DIFF
--- a/src/config/session-management.md
+++ b/src/config/session-management.md
@@ -85,6 +85,17 @@ user-specific runtime files.
 Install [elogind](#elogind) or [turnstile](#turnstile) as your session manager
 to automatically set up `XDG_RUNTIME_DIR`.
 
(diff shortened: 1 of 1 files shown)


src/config/session-management.md
```diff
@@ -85,6 +85,17 @@ user-specific runtime files.
 Install [elogind](#elogind) or [turnstile](#turnstile) as your session manager
 to automatically set up `XDG_RUNTIME_DIR`.
 
+If using [seatd](#seatd), you can consider those two options:
```
classabbyamp, line 88: turnstile can also be used with seatd

PR #876, opened 2025-12-20
src/config/rc-files.md: add core-services

jastintime:
This adds documentation on `/etc/core-services`.

This feels like the correct place for this addition, please let me know if it belongs elsewhere.

https://github.com/void-linux/void-runit/issues/23#issuecomment-1506310606 Is where I found this information and I felt it would be better to have it in the handbook.


classabbyamp:
(added oxford commas, a couple clarifications about how and when core-services runs)

PR #891, opened 2026-03-11
Add a page on scanning, call out some packages in pipewire.md and amd.md

KeepBotting:
This PR addresses a couple of stumbling blocks I've either experienced myself or come across while hanging out in #voidlinux

The changes are mostly calling out packages which aren't as discoverable as I figure they should be. No package depends upon them, so users might not know they are required for specific functionality.

I have also considered postinstall messages, e.g. the line about `ROCm-SMI` is _really_ about `btop` so maybe it doesn't belong on the docs page for AMD graphics drivers, and should be a postinst message for `btop` instead.

Or perhaps both: since `ROCm-SMI` _is_ undoubtedly related to AMD graphics drivers, the docs could mention it. And `btop` happens to have missing functionality if `ROCm-SMI` is not installed, so a postinst message could address this.

Let me know what you think.

### DIFF
--- a/src/config/graphical-session/graphics-drivers/amd.md
+++ b/src/config/graphical-session/graphics-drivers/amd.md
@@ -32,3 +32,8 @@ used, set the environment variable `LIBVA_DRIVER_NAME` to `radeonsi`.
 
 For VDPAU, install `libvdpau-va-gl`. Set the environment variable `VDPAU_DRIVER`
 to `va_gl`.
+
+## System Management Interface
+
+The `ROCm-SMI` package is required by some programs (like `btop`) in order to monitor
+the GPU.
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -69,6 +69,7 @@
    - [TeX Live](./config/texlive.md)
    - [External Applications](./config/external-applications.md)
    - [Printing](./config/print/index.md)
+   - [Scanning](./config/scan/index.md)
    - [Containers and Virtual Machines](./config/containers-and-vms/index.md)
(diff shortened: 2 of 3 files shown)


classabbyamp:
please separate this into multiple commits, one for each topic

KeepBotting:
This should pass CI now

src/config/graphical-session/graphics-drivers/amd.md
```diff
@@ -32,3 +32,8 @@ used, set the environment variable `LIBVA_DRIVER_NAME` to `radeonsi`.
 
 For VDPAU, install `libvdpau-va-gl`. Set the environment variable `VDPAU_DRIVER`
 to `va_gl`.
+
+## System Management Interface
+
+The `ROCm-SMI` package is required by some programs (`btop`) in order to monitor
```
classabbyamp:
```suggestion
The `ROCm-SMI` package is required by some programs (like `btop`) in order to monitor
```

src/config/media/pipewire.md
```diff
@@ -243,3 +243,5 @@ it](#session-management) and restart PipeWire.
 If a session manager is running, check if your user is in the `audio` and
 `video` groups. If not using `elogind`, this is necessary for PipeWire to access
 devices.
+
+On some platforms, the `sof-firmware` package may be required.
```
classabbyamp: `sof-firmware` is now a dependency of `linux-base`, so this might be unnecessary

KeepBotting:
I'm confused on why this fails the formatting step, something to do with line width maybe?

classabbyamp:
did you run `vmdfmt -w src`?

PR #904, opened 2026-07-17
Added note for users to test fstab before rebooting

SyntaxError2505:
Added instructions for running 'mount -a' to check fstab issues. This can help users at not having to trouble shoot in a scary recovery mode.


classabbyamp:
this will cause partitions to get mounted over the existing mount in the live environment.

SyntaxError2505:
Do you mean that this will break if ran in the live environment? I have installed void linux and use this to test fstab. mount -a will only mount partitions mentioned in fstab and it wouldn't make sense that the live environment usb is inside. Please explain further

classabbyamp:
well, I guess it's not as bad as I was thinking, but per [`mount(8)`](https://man.voidlinux.org/man8/mount.8#COMMAND):

```
           Note that it is a bad practice to use mount -a for fstab checking.
           The recommended solution is findmnt --verify.
```

It'd be good to add a similar note in other places where the fstab is created/edited.

------

btw, you can ignore the link check failure, that's unrelated to your change

SyntaxError2505:
Yep. So you add the notes? I've only noticed fstab being mentioned here. Mount -a is how Google did it when I finally ran into a problem with fstab

SyntaxError2505:
Fixed it for you. Also tried to fix the style thing

PR #906, opened 2026-08-07
asahi: drop firmware, note iso.

dkwo:
there have been some issues reported with outdated isos, so mention it.
the firmware section is outdated, so drop it.


dkwo:
failures are unrelated. @classabbyamp what do you think?

PR #909, opened 2026-09-11
config/external-applications.md: add Java 25 to the versions list

Alter73:
Version 25 has been added as an installable option from the repository.


ahesford:
Although a rewrite to be more generic would be desirable, attempting to keep this list up to date as we roll out Java versions over the years is not.

Also, please review the commit messages in the repository history and align your with the normal form.

Alter73:
> Although a rewrite to be more generic would be desirable, attempting to keep this list up to date as we roll out Java versions over the years is not.
> 
> Also, please review the commit messages in the repository history and align your with the normal form.

Hello, yes, I apologize for not catching the correct commit message format right away. Regarding the Java versions - I added version 25 because it's currently hard to tell if it's available in the repositories, which creates the impression that only the versions listed in this file are supported. It's not quite clear how to universally show which Java versions are available for download without explicitly listing them. Moreover, as far as I understand, older Java versions will eventually be deprecated and removed from the repositories anyway, so the list shouldn't get too cluttered over time.

ahesford:
The point is that we shouldn't have to reconcile this document with whatever state Java packages may be in. As you've noticed, it's easy for this list to become outdated. The guide should remove mention of specific versions and instead provide guidance for how to discover which versions are available.

src/config/external-applications.md
```diff
@@ -26,7 +26,7 @@ instead.
 ### Java
 
 Void provides LTS versions of the OpenJDK development kits and runtimes.
-Currently, versions 8, 11, 17, and 21 are available. To run Java-based
+Currently, versions 8, 11, 17, 21, and 25 are available. To run Java-based
```
classabbyamp, line 29:
```suggestion
The available OpenJDK versions can be listed with `xbps-query --regex -Rs 'openjdk[0-9]+-[0-9]'`. To run Java-based
```

Issue #857, opened 2025-10-24
Better NetworkManager instructions


vinnyhorgan:
Hi!

Was following the docs to switch the networking to use the NetworkManager tool. This was not enough to replace the default `wpa_supplicant` stack. The first issue I encountered was that adding the user to the `network` group as documented was not enough, nmcli still said my user had no permission to connect.

To fix this I had to install and enable the `polkitd` service. I think this is worth specifying in the documentation!

Also I had another issue with the dns. My `/etc/resolv.conf` file was not getting overridden by NetworkManager and had to create a manual symlink.

PR #911, opened 2026-09-13
config/network/networkmanager: clarify need for polkit and DNS config

abrumsen:
Improved NetworkManager instructions to fix issues raised in #857 , namely:
- Adding to the `network` group is not enough, polkitd needs to be installed and running.
- DNS configuration is not forwarded to `resolv.conf` by NetworkManager through `resolvconf` if a file or a symlink already exists (which is the case when installing from the base iso). A simple fix is to delete the file and let it be recreated on a first network connection using NetworkManager.


abrumsen:
The failing link checks are from parts of the handbook that were untouched here, maybe they could be handled in another issue/PR?

